PR #19, opened 2022-02-11
JSON Forms customizations

travjenkins:
1. Collapse all nested controls that are type `object`
2. Display a blank area when the type is `null`
3. Use a custom UI Schema generator
4. Fix typo in new capture details form

`null` setting (nothing under the tab... used to show an error)
image

Collapsible group:
Closed
image

Open
image


src/components/endpointCreation/NewCaptureSpecForm.tsx
```diff
@@ -43,6 +48,8 @@ function NewCaptureSpecForm(props: NewCaptureSpecFormProps) {
             </Alert>
         );
     } else if (sourceSchema !== null) {
+        const uiSchema = generateUISchema(sourceSchema);
+
```
travjenkins, line 52: If we generate this and then pass it into JsonForms (down below) then they will no generate this for us.

src/forms/Helper.ts
```diff
@@ -1,13 +1,27 @@
...
     restrict: true,
     showUnfocusedDescription: true,
 };
 
-export const defaultRenderers = [...materialRenderers];
+export const defaultRenderers = [
+    ...materialRenderers,
+    { tester: nullTypeTester, renderer: NullType },
+    { tester: collapsibleGroupTester, renderer: CollapsibleGroup },
+];
```
travjenkins, line 19: Registering the new renderers

src/forms/Helper.ts
```diff
@@ -1,13 +1,27 @@
...
 
 export const showValidation = (val: any): ValidationMode => {
     return val ? 'ValidateAndShow' : 'ValidateAndHide';
 };
+
+export const generateUISchema = (jsonschema: any): UISchemaElement => {
+    return generateCustomUISchema(jsonschema, 'VerticalLayout') as Layout;
+};
```
travjenkins, line 27: I added an extra layer for this call so if we want to we can mess with the UI Schema. Initially my plan was to put all the collapsible groups at the end but decided against it. However - I think having a place to "play" with the response is valuable.

It also keeps the generate more pure so we can diff it against the original copy.

src/forms/renderers/CollapsibleGroup.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const collapsibleGroupTester: RankedTester = rankWith(
+    999,
+    uiTypeIs('CollapsibleGroup')
+);
+
+const CollapsibleGroupRenderer = (props: any) => {
+    const { uischema, schema, path, visible, renderers } = props;
+
+    const labelElement = uischema.elements.shift();
```
travjenkins, line 24: Since I know we add a label to this type I am pulling it out here. Otherwise the collapsible portion would end up having a double label like this:

image

kiahna-tucker (approved):
Looks good to me.

PR #20, opened 2022-02-14
Improve Top Bar Menu Accessibility.

kiahna-tucker:

1. Remove the overlay background color of the external link buttons within the help menu.

1. Correct the shape of the help menu overlay.

1. Indicate when a menu in the header (i.e., _src/components/header/Topbar.tsx_) is brought in focus by the keyboard.

### DIFF
--- a/src/AppTheme.tsx
+++ b/src/AppTheme.tsx
@@ -3,7 +3,7 @@ import {
     PaletteOptions,
     ThemeOptions,
     ThemeProvider,
-    useMediaQuery,
+    useMediaQuery
 } from '@mui/material';
 import CssBaseline from '@mui/material/CssBaseline';
 import React from 'react';
--- a/src/AppTheme.tsx
+++ b/src/AppTheme.tsx
@@ -89,7 +89,7 @@ const themeSettings = createTheme({
                 //   https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/defaultTheme.ts
                 disableRipple: true,
                 sx: {
-                    '&:hover:active::after': {
+                    '&.Mui-focusVisible::after, &:hover::after': {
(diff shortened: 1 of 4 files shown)


src/AppTheme.tsx
```diff
@@ -89,7 +89,7 @@ const themeSettings = createTheme({
                 //   https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/defaultTheme.ts
                 disableRipple: true,
                 sx: {
-                    '&:hover:active::after': {
+                    '&.Mui-focusVisible::after, &:hover::after': {
```
kiahna-tucker, line 92: The MUI equivalent of the CSS pseudo-class _focus-visible_ was used here, as recommended by the [MUI documentation](https://mui.com/customization/how-to-customize/#state-classes).

src/components/shared/ExternalLink.tsx
```diff
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 ExternalLink.propTypes = {
     children: PropTypes.any.isRequired,
     link: PropTypes.string.isRequired,
+    disableHover: PropTypes.bool,
```
kiahna-tucker, line 8: Leaning towards renaming this prop to `hideOverlay`. To me, it more clearly communicates what is changed when the prop is set.

src/AppTheme.tsx
travjenkins: Good catch!

src/components/shared/ExternalLink.tsx
travjenkins: I agree... I think `hideOverlay` is a better name.

travjenkins (approved):
LGTM

PR #22, opened 2022-02-16
Open Help Menu External Links in New Tab.

kiahna-tucker:

1. Create a shared `ExternalLinkMenuItem` component, to be used whenever an external link needs to be included in a menu. The external links listed in the help menu will open in a new browser tab.

1. Remove the `ExternalLink` component's `disableHover` prop as it is no longer needed.

**NOTE:** The help menu is keyboard accessible with one caveat: to access its menu items, one must tab into the menu after it is opened. Once in focus, the menu handles standard keyboard interactions as expected. Additionally, the styles applied to the `ExternalLinkMenuItem` component replicate that of the `ExternalLink` component for external link uniformity.

### DIFF
--- a/src/components/shared/ExternalLinkMenuItem.tsx
+++ b/src/components/shared/ExternalLinkMenuItem.tsx
@@ -1,0 +1,35 @@
+import OpenInNewIcon from '@mui/icons-material/OpenInNew';
+import { Link, MenuItem } from '@mui/material';
+import PropTypes from 'prop-types';
+
+ExternalLinkMenuItem.propTypes = {
+    children: PropTypes.any.isRequired,
+    link: PropTypes.string.isRequired,
+};
+
+function ExternalLinkMenuItem(
+    props: PropTypes.InferProps<typeof ExternalLinkMenuItem.propTypes>
+) {
+    return (
+        <MenuItem
+            component={Link}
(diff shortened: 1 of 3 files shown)


src/components/shared/ExternalLinkMenuItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+) {
+    return (
+        <MenuItem
+            component={Link}
+            href={props.link}
+            target="_blank"
+            rel="noopener"
+            tabIndex={0}
+            sx={{
+                px: 3,
```
kiahna-tucker, line 21: If you feel strongly about the use of MUI CSS property aliases, let me know.

src/components/shared/ExternalLinkMenuItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+                py: 1.5,
+                fontSize: 14,
+                fontWeight: 700,
+                color: '#3c5584',
```
kiahna-tucker, line 25: A TODO is missing from this line to indicate that the hex color code should be replaced with a theme asset. If you would like that comment added before the code is merged, let me know.

travjenkins (approved):
lgtm

PR #24, opened 2022-02-21
Cleaning out initial experiments from the UI

travjenkins:
1. Removed the "Login Help" page as we won't need it
2. Moved Connectors table to a stand alone tables folder 
    - I like having a small component that fetches its own data and then displays a table of it. I think this pattern might make sense. We shall see though. I have no super strong feelings either way. Could easily see us never needing to include these tables anywhere but a single page... so then maybe it is just needless overhead?
    
3. Updated prettier to be consistent with quoting keys
    - This keeps files more consistent and I think is proper.
   
4. Hiding the spec form on capture creation when no connector has been selected
    - We were showing the skeleton loading screen too soon
   
5. Disabled the pages we are not actively working on
    - No reason to go to an empty page
6. Removed the fake catalog page stuff
    - This was made when I was testing out MUI/React. It has some value from a styling perspective but we can always look at the file in Github history. Since it does not follow any of the current patterns and uses all fake data I am removing most of it.
7. Cleaned up the sortable table stuff
    - The sortable table was very focused on Collections. Made it a bit more reusable.


src/App.tsx
```diff
@@ -33,7 +32,6 @@ const App: React.FC = () => {
                 <Route path="" element={<Login />} />
                 <Route path="login/*">
                     <Route path="" element={<Login />} />
-                    <Route path="help" element={<LoginHelp />} />
```
travjenkins, line 36: Still leaving the ability to nest in case we need it for our proper auth implementation

src/components/capture/creation/index.tsx
```diff
@@ -190,21 +190,26 @@ function NewCaptureModal() {
...
-                                <Divider />
-                                <NewCaptureSpecForm
-                                    displayValidation={showValidation}
-                                    readonly={formSubmitting}
-                                    state={spec.data}
-                                    dispatch={dispatch}
-                                    endpoint={links.spec}
-                                />
-                            </Paper>
+                            {links.connectorImage ? (
```
travjenkins, line 193: Don't need to display this form until we can fetch some connector details.

src/components/header/Topbar.tsx
```diff
@@ -10,7 +10,7 @@ import Logo from '../navigation/Logo';
 
 type TopbarProps = {
     isNavigationOpen: boolean;
-    onNavigationToggle: Function;
+    onNavigationToggle?: Function;
```
travjenkins, line 13: We don't need this function on the login page and previously an empty function was being passed... seems like a foolish pattern to me.

src/components/menus/IconMenu.tsx
```diff
@@ -49,9 +49,9 @@ const IconMenu = (props: IconMenuProps) => {
                 PaperProps={{
                     elevation: 0,
                     sx: {
-                        overflow: 'visible',
-                        filter: 'drop-shadow(0px 2px 8px rgba(0,0,0,0.32))',
-                        mt: 1.5,
+                        'overflow': 'visible',
+                        'filter': 'drop-shadow(0px 2px 8px rgba(0,0,0,0.32))',
+                        'mt': 1.5,
```
travjenkins, line 54: Prettier fixing

src/components/navigation/Navigation.tsx
```diff
@@ -66,19 +66,21 @@ const Navigation = (props: PropTypes.InferProps<typeof NavigationProps>) => {
...
                         isOpen={open}
+                        disabled={true}
+                    />
+                    <ListItemLink
+                        icon={<CodeIcon />}
+                        title="Derivations"
+                        link="/app/derivations"
+                        key="Derivations"
+                        isOpen={open}
+                        disabled={true}
```
travjenkins, line 83: Not important (for now) but I feel like we should order the navigation in order of how folks might use them? Not sure. So I feel like you'd setup a capture and materialization first.

src/components/table/SortableTable.tsx
```diff
@@ -71,40 +59,6 @@ function SortableTable(props: SortableTableProps) {
...
-                selected.slice(0, selectedIndex),
-                selected.slice(selectedIndex + 1)
-            );
-        }
-
-        console.log('Going to set selected to', newSelected);
-
-        setSelected(newSelected);
-    };
-
```
travjenkins, line 107: The table has checkbox/selector functionality that did nothing. Removing it as there were no endpoints that could use something like this.

src/components/table/SortableTable.tsx
```diff
@@ -114,130 +68,83 @@ function SortableTable(props: SortableTableProps) {
...
-                        Paused
-                    </ToggleButton>
-                </ToggleButtonGroup>
-                <TextField
-                    id="outlined-search"
-                    label="Search for keyword"
-                    type="search"
-                    variant="filled"
-                />
-            </Stack>
```
travjenkins, line 155: Removing the fake search / filtering stuff as we don't have any true use for it yet.

src/components/table/SortableTable.tsx
```diff
@@ -114,130 +68,83 @@ function SortableTable(props: SortableTableProps) {
...
+                                                <TableCell align="left">
+                                                    {row.owner}
+                                                </TableCell>
+                                                <TableCell align="left">
+                                                    {row.description}
+                                                </TableCell>
+                                                <TableCell align="left">
+                                                    {row.org}
+                                                </TableCell>
+                                            </TableRow>
```
travjenkins, line 138: All the rows are still obviously only work for data that fits this. Leaving in for now since we aren't really using this component yet.

src/components/table/SortableTableHeader.tsx
```diff
@@ -1,63 +1,35 @@
...
-                <Checkbox
-                    color="primary"
-                    indeterminate={numSelected > 0 && numSelected < rowCount}
-                    checked={rowCount > 0 && numSelected === rowCount}
-                    onChange={onSelectAllClick}
-                    inputProps={{
-                        'aria-label': 'select all collections',
-                    }}
-                />
-            </TableCell>
```
travjenkins, line 60: We aren't including the checkbox in the table so no need in the header.

src/components/table/SortableTableToolbar.tsx
```diff
@@ -1,55 +1,28 @@
...
 
 interface EnhancedTableToolbarProps {
+    toolbarID: string;
     header: string;
-    numSelected: number;
 }
 
 function SortableTableToolbar(props: EnhancedTableToolbarProps) {
-    const { header, numSelected } = props;
+    const { header, toolbarID } = props;
```
travjenkins, line 9: We should have the ID passed in and not hard coded.

src/components/table/SortableTableToolbar.tsx
```diff
@@ -1,55 +1,28 @@
...
             sx={{
                 pl: { sm: 2 },
                 pr: { xs: 1, sm: 1 },
-                ...(numSelected > 0 && {
-                    bgcolor: (theme) =>
-                        alpha(
-                            theme.palette.primary.main,
-                            theme.palette.action.activatedOpacity
-                        ),
-                }),
```
travjenkins, line 26: Can't select anything so no need for styling.

src/components/table/SortableTableToolbar.tsx
```diff
@@ -1,55 +1,28 @@
...
-                >
-                    {numSelected} selected
-                </Typography>
-            )}
-
-            <Tooltip title="Filter list">
-                <IconButton>
-                    <FilterListIcon />
-                </IconButton>
-            </Tooltip>
```
travjenkins, line 52: Removing the selected counter and fake filter button

src/lang/en-US.ts
```diff
@@ -18,6 +18,20 @@ const enUSMessages: ResolvedIntlConfig['messages'] = {
     'cta.download': `Download`,
     'cta.login': `Login`,
     'cta.clickHere': `click here`,
+    'cta.moreDetails': `More Details`,
+
+    // Terms
+    'terms.connectors': `Connectors`,
```
travjenkins, line 24: No clue what to call this... maybe this should just go up with `common`? Basically a place for us to store all the common terms we'll be using.

src/lang/en-US.ts
```diff
@@ -18,6 +18,20 @@ const enUSMessages: ResolvedIntlConfig['messages'] = {
...
+
+    // Data
+    'data.name': `Name`,
+    'data.description': `Description`,
+    'data.type': `Type`,
+    'data.maintainer': `Maintainer`,
+    'data.updated_at': `Last Updated`,
```
travjenkins, line 31: The idea here is that we'll be displaying keys of data from the server in lots of places. So we should have standard keys (when possible) to match them. The second portion of the key is the key from the server. Not truly sure if we want this kind of dependency though.

src/lang/en-US.ts
```diff
@@ -31,11 +45,7 @@ const enUSMessages: ResolvedIntlConfig['messages'] = {
     'username.label': `User Name`,
 
     // Login Page Stuff
-    'login.main.message': `This isn't a real login form. Whatever username you enter will be used in the UI.`,
-    'login.help.message': `If you need help logging in`,
-
-    //Login Help Page
-    'loginHelp.header': `Login Help`,
+    'login.main.message': `When running locally you can login with whatever name you want.`,
```
travjenkins, line 48: Make the message a bit nicer

src/lang/en-US.ts
```diff
@@ -18,6 +18,20 @@ const enUSMessages: ResolvedIntlConfig['messages'] = {
...
+
+    // Data
+    'data.name': `Name`,
+    'data.description': `Description`,
+    'data.type': `Type`,
+    'data.maintainer': `Maintainer`,
+    'data.updated_at': `Last Updated`,
+
+    // Connectors table
+    'connectors.title': `Table of all connectors`,
```
travjenkins, line 34: This is used for ARIA tag. Do we want to call out when something is ARIA specific? I flip flopped a lot on that.

src/pages/Admin.tsx
```diff
@@ -27,82 +12,13 @@ const Admin = () => {
...
-                        </TableContainer>
-                    ) : null}
-                </Box>
+            <Box
+                sx={{
+                    maxHeight: 250,
+                    overflow: 'auto',
+                }}
+            >
+                <ConnectorsTable />
```
travjenkins, line 21: When including the data table I think we should still include a wrapper so the parent can set the size. Left the header inside the table that way we have consistent language.

src/pages/Collections.tsx
```diff
@@ -1,76 +1,15 @@
...
-            maxWidth={false}
             sx={{
-                paddingTop: 2,
+                pt: 2,
             }}
         >
             <Toolbar />
-            <CardCarousel cards={fakeCards} />
-            <ExploreTable />
+            <SortableTable rows={[]} headers={[]} />
```
travjenkins, line 12: Removed calling to the Carousel and the ExplorTable. The extra wrapper around the sortable table made no sense. Removing the card carousel as we do not have an actual use for it yet.

src/pages/Login.tsx
```diff
@@ -47,9 +47,9 @@ const Login: React.FC = () => {
                 height: '100vh',
             }}
         >
-            <Topbar isNavigationOpen={false} onNavigationToggle={() => {}} />
+            <Topbar isNavigationOpen={false} />
```
travjenkins, line 50: No longer have to pass in an empty function.

src/pages/Login.tsx
```diff
@@ -80,14 +80,6 @@ const Login: React.FC = () => {
...
                         </form>
                     </CardActions>
-                    <CardContent>
-                        <Typography variant="caption" color="initial">
-                            <FormattedMessage id="login.help.message" />
-                            <NavLink to="/login/help">
-                                <FormattedMessage id="cta.clickHere" />
-                            </NavLink>
-                        </Typography>
-                    </CardContent>
```
travjenkins, line 90: We probably won't ever have a "help logging in" since we'll be using OpenID or something like that.

src/tables/Connectors.tsx
```diff
@@ -0,0 +1,117 @@
...
+        },
+        {
+            field: 'attributes.maintainer',
+            headerIntlKey: 'data.maintainer',
+        },
+        {
+            field: 'attributes.updated_at',
+            headerIntlKey: 'data.updated_at',
+        },
+    ];
```
travjenkins, line 41: Storing the columns to make the JSX easier to generate.

src/tables/Connectors.tsx
```diff
@@ -0,0 +1,117 @@
...
+                                            key={`${column.field}-${index}`}
+                                        >
+                                            <FormattedMessage
+                                                id={column.headerIntlKey}
+                                            />
+                                        </TableCell>
+                                    );
+                                })}
+                            </TableRow>
+                        </TableHead>
```
travjenkins, line 85: Styled the header with the default background color... not sure if this is the best pattern but it made it work. We should check back in on this later.

src/tables/Connectors.tsx
```diff
@@ -0,0 +1,117 @@
...
+                                        {row.attributes.maintainer}
+                                    </TableCell>
+                                    <TableCell>
+                                        <FormattedDate
+                                            day="numeric"
+                                            month="long"
+                                            year="numeric"
+                                            value={row.attributes.updated_at}
+                                        />
+                                    </TableCell>
```
travjenkins, line 106: The rows are still hardcoded and not using the column headers. This makes it much easier to have custom formatting for each piece of data... without just recreating the `datagrid` component from MUI.

PR #26, opened 2022-02-22
Auth related changes

travjenkins:
1. Add default Axios config
2. Created a `service` pattern
3. Added new high route that checks for auth
4. Save "auth" into local storage
5. Move API defaults to `.env`


src/App.tsx
```diff
@@ -25,39 +26,50 @@ const Logs = React.lazy(() => import('./pages/Logs'));
 const App: React.FC = () => {
     return (
         <Suspense fallback={<Skeleton animation="wave" />}>
-            <Routes>
-                {
-                    // TODO - do we need the default path to use a "redirector"/"landing" page kind of thing?
-                }
-                <Route path="" element={<Login />} />
-                <Route path="login/*">
+            <CheckAuth>
```
travjenkins, line 29: This is the  real change here. We want to check if the user is already logged in way up in the stack so we can handle stuff early. This is basically just a "is there a user in local storage? Then make sure they're logged in"

src/auth/CheckAuth.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    useEffect(() => {
+        if (user) {
+            setAuthHeader(user);
+            if (pathname === '/' || pathname.includes('login')) {
+                navigate('/dashboard', { replace: true });
+            } else {
+                navigate(pathname, { replace: false });
+            }
+        }
```
travjenkins, line 19: As long as there is a user we should set the auth header for axios. This includes if they log out and user is null.

Also - if they are already logged in and trying to go to either of the "log in paths" then we should send 'em to the main page.

src/auth/Context.ts
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 interface AuthContextType {
-    user: string | null;
+    user: string | null | undefined;
```
travjenkins, line 4: The `useLocalStorage` allows undefined so I had to add it.

src/auth/Provider.tsx
```diff
@@ -1,20 +1,27 @@
...
 import AuthContext from './Context';
-import { fakeAuthProvider } from './fakeAuth';
+import { authToken, fakeAuthProvider } from './fakeAuth';
 
 const AuthProvider = ({ children }: { children: React.ReactNode }) => {
-    const [user, setUser] = React.useState<string | null>(null);
+    const [user, setUser, removeUser] = useLocalStorage<string | null>(
+        authToken,
+        null
+    );
```
travjenkins, line 11: Storing the user in local storage so it works between reloads.

src/auth/Provider.tsx
```diff
@@ -1,20 +1,27 @@
...
 
     const signin = (newUser: string, callback: VoidFunction) => {
         return fakeAuthProvider.signin(() => {
             setUser(newUser);
+            setAuthHeader(newUser);
```
travjenkins, line 16: When a use logs in we want to set the auth headers in axios.

src/auth/Provider.tsx
```diff
@@ -1,20 +1,27 @@
...
             callback();
         });
     };
 
     const signout = (callback: VoidFunction) => {
         return fakeAuthProvider.signout(() => {
-            setUser(null);
+            removeUser();
+            setAuthHeader();
```
travjenkins, line 24: Removing auth heads.

src/auth/fakeAuth.ts
```diff
@@ -1,13 +1,12 @@
 const fakeAuthProvider = {
-    isAuthenticated: false,
     signin(callback: VoidFunction) {
-        fakeAuthProvider.isAuthenticated = true;
         setTimeout(callback, 100); // fake async
     },
     signout(callback: VoidFunction) {
-        fakeAuthProvider.isAuthenticated = false;
```
travjenkins, line 8: We weren't using this so I removed it.

src/auth/fakeAuth.ts
```diff
@@ -1,13 +1,12 @@
...
         setTimeout(callback, 100);
     },
 };
 
-export { fakeAuthProvider };
+const authToken = 'auth-token';
```
travjenkins, line 10: The key we use in local storage.

src/components/capture/creation/DetailsForm.tsx
```diff
@@ -89,7 +93,7 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
 
             <Stack direction="row" spacing={2}>
                 {fetchingConnectorsError ? (
-                    <>Error</>
+                    fetchingConnectorsError
```
travjenkins, line 96: Displaying the proper error

src/components/capture/creation/index.tsx
```diff
@@ -122,7 +122,7 @@ function NewCaptureModal() {
                 setFormSubmitting(true);
                 setFormSubmitError(null);
                 setActiveStep(Steps.WAITING_FOR_DISCOVER);
-                axios
+                axiosInstance
```
travjenkins, line 125: This change is all over... we no longer want the default axios we want to make sure we fetch the one with our settings.

src/services/axios.ts
```diff
@@ -0,0 +1,45 @@
+import axios, { AxiosResponse } from 'axios';
+
+axios.defaults.baseURL = process.env.REACT_APP_API_BASE_URL;
+axios.defaults.headers.post['Content-Type'] = 'application/json';
```
travjenkins, line 4: Using default url from env properties. Also, _pretty_ sure the best default content type will be JSON... but not totally sure.

src/services/axios.ts
```diff
@@ -0,0 +1,43 @@
...
+        // Do something before request is sent
+        console.log('Request interceptor precall', config);
+        return config;
+    },
+    (error) => {
+        // Do something with request error
+        console.log('Request interceptor error', error);
+        return Promise.reject(error);
+    }
+);
```
travjenkins, line 17: Nothing here... just including as an example for now.

src/services/axios.ts
```diff
@@ -0,0 +1,43 @@
...
+        console.log('Response interceptor precall', response);
+        return response;
+    },
+    (error) => {
+        // Any status codes that falls outside the range of 2xx cause this function to trigger
+        // Do something with response error
+        console.log('Response interceptor error', error);
+        return Promise.reject(error);
+    }
+);
```
travjenkins, line 32: Nothing here right now but this will be where we handle the server responding if the auth failed a check.

src/services/axios.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const setAuthHeader = (token?: string) => {
+    console.log('This is where we would set the header', token);
+    // if (token) {
+    //     axios.defaults.headers.common.Authorization = token;
+    // } else {
+    //     delete axios.defaults.headers.common.Authorization;
+    // }
+};
```
travjenkins, line 41: Need an easy public way to set the header. Commenting out the actual setting as there is still some CORS related stuff on the server to handle.

src/services/jsonforms.ts
```diff
@@ -1,29 +1,3 @@
-/*
```
travjenkins, line 1: Nothing really changed... just moved this over to the `services` folder.

PR #28, opened 2022-02-25
Lint ordering, useAsync hook, refactor auth, refactor providers, typing

travjenkins:
1. Added basic tests for Auth and Unauth views
    - Super basic tests but I think they are about the granularity we want right now. Moving forward we probably do not want to do snapshot tests until we know we are "done" with the view.
3. Added the userAsync hook
    - This is a helpful pattern I learned from Kent C Dodds. Allows us to run async code on the load of a page.
4. Moved the context providers around
    - I think keeping them all together in a single folder makes thins easier. Also, wrapping them all up in a "barrel component" makes using them in unit testing easier
5. Enabled eslint to fix ordering
    - It was annoying fixing order mistakes manually. This CAN cause issues but I think we should just steer clear of writing code relying on the order of properties.
7. Added vscode stuff for testing
    - Recommend the Jest VSCode plugin and checked in the launcher for it. Makes working through tests a lot easier.
8. More typing for server responses


src/apps/index.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        user: undefined,
+    });
+
+    await waitFor(() => {
+        expect(
+            screen.getByText(
+                'When running locally you can login with whatever name you want.'
+            )
+        ).toBeInTheDocument();
+    });
```
travjenkins, line 17: This await is to allow time for the loading spinner to disappear. Not sure why I couldn't get the waiter added to the `customRender` working... but we can add that functionality later.

PR #32, opened 2022-03-02
Authentication - round 2

travjenkins:
1. Making proper calls to session
2. Make tests pass


saterus:
@travjenkins Does this mean the UI works with https://github.com/estuary/flow/pull/387? :grin:

PR #34, opened 2022-03-03
Auth V3

travjenkins:
### Changes
1. Saving off the session and authdetails (user) into local storage
2. Checking if the token has expired on load
3. Axios fetches headers every request
4. Super simple unit test of the Auth Context auto fetching user

### Tests
1. Added some super simple unit tests.
2. Generally I think this functionality will be getting SLAMMED by people just using/deving the app right now and we should notice fast if authentication isn't working.

### Content
N/A

### Screenshots
While we try to figure out if the user is logged in we should a full page spinner
image

Show an error if the Control Plane is down.
More to come with [Issue 35](https://github.com/estuary/ui/issues/35)
image

Display username when logged in
image


src/context/Auth.tsx
```diff
@@ -14,25 +14,16 @@ export interface AuthContextType {
...
-                    user = null;
-                });
+    const authDetails = auth.getAuthDetails();
 
-            return user;
+    if (authDetails?.session) {
+        if (isFuture(new Date(authDetails.session.expires_at))) {
+            user = authDetails.user?.display_name;
+        } else {
+            user = null;
```
travjenkins, line 23: We now have all the details in the local storage so we no longer have to call the server.

src/context/__tests__/Auth.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return render(<AppContent>{ui}</AppContent>);
+};
+const fakeUser = 'FooBar123';
+const userMissing = 'No User Found';
+
+const FakeComponent = () => {
+    const { user } = useAuth();
+
+    return user ? user : userMissing;
+};
```
travjenkins, line 20: Super basic setup to render out a fake component just to test the user value from `useAuth`. Obviously we could handle this with a real component test elsewhere but I like having this little stand alone test since there is logic inside the context provider.

src/services/auth.ts
```diff
@@ -1,8 +1,14 @@
 import { AxiosResponse } from 'axios';
 import { AccountResponse, AuthLocalResponse } from '../types';
-import axios, { setAuthHeader } from './axios';
+import axios from './axios';
 
-export const authDetailsKey = '__auth_details__';
+export interface AuthDetails {
+    session: AuthLocalResponse['data']['attributes'];
+    user?: AccountResponse['data']['attributes'];
+}
```
travjenkins, line 8: Adding some basic typing for what we're gonna store in local storage

src/services/auth.ts
```diff
@@ -1,8 +1,14 @@
...
+
+export const sessionStorageKey = '__auth_session__';
+export const userStorageKey = '__auth_user__';
```
travjenkins, line 11: Storing the two items separate as I have a feeling this gives us the most flexibility for future changes.

src/services/auth.ts
```diff
@@ -19,23 +25,30 @@ export const auth = {
...
-            return null;
+            response = { session: JSON.parse(session) };
+
+            if (user) {
+                response.user = JSON.parse(user);
+            }
         }
+
+        return response;
+    },
```
travjenkins, line 43: We store session and use in different slots in storage but we return them together. That way there is just a single call to fetch.

src/services/axios.ts
```diff
@@ -4,22 +4,23 @@ import { auth } from './auth';
...
+            config.auth = {
+                password: authDetails.session.token,
+                username: authDetails.session.account_id,
+            };
+        }
+
+        return config;
+    },
+    (error) => {
+        return Promise.reject(error);
```
travjenkins, line 21: Instead of setting these once we fetch them fresh everytime. This is helpful so we can't run into an issue where local storage got cleared out and the `setAuthHeader` function wasn't called.

src/services/axios.ts
```diff
@@ -4,22 +4,23 @@ import { auth } from './auth';
...
+axios.interceptors.request.use(
+    (config) => {
+        const authDetails = auth.getAuthDetails();
+
+        if (authDetails) {
+            config.auth = {
+                password: authDetails.session.token,
+                username: authDetails.session.account_id,
+            };
+        }
```
travjenkins, line 16: This is still only for "simple auth"

src/services/axios.ts
```diff
@@ -41,11 +42,7 @@ export const withAxios = (
 ) => {
     return new Promise<AxiosResponse<any, any>>((resolve: any, reject: any) => {
         fn.then((response) => {
-            if (response.data.redirect) {
-                void auth.signout();
-            } else {
-                resolve(response);
-            }
+            resolve(response);
```
travjenkins, line 45: This was fake code and no longer needed.

src/setupTests.js
```diff
@@ -1,6 +1,10 @@
 import '@testing-library/jest-dom';
 import { act, configure } from '@testing-library/react';
 
+// https://github.com/clarkbw/jest-localstorage-mock#in-create-react-app
+require('jest-localstorage-mock');
```
travjenkins, line 5: Due to using CRA we are a bit limited with getting "extras" into our test.

src/setupTests.js
```diff
@@ -26,10 +30,16 @@ export const setupTests = () => {
...
 };
-// REACT INTL
+
+// Cleaning up after our localstorage mock. It requires we disable
+//  jest auto cleaning up mocks so we call the clerAllMocks ourselves
 afterEach(async () => {
-    // await waitFor(() => expect(queryCache.isFetching).toBe(0));
+    localStorage.clear();
+    jest.clearAllMocks();
+});
```
travjenkins, line 39: Want to make sure everything is cleaned up

src/utils/test-utils.tsx
```diff
@@ -7,19 +7,37 @@ import AppContent from 'context/Content';
...
+        user: {
+            created_at: '',
+            display_name: username,
+            email: `${username}@${username}`,
+            id: '',
+            name: username,
+            norm_name: username,
+            updated_at: '',
+        },
     };
```
travjenkins, line 33: Stand in for a "default" mock session and user in local storage. Not really used right now but will eventually be valuable.... maybe? :)

kiahna-tucker (approved):
Overall, these changes look fine to me. There are a couple places where the code could be simplified, but these improvements would be minor and can be addressed in the near future.

PR #36, opened 2022-03-04
Removing Axios

travjenkins:
These two references  were helpful in building this.
https://danlevy.net/you-may-not-need-axios/
https://kentcdodds.com/blog/replace-axios-with-a-simple-custom-fetch-wrapper

### Changes

1. Remove axios and replace with custom `fetch` wrapper
2. No longer have hooks all manage their own state and just use `useAsync`
3. Creating `endpoints` and storing the server request/response types and functions there
4. Some minor QOL changes.


src/components/capture/creation/DetailsForm.tsx
```diff
@@ -22,12 +22,9 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
     const intl = useIntl();
     const { state, dispatch, readonly, displayValidation } = props;
 
-    const {
-        data: { connectors },
-        loading,
-        error,
-    } = useConnectors();
+    const { data: connectorsData, isError, isSuccess, error } = useConnectors();
```
travjenkins, line 25: This is how the new hooks are consumed in forms... just using the `useAsync` response.

src/components/capture/creation/DetailsForm.tsx
```diff
@@ -22,12 +22,9 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
...
 
+    const [isPostProcessingDone, setPostProcessingDone] = useState(false);
```
travjenkins, line 27: One pattern for handling post processing.... there is another in here as well. We'll eventually figure out what one we like better.

src/components/capture/creation/DetailsForm.tsx
```diff
@@ -76,21 +73,26 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
...
-                previous.properties.image.oneOf = connectors.map(
-                    (connector) => {
-                        return {
-                            const: connector.links.images,
-                            title: connector.attributes.name,
-                        };
-                    }
-                );
+        if (isSuccess) {
+            if (connectorsData && connectorsData.data.length > 0) {
```
travjenkins, line 77: Main change is now we don't personally setup the state in the form so have to do some extra checks.

src/components/capture/creation/DetailsForm.tsx
```diff
@@ -76,21 +73,26 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
...
+                    );
+                    previous.properties.image.oneOf = listOfConnectors;
 
-                return previous;
-            });
+                    return previous;
+                });
+            }
+
+            setPostProcessingDone(true);
```
travjenkins, line 93: This is where a form wants to do some "post processing" of the data before displaying itself. So we run some code and then set that state.

src/components/capture/creation/DetailsForm.tsx
```diff
@@ -99,22 +101,7 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
...
-                            width="50%"
-                        />
-                        <Skeleton
-                            variant="rectangular"
-                            height={40}
-                            width="50%"
-                        />
-                    </>
-                ) : connectors.length > 0 ? (
+                {isPostProcessingDone ? (
```
travjenkins, line 104: When you have a post processor it does not matter if it is idle, loading, whatever... it shows loading until the processing is done.

src/components/capture/creation/DetailsForm.tsx
```diff
@@ -138,10 +125,26 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
...
                         }}
                     />
-                ) : (
+                ) : isError ? (
+                    error
+                ) : schema.properties.image.oneOf.length < 0 ? (
                     <Alert severity="warning">
                         <FormattedMessage id="captureCreation.missingConnectors" />
+                        {schema.properties.image.oneOf.length}
                     </Alert>
```
travjenkins, line 134: Need to handle a rare case where no connectors are installed.

src/components/capture/creation/DetailsForm.tsx
```diff
@@ -138,10 +125,26 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
                             }
                         }}
                     />
```
travjenkins, line 127: If the processing is done... then show the form.

src/components/capture/creation/DetailsForm.tsx
```diff
@@ -138,10 +125,26 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
...
+                            height={40}
+                            width="50%"
+                        />
+                        <Skeleton
+                            variant="rectangular"
+                            height={40}
+                            width="50%"
+                        />
+                    </>
                 )}
```
travjenkins, line 148: The fall through "default" is loading since our first check is if post processing is done.

src/components/capture/creation/SpecForm.tsx
```diff
@@ -29,45 +29,50 @@ const defaultAjv = createAjv({ useDefaults: true });
...
-        loading,
-        error,
-        data: {
-            specSchema,
-            links: { discovery, documentation },
-        },
-    } = useConnectorImageSpec(endpoint);
+        links: { discovery, documentation },
+        endpointSchema,
+    } = data;
```
travjenkins, line 38: This hook provides a default state. Makes it nice to destructure.... but does mean down below we need some more specific tests.

src/components/capture/creation/SpecForm.tsx
```diff
@@ -29,45 +29,50 @@ const defaultAjv = createAjv({ useDefaults: true });
...
-        dispatch({
-            payload: discovery,
-            type: ActionType.NEW_DISCOVERY_LINK,
-        });
+        if (discovery.length > 0) {
+            dispatch({
+                payload: discovery,
+                type: ActionType.NEW_DISCOVERY_LINK,
+            });
+        }
```
travjenkins, line 46: Since we have a default state we need to make sure some real values were put in before dispatching.

src/components/capture/creation/SpecForm.tsx
```diff
@@ -78,28 +83,32 @@ function NewCaptureSpecForm(props: NewCaptureSpecFormProps) {
...
+        const handlers = {
+            onChange: (form: any) => {
+                if (!isEmpty(form.data)) {
+                    dispatch({
+                        payload: form,
+                        type: ActionType.CAPTURE_SPEC_CHANGED,
+                    });
+                }
+            },
+        };
```
travjenkins, line 98: Two of these were inline with the JSX... think it makes it a bit cleaner. Handlers are only needed when the form is going to display so that is why they are currently down here and not up above.

src/components/capture/creation/index.tsx
```diff
@@ -141,10 +141,11 @@ function NewCaptureModal() {
...
                 setFormSubmitError(null);
                 setActiveStep(Steps.WAITING_FOR_DISCOVER);
-                axiosInstance
-                    .post(links.discovery, spec.data)
+
+                discoveryEndpoint
+                    .create(links.discovery, spec.data)
                     .then((response) => {
-                        setCatalogResponse(response.data.data.attributes);
+                        setCatalogResponse(response.data.attributes);
```
travjenkins, line 148: This is basically the core change of removing AXIOS. You now call a specific endpoint and the promise return no longer requires the `.data.data` approaching to getting data.

src/context/Auth.tsx
```diff
@@ -42,7 +42,7 @@ export const AuthProvider = (props: any) => {
         setData,
         setError,
         status,
-    } = useAsync();
+    } = useAsync<string | null>();
```
travjenkins, line 45: Just typing that right now "user" is just a "username"

src/endpoints/account.ts
```diff
@@ -0,0 +1,28 @@
...
+            self: string;
+        };
+    };
+}
+
+export const accountEndpoint = {
+    read: (endpoint: SessionLocalResponse['data']['links']['account']) => {
+        return client<AccountResponse>(endpoint);
+    },
+};
```
travjenkins, line 28: This is a current pattern for handling storing details about endpoints. We can type our requests and responses as needed. I am thinking the methods can be the basic CRUD ones. However, if that ends up being limiting we can change it.

src/endpoints/accounts.ts
```diff
@@ -0,0 +1,16 @@
...
+    links: {
+        self: string;
+    };
+}
+
+export const accountsEndpoint = {
+    read: () => {
+        return client<AccountsResponse>('accounts');
+    },
+};
```
travjenkins, line 16: This one is kinda weird and not sure about.... we have the concept of singular and plural endpoints. So I broke them into two files and just had the plural one reference the singular one's data.

src/endpoints/connectors.ts
```diff
@@ -0,0 +1,86 @@
...
+        spec: {
+            read: (endpoint: ConnectorImageData['links']['spec']) => {
+                return client<ConnectorImagesSpecResponse>(endpoint);
+            },
+        },
+    },
+    read: () => {
+        return client<ConnectorsResponse>('connectors');
+    },
+};
```
travjenkins, line 86: This is the first endpoint with actual pathing in it. Decided the quickest and easier approach now is just nest things where the path matches the JSON key.

src/endpoints/discovery.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const discoveryEndpoint = {
+    create: (
+        endpoint: ConnectorImagesSpecResponse['data']['links']['discovery'],
+        data: any
+    ) => {
+        return client<DiscoveryResponse>(endpoint, { data });
+    },
+};
```
travjenkins, line 25: This is an example where the CRUD naming kind of falls down. It feels weird to say we are "creating" a discovery... but we kind of are... but it still feels weird.

travjenkins: Still need to make a request type... will handle that later.

src/hooks/useAccounts.ts
```diff
@@ -1,42 +1,16 @@
...
-        setLoading(true);
-        setError(null);
-        withAxios(axios.get('/accounts'), setError, setLoading)
-            .then((response: AxiosResponse<AccountsResponse>) => {
-                setAccounts(response.data.data);
-            })
-            .catch(() => {});
-    }, []);
+    const response = useAsync<AccountsResponse>();
+    const { run } = response;
```
travjenkins, line 7: We no no longer need each hook to handle their own state of loading, errors, data, etc. We just pass the useAsync hook right now. This could be bad... like we might want to make a mediator type and only return certain things. Like we don't care much about 'idle' yet so there is some extra fluff being passed along. I think as we work with this and see how it feels to use we'll figure out a good pattern

src/hooks/useAsync.ts
```diff
@@ -33,10 +27,28 @@ const useSafeDispatch = (dispatch: any) => {
...
+    reset: () => any;
+    run: (promise: Promise<any>, dataFetcher?: Function | undefined) => any;
+    setData: (setDataResponse: T) => void;
+    setError: (setErrorResponse: any) => any;
+    status: States;
+}
+
+function useAsync<T>(): UseAsyncResponse<T, undefined>;
+function useAsync<T>(initialData: T): UseAsyncResponse<T, T>;
+function useAsync<T>(initialData?: T): UseAsyncResponse<T, undefined> {
```
travjenkins, line 46: Easiest way I would figure out how to type that if there is initial data then data should not be nullable.

src/hooks/useAsync.ts
```diff
@@ -33,10 +27,28 @@ const useSafeDispatch = (dispatch: any) => {
...
     const initialStateRef = useRef({
         ...defaultInitialState,
-        ...initialState,
+        ...{
+            data: initialData ? initialData : null,
+        },
```
travjenkins, line 51: Only currently have a use for default the data... so only defaulting that for now. If we later have a reason to default other portions of the state then we can update this then.

src/hooks/useConnectorImages.ts
```diff
@@ -1,44 +1,26 @@
...
-                .catch(() => {});
+            run(
+                connectorsEndpoint.images
+                    .read(imagesURL)
+                    .then((serverResponse) => {
+                        const newestImage = serverResponse.data[whichOne];
+
+                        return Promise.resolve(newestImage);
+                    })
+            );
```
travjenkins, line 19: This is _kinda of_ post processing pattern. I think this one is cleaner and puts more in the hook but then it must always be used (or we start passing settings to the hook like we do here) The connector images hook is a lot more complex example of post processing of data.

src/hooks/useConnectorImagesSpec.ts
```diff
@@ -1,57 +1,61 @@
...
-        links: Pick<NewCaptureState['links'], 'discovery' | 'documentation'>;
+import { useEffect } from 'react';
+
+interface Data {
+    endpointSchema: ConnectorImagesSpecResponse['data']['attributes']['endpointSpecSchema'];
+    links: {
+        discovery: ConnectorImagesSpecResponse['data']['links']['discovery'];
+        documentation: ConnectorImagesSpecResponse['data']['attributes']['documentationURL'];
     };
 }
```
travjenkins, line 15: We can see how this goes. This hook filters out data and returns specific portions of it. All the other hooks just pass along all of it. I think eventually we'll figure out which way is best.

Also... handling the typing like this is kind of silly since it does not really add any benefits other than helping "document" what values were intended for this object.

src/hooks/useConnectorImagesSpec.ts
```diff
@@ -1,57 +1,61 @@
...
+                documentation: data.attributes.documentationURL,
             },
-            specSchema: schema,
-        },
-        error,
-        loading,
-    };
+        });
+    });
+};
```
travjenkins, line 42: I think keeping the "post processing" handled in a stand along function makes things cleaner down below. Also... should end up using this pattern by exporting this and having folks that need it run run it themselves vs always running / passing switches.

src/services/auth.ts
```diff
@@ -13,11 +12,11 @@ export const userStorageKey = '__auth_user__';
...
-            (resolve: any, reject: any) => {
-                return axios
-                    .get(path)
-                    .then((response: AxiosResponse<AccountResponse>) => {
-                        resolve(response.data.data.attributes);
+            (resolve, reject) => {
+                return accountEndpoint
+                    .read(path)
+                    .then((response) => {
+                        resolve(response.data.attributes);
```
travjenkins, line 19: Remove axios and replace with our endpoints

src/services/auth.ts
```diff
@@ -41,6 +40,20 @@ export const auth = {
...
+        if (authDetails?.session) {
+            const token = window.btoa(
+                `${authDetails.session.account_id}:${authDetails.session.token}`
+            );
+
+            response = `Basic ${token}`;
+        }
+
+        return response;
+    },
```
travjenkins, line 56: Axios no longer handles Auth headers for us so we need to generate them now. Decided to put this in the auth service just because it seems heavily related even though it will only ever be used by the client.

src/services/client.ts
```diff
@@ -0,0 +1,53 @@
...
+    const fullEndpoint = /^(http)s?:\/\//i.test(endpoint)
+        ? endpoint
+        : `${process.env.REACT_APP_API_BASE_URL}/${endpoint}`;
+
+    const fetchPromise = window
+        .fetch(fullEndpoint, config)
+        .then(async (response) => {
+            if (response.status === 401) {
+                await auth.signout();
+                return Promise.reject({ message: 'common.loggedOut' });
```
travjenkins, line 40: This does not really handle anything special right now but eventually we can handle it.

src/components/tables/Accounts.tsx
```diff
@@ -72,24 +68,20 @@ function AccountsTable() {
                         </TableRow>
                     </TableHead>
                     <TableBody>
-                        {loading ? (
+                        {isLoading ? (
```
kiahna-tucker, line 71: We should probably touch base on what variable prefix and postfix patterns would like to adopt soon. Using the highlighted code as an example, I prefer for variables to be written in an active voice (e.g. _loading_ as opposed to _isLoading_). When a variable is involved in a conditional expression, the expression is naturally interpreted as a question, therefore the prefix results in a wordy translation (e.g., _if loading, do x_ as opposed to _if is loading, do x_). Just my two cents.

src/endpoints/accounts.ts
kiahna-tucker: Curious if endpoints that return and array of items could be appended with `-list` (e.g., `/accounts` becomes `/account-list`). If it is determined that the endpoint should not change, applying the aforementioned convention to file names and assets of plural endpoints may be worth consideration (e.g., _src/endpoints/accounts.ts_ becomes _src/endpoints/account-list.ts_).

src/hooks/useAsync.ts
```diff
@@ -33,10 +27,28 @@ const useSafeDispatch = (dispatch: any) => {
...
+}
+
+function useAsync<T>(): UseAsyncResponse<T, undefined>;
+function useAsync<T>(initialData: T): UseAsyncResponse<T, T>;
+function useAsync<T>(initialData?: T): UseAsyncResponse<T, undefined> {
     const initialStateRef = useRef({
         ...defaultInitialState,
-        ...initialState,
+        ...{
+            data: initialData ? initialData : null,
```
kiahna-tucker, line 50: Unless mistaken, this statement can be reduced to `initialData || null`.

src/hooks/useAsync.ts
```diff
@@ -62,17 +74,12 @@ const useAsync = (initialState?: any) => {
...
-                );
-            }
+        (promise: Promise<T>) => {
             safeSetState({ status: States.LOADING });
             return promise.then(
-                (runResponseData: any) => {
-                    setData(runResponseData);
-                    return runResponseData;
+                (runResponse) => {
+                    setData(runResponse);
```
kiahna-tucker, line 81: It may be worthwhile to touch base on basic formatting standards for the code base as well. Code density tends to fluctuate on a file basis.

src/hooks/useConnectorImages.ts
```diff
@@ -1,44 +1,26 @@
...
-const useConnectorImages = (
-    imagesURL: string,
-    whichOne: number = 0
-): ConnectorImagesService => {
-    const [loading, setLoading] = useState<boolean>(true);
-    const [error, setError] = useState<string | null>(null);
-
-    const [attributes, setAttributes] = useState<object>({});
-    const [links, setLinks] = useState<object>({});
+const useConnectorImages = (imagesURL: string, whichOne: number = 0) => {
```
kiahna-tucker, line 5: The name `whichOne` does not sit well with me. If there is room to adjust it, would something along the lines of `selection` or `selectedIndex` work?

kiahna-tucker (approved):
Overall, these changes look fine to me. There are a couple places where the code could be simplified and a few general comments, but these items can be addressed in the near future. I would recommend reviewing the comments as there are two placeholders for UI discussion and an inquiry worth reviewing the the control plane meeting tomorrow.

src/hooks/useConnectorImages.ts
travjenkins: Yup... I don't really have strong feelings on this naming.

src/hooks/useAsync.ts
travjenkins: I don't really care much about formatting unless it can be handled with a linter... so basically automatically handled. Otherwise I feel it is something that ends up being worth less than the amount of energy it takes to enforce.

So if there is formatting you'd like to see check if Prettier or Eslint can help and then we can decide.

PR #38, opened 2022-03-08
Create Foundation for Page-Level and Badge Notifications

kiahna-tucker:
### Changes

1. Display a badge notification on the _Change Set_ button of the navigation menu when a new capture is created.

1. Display a page-level notification directly above primary page content when a new capture is created.

1. Update the approach to state selector definitions in components.

**NOTE:** The notification styling is a known work-in-progress and will be addressed separately. Furthermore, the addition of the badge notification on the navigation menu did result in the following MUI warning: _The `anchorEl` prop provided to the component (i.e. `PopperUnstyled`) is invalid. The anchor element should be part of the document layout. Make sure the element is present in the document or that it's not display none._ This warning is encountered when the badge notification on the _Change Set_ button in the navigation menu is cleared. The `Tooltip` component used in the change set table is believed to be the source of the issue. I plan to resolve this warning in a separate PR.

### Screenshots

**Banner Notification Shown | Navigation Menu Collapsed**
pr-screenshot__38__banner-notification-(success)__badge-notification-(collapsed)

<br />

**Banner Notification Shown | Navigation Menu Expanded**
pr-screenshot__38__banner-notification-(success)__badge-notification-(expanded)

<br />

**Banner Notification Hidden | Navigation Menu Collapsed**
pr-screenshot__38__banner-notification-(hidden)__badge-notification-(collapsed)

<br />

**Banner Notification Hidden | Navigation Menu Expanded**
pr-screenshot__38__banner-notification-(hidden)__badge-notification-(expanded)

### DIFF
--- a/src/components/capture/creation/index.tsx
+++ b/src/components/capture/creation/index.tsx
@@ -88,7 +88,6 @@ function NewCaptureModal() {
                     user: 'temp@gmail.com',
                 },
                 schema: schemaFromEditor || catalogResponse,
-                viewed: false,
             };
 
             addCaptureToChangeSet(catalogNamespace, capture);
--- a/src/components/capture/creation/index.tsx
+++ b/src/components/capture/creation/index.tsx
@@ -43,9 +43,9 @@ enum Steps {
 
 const selectors = {
     addCapture: (state: CaptureState) => state.addCapture,
-    addNotification: (state: NotificationState) => state.addNotification,
     removeSchema: (state: SchemaEditorState) => state.removeSchema,
     schema: (state: SchemaEditorState) => state.schema,
+    showNotification: (state: NotificationState) => state.showNotification,
 };
 
 function NewCaptureModal() {
@@ -61,7 +61,7 @@ function NewCaptureModal() {
     const addCaptureToChangeSet = useChangeSetStore(selectors.addCapture);
 
     // Notification store
-    const addNotification = useNotificationStore(selectors.addNotification);
+    const showNotification = useNotificationStore(selectors.showNotification);
 
     // Form data state
     const [state, dispatch] = useReducer(newCaptureReducer, getInitialState());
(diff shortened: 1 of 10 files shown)


src/components/shared/PageContainer.tsx
```diff
@@ -9,8 +20,48 @@ const PageContainerPropTypes = {
...
 
+const selectors = {
+    hideNotification: (state: NotificationState) => state.hideNotification,
+    notification: (state: NotificationState) => state.notification,
+    updateNotificationHistory: (state: NotificationState) =>
+        state.updateNotificationHistory,
+};
+
 function PageContainer(props: PageContainerProp) {
+    const alertTransitionRef = useRef<any>(null);
```
kiahna-tucker, line 31: The attributed type will be narrowed in a subsequent PR. Since I was experimenting with different transition components, the `any` type was temporarily employed to minimize the code changes required for each transition.

src/components/capture/creation/index.tsx
```diff
@@ -37,22 +41,27 @@ enum Steps {
...
-const schemaSelector = (state: SchemaEditorState) => state.schema;
-const removeSchemaSelector = (state: SchemaEditorState) => state.removeSchema;
-
-const addCaptureSelector = (state: CaptureState) => state.addCapture;
+const selectors = {
+    addCapture: (state: CaptureState) => state.addCapture,
+    removeSchema: (state: SchemaEditorState) => state.removeSchema,
+    schema: (state: SchemaEditorState) => state.schema,
+    showNotification: (state: NotificationState) => state.showNotification,
+};
```
travjenkins, line 49: I like it.

travjenkins (approved):
The code is fine and the changes I called out can be handled in another PR in the future.

However, I think the notification is wrong. I don't think the UI should shift around like that for anything (almost... the only case I can think of is a network/server error and we should probably use this for that). The notification should be at a higher zindex when displaying. That way the creation CTA is always in the same place.

PR #43, opened 2022-03-11
Travjenkins/include flow files

travjenkins:
### Changes

1. Pulled in the types from Flow
2. Cleaned up a lot of typing/any issues all over the codebase
3. Removed ESLinting option to reorder object keys
4. Removed Prop Types since we're just using TS for typing needs
5. Cleaned up a random dep from packages that wasn't being used
6. Brock up the `endpoints` interfaces so we could more easily reference nested stuff

### Tests

1. Ran tests and they still pass
2. Built the app and clicked around on my local

### Screenshots

Got the warnings from ESLint down to the ones below. Some are not fixable due to current implementation of JSONForms and Reducers.
image


flow_deps/README.MD
```diff
@@ -0,0 +1,13 @@
+### What is this folder?
+This folder houses the files we pull in from the Estuary/flow project.
```
kiahna-tucker, line 2: Since the name of the repository is _estuary/flow_, you may want to replace the capital _E_ with its lowercase counterpart.

flow_deps/README.MD
```diff
@@ -0,0 +1,13 @@
...
+The types should be used along with the server responses.
+
+### How do we update the types?
+For this you'll need `flowctl` installed and the Flow project pulled locally. In the root of the flow project you will need to run the following command:
+`flowctl schemalate typescript --name catalog --hoist-definitions < flow.schema.json > flow.d.ts`
+
+After that you need to copy the `flow.d.ts` file into the `flow_deps` folder.
+
+### How often do we update these files?
+This is not totally decided yet. Probably whenever the folks that "own" this tell us we need to update. This should not happen _all_ the time... but will be frequent enough we'll want to automate this.
```
kiahna-tucker, line 13: A couple things to consider here:

> This is not totally decided yet.

The following may be a more concise way of communicating the sentiment: It is still undecided.

> Probably whenever the folks that "own" this tell us we need to update.

The object of this statement is slightly ambiguous. Could you remove the noun this and replace it with the actual descriptor?

> This should not happen all the time... but will be frequent enough we'll want to automate this.

I would recommend replacing the ellipsis with a comma.

src/components/navigation/ListItemLink.tsx
```diff
@@ -7,56 +7,52 @@ import {
...
-const ListItemLinkProps = {
-    badgeContent: PropTypes.number,
-    disabled: PropTypes.bool,
-    icon: PropTypes.element,
-    isOpen: PropTypes.bool,
-    link: PropTypes.string.isRequired,
-    menuWidth: PropTypes.number,
-    title: PropTypes.string.isRequired,
-};
+interface ListItemLinkProps {
```
kiahna-tucker, line 14: Personal preference would be to group required and optional properties, with required properties taking precedence.

src/context/__tests__/Auth.tsx
```diff
@@ -16,7 +16,11 @@ const userMissing = 'No User Found';
 const FakeComponent = () => {
     const { user } = useAuth();
 
-    return user ? user : userMissing;
+    if (user) {
```
kiahna-tucker, line 19: Could this not be written with a single return statement?

kiahna-tucker (approved):
Approved with suggestions.

src/context/__tests__/Auth.tsx
travjenkins: Yeah... I was getting frustrated with this one and was trying different things and just landed on this and stopped here. I'll fix it with my next PR.

PR #44, opened 2022-03-15
Add Last Updated Column to Change Set Table

kiahna-tucker:
### Changes

The change set table now includes a _Last Updated_ column that consists of statements indicating when an entity was last altered. For the time being, an _alteration_ is synonymous with the creation of a new entity.

**NOTE:** The time-related message shown in the change set table is updated when the table is rendered. Therefore, the _Builds_ page must be refreshed to update the message displayed in the _Last Updated_ column. This is temporary.

### Screenshots

**Change Set Table with Last Updated Column**
pr-screenshot__44__change-set-table__last-updated-column-added


travjenkins (approved):
LGTM

PR #48, opened 2022-03-17
Some general changes to prepare for next stage of build/changeset dev

travjenkins:
### Changes

- Tweaking readmes/naming/folders/etc. due to feedback
- Refactoring a bit for capture creation.
- Putting in the stub for the confirmation/alert/dialog/etc.
- Added filtering of type to the connectors hooks
- Removing old form reducer pattern and changing to stores

### Tests

Manually tested
Ran unit tests

### Content

There is some content but leaving it hard coded right now

### Screenshots

N/A


src/components/capture/creation/index.tsx
```diff
@@ -46,6 +52,17 @@ const selectors = {
...
+    captureName: (state: CaptureCreationState) => state.details.data.name,
+    setDetails: (state: CaptureCreationState) => state.setDetails,
+    resetState: (state: CaptureCreationState) => state.resetState,
+    hasChanges: (state: CaptureCreationState) => state.hasChanges,
+    errors: (state: CaptureCreationState) => [
+        state.details.errors,
+        state.spec.errors,
+    ],
+    specFormData: (state: CaptureCreationState) => state.spec.data,
+    disoverLink: (state: CaptureCreationState) =>
```
kiahna-tucker, line 64: _Discover_ is misspelled here.

kiahna-tucker (approved):
Approved with a suggestion.

PR #50, opened 2022-03-17
Add confirmation dialog

travjenkins:
### Changes

- Very basic context driven confirmation dialog
- Capture creation has a stand along "exit" function now so we don't call this confirmation when adding to changeset
- Modal will not close if there are changes and the user does not agree
- If the user closes the modal without choosing an option we default that as "no" and don't close the modal
- No real styling added beyond the buttons as I assume we'll handle that in the future.

### Tests

Manual

### Content

- Some new default messaging in the lang folder
- Also created a new "category" for confirmations
- 
image

### Screenshots

image

### DIFF
--- a/src/components/capture/creation/index.tsx
+++ b/src/components/capture/creation/index.tsx
@@ -227,7 +227,6 @@ function NewCaptureModal() {
             fullScreen={fullScreen}
             fullWidth={!fullScreen}
             maxWidth="md"
-            disableEscapeKeyDown
             sx={{
                 '.MuiDialog-container': {
                     alignItems: 'flex-start',
--- a/src/components/capture/creation/index.tsx
+++ b/src/components/capture/creation/index.tsx
@@ -18,6 +18,7 @@ import useCaptureCreationStore, {
     CaptureCreationState,
 } from 'components/capture/creation/Store';
 import ErrorBoundryWrapper from 'components/shared/ErrorBoundryWrapper';
+import { useConfirmationModalContext } from 'context/Confirmation';
 import {
     DiscoveredCatalog,
     discoveredCatalogEndpoint,
@@ -69,6 +70,7 @@ function NewCaptureModal() {
     const theme = useTheme();
     const fullScreen = useMediaQuery(theme.breakpoints.down('sm'));
     const navigate = useNavigate();
+    const confirmationModalContext = useConfirmationModalContext();
 
     // Schema editor store
     const schemaFromEditor = useSchemaEditorStore(selectors.schema);
@@ -104,6 +106,16 @@ function NewCaptureModal() {
     // TODO Schema Editor
     //const [availableSchemas, setAvailableSchemas] = useState<any[]>([]);
 
+    const exitModal = () => {
+        if (schemaFromEditor) {
+            removeSchema();
+        }
+
+        cleanUp();
+
+        navigate('..');
+    };
+
     // Form Event Handlers
     const handlers = {
         addToChangeSet: (event: MouseEvent<HTMLElement>) => {
@@ -136,20 +148,24 @@ function NewCaptureModal() {
 
             setFormSubmitting(true);
 
-            handlers.close();
+            exitModal();
         },
 
         close: () => {
             if (hasChanges()) {
-                console.log('uh oh');
+                confirmationModalContext
+                    ?.showConfirmation({
+                        message: 'You will lose unsaved work if you exit now.',
+                    })
+                    .then((confirmed) => {
+                        if (confirmed) {
+                            exitModal();
+                        }
+                    })
+                    .catch(() => {});
+            } else {
+                exitModal();
             }
-            if (schemaFromEditor) {
-                removeSchema();
-            }
-
-            cleanUp();
-
-            navigate('..');
         },
 
         test: (event: MouseEvent<HTMLElement>) => {
--- a/src/components/capture/creation/index.tsx
+++ b/src/components/capture/creation/index.tsx
@@ -155,7 +155,7 @@ function NewCaptureModal() {
             if (hasChanges()) {
                 confirmationModalContext
                     ?.showConfirmation({
-                        message: 'You will lose unsaved work if you exit now.',
+                        message: 'confirm.loseData',
                     })
                     .then((confirmed) => {
                         if (confirmed) {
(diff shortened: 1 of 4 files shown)


src/components/capture/creation/index.tsx
```diff
@@ -104,6 +106,16 @@ function NewCaptureModal() {
...
 
+    const exitModal = () => {
+        if (schemaFromEditor) {
+            removeSchema();
+        }
+
+        cleanUp();
+
+        navigate('..');
+    };
```
travjenkins, line 117: Sometimes we just want to exit the modal, so making a stand alone function.

src/components/capture/creation/index.tsx
```diff
@@ -136,20 +148,24 @@ function NewCaptureModal() {
...
+                        message: 'confirm.loseData',
+                    })
+                    .then((confirmed) => {
+                        if (confirmed) {
+                            exitModal();
+                        }
+                    })
+                    .catch(() => {});
+            } else {
+                exitModal();
```
travjenkins, line 167: If there are changes we now want to make sure they mean to close. 

If the modal explodes somehow and they get into the `catch` path I _think_ keeping them in the modal is correct. Might be annoying to be stuck there but better than losing work on accident.

src/components/capture/creation/index.tsx
```diff
@@ -211,7 +227,6 @@ function NewCaptureModal() {
             fullScreen={fullScreen}
             fullWidth={!fullScreen}
             maxWidth="md"
-            disableEscapeKeyDown
```
travjenkins, line 214: I only added this while we waiting on the confirmation

src/context/Confirmation.tsx
```diff
@@ -0,0 +1,111 @@
...
+} from '@mui/material';
+import React, { createContext, useContext, useRef, useState } from 'react';
+import { FormattedMessage } from 'react-intl';
+
+export interface IConfirmationModalOptions {
+    confirmText?: string;
+    cancelText?: string;
+    title?: string;
+    message: string;
+}
```
travjenkins, line 17: All things are option except for the main message that is getting displayed.

src/context/Confirmation.tsx
```diff
@@ -0,0 +1,111 @@
...
+            setSettings({
+                ...getDefaultSettings(),
+                ...userSettings,
+            });
+            setShowConfirmationModal(true);
+
+            return new Promise((resolve) => {
+                resolver.current = resolve;
+            });
+        },
```
travjenkins, line 66: We set the settings at the time the confirmation is displayed. I don't think this would be the best way to handle this if this was used in multiple places but we should never have multiple confirmation dialogs at once.

src/context/index.tsx
```diff
@@ -8,7 +9,11 @@ const AppProviders: React.FC = ({ children }) => {
         <AppContent>
             <AuthProvider>
                 <AppTheme>
-                    <AppRouter>{children}</AppRouter>
+                    <AppRouter>
+                        <ConfirmationModalContextProvider>
+                            {children}
+                        </ConfirmationModalContextProvider>
```
travjenkins, line 15: Needs to be at least inside the theme... I think putting it under the router makes the most sense. That way if we ever need to do a "Are you sure you want to leave this page" kinda thing we can.

src/context/Confirmation.tsx
kiahna-tucker: I find interfaces with optional properties easier to understand when the required properties are listed before those that are optional. More often than not, required properties are of primary concern (as without them, the object cannot be created); thus, as the file is read from top to bottom, having them be encountered before the optional properties is a preference of mine. Just sharing my two cents on the matter.

src/lang/en-US.ts
```diff
@@ -36,6 +37,10 @@ const enUSMessages: ResolvedIntlConfig['messages'] = {
     'data.email': `Email`,
     'data.display_name': `User Name`,
 
+    // Confirmations
+    'confirm.title': `Are you sure?`,
+    'confirm.loseData': `You have unsaved worked. If you continue you will lose your changes.`,
```
kiahna-tucker, line 42: The first sentence of the `confirm.loseData` message should be _You have unsaved work._ Currently, the string is using the wrong tense for the verb _work_.

src/components/capture/creation/index.tsx
```diff
@@ -104,6 +106,16 @@ function NewCaptureModal() {
     // TODO Schema Editor
     //const [availableSchemas, setAvailableSchemas] = useState<any[]>([]);
 
+    const exitModal = () => {
+        if (schemaFromEditor) {
+            removeSchema();
+        }
+
+        cleanUp();
```
kiahna-tucker, line 114: The corresponding selector was renamed to `resetState` and I feel this constant should follow suit. Given the number of states accessed by this component, it may be beneficial to indicate which state is being reset: `resetCaptureCreationState`.

The generically named selector does not pose an issue because it is adequately typed and fed to `useCaptureCreationStore()`. However, I believe a stand alone `cleanUp()` or `resetState()` verges on being too ambiguous deeper within the code. Thoughts?

src/context/Confirmation.tsx
```diff
@@ -0,0 +1,111 @@
...
+    Dialog,
+    DialogActions,
+    DialogContent,
+    DialogContentText,
+    DialogTitle,
+} from '@mui/material';
+import React, { createContext, useContext, useRef, useState } from 'react';
+import { FormattedMessage } from 'react-intl';
+
+export interface IConfirmationModalOptions {
```
kiahna-tucker, line 12: I am not a fan of the _I_ prefix on interfaces. They register as superfluous when an adopted pattern already distinguishes interfaces and types. Could you explain the value you find in this pattern?

kiahna-tucker (commented):
There are a couple, minor tweaks that should be made before the code is merged in (e.g., typo in displayed text, etc.) and a few discussion items. While none of the suggestions are blocking, I imagine it may be best to bundle any patches related to this PR at this time. If you would prefer to resolve these comments in the near future, let me know.

src/context/Confirmation.tsx
travjenkins: They make a lot of sense in context files like Confirm and Auth since they have a lot of stuff packed into 'em. That's why I used this here and no where else.

kiahna-tucker (approved):
Approved with suggestions.

src/components/capture/creation/index.tsx
travjenkins: I thought I renamed this when we were on a call - I can change it.

Again I don't think it is an issue as it is the state of the  thing you're in.

PR #52, opened 2022-03-18
Store Multiple Schemas in Schema Editor State

kiahna-tucker:
### Changes

1. Store all resources returned by the _discover_catalog_ endpoint in the schema editor state when the editor is mounted.

1. Retain the alterations made to a given schema in the schema editor state.

1. Rename `CaptureState` to `ChangeSetState` and remove to-dos related to their separation.

1. Rename and reorder assets within the change set and schema editor states.

**NOTE:** To avoid any errors, any change set state in local storage should be deleted once this code is merged.

### DIFF
--- a/src/components/capture/creation/index.tsx
+++ b/src/components/capture/creation/index.tsx
@@ -12,16 +12,16 @@ import {
     Paper,
     Typography,
     useMediaQuery,
-    useTheme,
+    useTheme
 } from '@mui/material';
 import useCaptureCreationStore, {
-    CaptureCreationState,
+    CaptureCreationState
 } from 'components/capture/creation/Store';
 import ErrorBoundryWrapper from 'components/shared/ErrorBoundryWrapper';
 import { useConfirmationModalContext } from 'context/Confirmation';
 import {
     DiscoveredCatalog,
-    discoveredCatalogEndpoint,
+    discoveredCatalogEndpoint
 } from 'endpoints/discoveredCatalog';
 import { MouseEvent, useState } from 'react';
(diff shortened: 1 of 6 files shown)


src/components/capture/creation/index.tsx
```diff
@@ -107,8 +110,8 @@ function NewCaptureModal() {
     //const [availableSchemas, setAvailableSchemas] = useState<any[]>([]);
```
kiahna-tucker, line 110: Can the schema editor to-dos be removed? And would you mind elaborating on the approach you intended to take here?

src/components/capture/creation/index.tsx
```diff
@@ -73,8 +76,8 @@ function NewCaptureModal() {
     const confirmationModalContext = useConfirmationModalContext();
 
     // Schema editor store
-    const schemaFromEditor = useSchemaEditorStore(selectors.schema);
-    const removeSchema = useSchemaEditorStore(selectors.removeSchema);
+    const resourcesFromEditor = useSchemaEditorStore(selectors.resources);
+    const clearResources = useSchemaEditorStore(selectors.clearResources);
```
kiahna-tucker, line 80: Renaming this to `clearResourcesFromEditor` increases code clarity.

src/components/capture/creation/index.tsx
travjenkins: No - because these are not for the _Schema_ Editor they are for an _Editor_ of Schemas. Meaning allowing the customer to choose what schemas that want to bring in from a connector.

travjenkins (approved):
LGTM

PR #54, opened 2022-03-24
Bug fixes and Preparing UI for Alpha MVP

travjenkins:
### Changes

Typing that does not rely on React.FC
Disable "save" button in capture creation during form submit
Memoing ui schema generation just to be safe

### Tests

Manually tested

### Content

None

### Screenshots

None

### DIFF
--- a/src/components/capture/creation/SpecForm.tsx
+++ b/src/components/capture/creation/SpecForm.tsx
@@ -8,7 +8,7 @@ import useCaptureCreationStore, {
 import FormLoading from 'components/shared/FormLoading';
 import useConnectorImageSpec from 'hooks/useConnectorImagesSpec';
 import { isEmpty } from 'lodash';
-import { useEffect } from 'react';
+import { useEffect, useMemo } from 'react';
 import { FormattedMessage } from 'react-intl';
 import {
     defaultOptions,
@@ -53,6 +53,16 @@ function NewCaptureSpecForm(props: NewCaptureSpecFormProps) {
         }
     }, [discovered_catalog, setLink]);
 
+    const uiSchema = useMemo(() => {
+        let response = null;
+
+        if (endpointSchema.type) {
+            response = generateCustomUISchema(endpointSchema);
+        }
+
+        return response;
+    }, [endpointSchema]);
+
     useEffect(() => {
         if (documentation.length > 0) {
             setLink('documentation', documentation);
@@ -86,8 +96,7 @@ function NewCaptureSpecForm(props: NewCaptureSpecFormProps) {
                 {error}
             </Alert>
         );
-    } else if (endpointSchema.type) {
-        const uiSchema = generateCustomUISchema(endpointSchema);
+    } else if (uiSchema) {
         const showValidationVal = showValidation(displayValidation);
         const handlers = {
             onChange: (form: any) => {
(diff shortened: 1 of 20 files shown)


src/components/capture/creation/SpecForm.tsx
```diff
@@ -53,6 +53,16 @@ function NewCaptureSpecForm(props: NewCaptureSpecFormProps) {
...
 
+    const uiSchema = useMemo(() => {
+        let response = null;
+
+        if (endpointSchema.type) {
+            response = generateCustomUISchema(endpointSchema);
+        }
+
+        return response;
+    }, [endpointSchema]);
```
travjenkins, line 64: This can easily be an expensive call.

src/components/capture/creation/SpecFormHeader.tsx
```diff
@@ -20,7 +20,7 @@ function NewCaptureSpecFormHeader() {
     const { data, error } = useConnectorImages(endpoint);
 
     useEffect(() => {
-        if (data?.links.spec && data.links.spec.length > 0) {
+        if (endpoint && data?.links.spec && data.links.spec.length > 0) {
```
travjenkins, line 23: Need to make sure there is an endpoint before setting the spec as it depends on the endpoint

src/components/capture/creation/index.tsx
```diff
@@ -336,7 +336,7 @@ function NewCaptureModal() {
                 ) : (
                     <Button
                         onClick={handlers.test}
-                        disabled={false}
+                        disabled={formSubmitting}
```
travjenkins, line 339: Trying to reduce double clicks

src/context/Confirmation.tsx
```diff
@@ -37,7 +38,7 @@ const getDefaultSettings = (): IConfirmationModalOptions => {
 const ConfirmationModalContext =
     createContext<IConfirmationModalContext | null>(null);
 
-const ConfirmationModalContextProvider: React.FC = ({ children }) => {
+const ConfirmationModalContextProvider = ({ children }: BaseComponentProps) => {
```
travjenkins, line 41: Moving this over to a base component of our own.

src/components/capture/creation/Store.ts
```diff
@@ -4,15 +4,6 @@ import { isEqual } from 'lodash';
 import create from 'zustand';
 import { devtools } from 'zustand/middleware';
 
-enum EventNames {
```
kiahna-tucker, line 7: It should be noted that each word in these string constants are capitalized. When this enumeration was broken down, only the first word in each string is capitalized. While a minor change, this is an opportunity to achieve consistency in the codebase.

src/components/capture/creation/index.tsx
```diff
@@ -94,7 +94,7 @@ function NewCaptureModal() {
     );
     const specFormData = useCaptureCreationStore(selectors.specFormData);
     const disoverLink = useCaptureCreationStore(selectors.disoverLink);
```
kiahna-tucker, line 96: There is a typo in the variable name `disoverLink`. The _c_ is missing in the word _discover_.

kiahna-tucker (approved):
Approved with suggestions.

PR #55, opened 2022-03-29
Lots of tweaks to prepare for deploying to stage

travjenkins:
### Changes

1. Added a local build and deploy to test our prod builds.
2. Fixed bug with too many renders happening when nav panel is opened
3. Made capture creation a page and not a modal
4. Updated paths to be closer to CRUD naming
5. Don't include `devtools` on prod builds
6. Made the errors panel better
7. Made separate buttons for testing/discover and saving
8. Removed items from nav that are not a part of MVP
9. Error page has a "return home" link
10. Disable test button on capture create when no connectors
11. Allowed to update API and Auth endpoints at anytime. 
12. Update to use new auth endpoint
13. Cleaned up old endpoint that are no longer used
14. https://github.com/estuary/ui/pull/64

### Tests

Manual and ran unit tests

### Content

Content changes can be handled before we release

### Screenshots

New navigation links
image

Capture creation is now a page
image

New error panel
image

Disabled buttons when missing connectors
image

Error page can return home now
image


src/app/Authenticated.tsx
```diff
@@ -8,52 +8,39 @@ import Error from '../pages/Error';
...
-const Logs = lazy(() => import('../pages/Logs'));
-
 const Authenticated = () => {
     return (
         <Suspense fallback={<Skeleton animation="wave" />}>
             <Routes>
                 <Route element={<AppLayout />}>
                     <Route path="/" element={<Home />} />
                     <Route path="/dashboard" element={<Home />} />
-                    <Route path="/app">
```
travjenkins, line 35: Go rid of the `app` wrapper. Initially with how auth was handled this was useful but now it adds nothing.

src/app/Authenticated.tsx
```diff
@@ -8,52 +8,39 @@ import Error from '../pages/Error';
...
-                        <Route path="collections" element={<Collections />} />
-                        <Route path="captures" element={<Captures />}>
-                            <Route path="new" element={<CaptureCreation />} />
-                        </Route>
-                        <Route path="derivations" element={<Error />} />
+                    <Route path="collections" element={<Collections />} />
+                    <Route path="captures" element={<Captures />} />
+                    <Route path="capture">
+                        <Route path="create" element={<CaptureCreate />} />
+                    </Route>
```
travjenkins, line 31: Trying to make the UI paths a bit more "REST like" in nature.

travjenkins: create is nested because I know we'll need to add Edit soon

src/app/Authenticated.tsx
```diff
@@ -8,52 +8,39 @@ import Error from '../pages/Error';
...
-                                path="new"
-                                element={<NewMaterialization />}
-                            />
-                        </Route>
-                        <Route path="builds" element={<Builds />} />
-                        <Route path="admin/*" element={<Admin />}>
-                            <Route path="logs" element={<Logs />} />
-                            <Route path="alerts" element={<Alerts />} />
-                            <Route path="users" element={<Users />} />
-                        </Route>
```
travjenkins, line 55: These are not being used or built right now. Only keeping stuff in the nav that will be launched with the MVP

src/components/capture/create/SpecForm.tsx
```diff
@@ -53,16 +53,6 @@ function NewCaptureSpecForm(props: NewCaptureSpecFormProps) {
...
-    const uiSchema = useMemo(() => {
-        let response = null;
-
-        if (endpointSchema.type) {
-            response = generateCustomUISchema(endpointSchema);
-        }
-
-        return response;
-    }, [endpointSchema]);
-
```
travjenkins, line 65: This causes a weird issue with the collapsible group not having the data it needs. Still not fully sure why but we can figure that out in the future.

src/AppLayout.tsx
```diff
@@ -1,35 +1,42 @@
...
+    const isBelowMd = useMediaQuery(theme.breakpoints.down('md'));
+
     const [navigationOpen, setNavigationOpen] = useState(false);
     const [navWidth, setNavigationWidth] = useState<Widths>(Widths.RAIL);
 
     const toggleNavigationDrawer = () => {
-        setNavigationWidth(navigationOpen ? Widths.RAIL : Widths.FULL);
+        if (!isBelowMd) {
+            setNavigationWidth(navigationOpen ? Widths.RAIL : Widths.FULL);
+        }
```
travjenkins, line 23: We can skip for smaller screens because they don't have a rail to set the width to.

src/AppLayout.tsx
```diff
@@ -40,15 +47,13 @@ function AppLayout() {
...
                     open={navigationOpen}
                     onNavigationToggle={toggleNavigationDrawer}
-                    width={navWidth}
+                    width={isBelowMd ? Widths.FULL : navWidth}
                 />
             </Box>
-            <Box sx={{ gridArea: 'main', overflow: 'auto' }}>
-                <Suspense fallback={<Skeleton animation="wave" />}>
-                    <Outlet />
-                </Suspense>
```
travjenkins, line 49: Removed the suspense wrapper as it wasn't really useful here for right now.

src/components/capture/create/Store.ts
```diff
@@ -72,8 +72,16 @@ const useCaptureCreationStore = create<CaptureCreationState>(
             setDetails: (details) => {
                 set(
                     produce((state) => {
-                        if (state.details.image !== details.data.image) {
+                        if (
+                            details.data.image.length > 0 &&
+                            state.details.data.image !== details.data.image
+                        ) {
```
travjenkins, line 78: We only want to kick this stuff off if there is an image selected. Otherwise we can skip.

src/components/capture/create/Store.ts
```diff
@@ -72,8 +72,16 @@ const useCaptureCreationStore = create<CaptureCreationState>(
...
+                            const initState = getInitialStateData();
+
+                            state.links = initState.links;
                             state.links.connectorImage = details.data.image;
+
+                            state.spec = initState.spec;
```
travjenkins, line 84: When the connector changes we need to reset some values hence getting the init state here.

src/components/capture/create/index.tsx
```diff
@@ -106,19 +91,15 @@ function NewCaptureModal() {
...
-    const exitModal = () => {
+    const exit = () => {
         if (Object.keys(resourcesFromEditor).length > 0) {
             clearResourcesFromEditor();
         }
 
         resetState();
 
-        navigate('..');
+        navigate('/captures');
```
travjenkins, line 102: Create is no longer a child

src/components/capture/create/index.tsx
```diff
@@ -228,128 +191,81 @@ function NewCaptureModal() {
...
-                    onClick={handlers.close}
+        <PageContainer>
+            <Toolbar>
+                <Typography variant="h6" noWrap>
+                    <FormattedMessage id="captureCreation.heading" />
+                </Typography>
+
+                <Stack
+                    direction="row"
+                    alignItems="center"
```
travjenkins, line 202: Still not 100% sure if this is the best way to make a button group... but it works and is easy. Only weird part is the buttons need to be in reverse order in the JSX.

src/components/capture/create/index.tsx
```diff
@@ -228,128 +191,81 @@ function NewCaptureModal() {
...
-                    <CloseIcon />
-                </IconButton>
-            </DialogTitle>
-
-            {formSubmitError && (
-                <NewCaptureError title={formSubmitError.message} errors={[]} />
-            )}
-
-            <DialogContent dividers>
-                {activeStep === Steps.DETAILS_AND_SPEC ? (
```
travjenkins, line 265: Most of the code is identical. The main change here is removing all the `Dialog` stuff and the `Steps` stuff. But that altered the indentation and that made the changes look bigger than they are.

src/components/capture/create/index.tsx
```diff
@@ -228,128 +191,81 @@ function NewCaptureModal() {
...
-                ) : (
+
                     <Button
-                        onClick={handlers.test}
-                        disabled={formSubmitting}
-                        form={FORM_ID}
-                        size="large"
-                        type="submit"
+                        onClick={handlers.addToChangeSet}
+                        disabled={!catalogResponse || formSubmitting}
```
travjenkins, line 225: Don't want the "save" button until there is a catalog response to save.

src/lang/en-US.ts
```diff
@@ -107,4 +90,31 @@ const enUSMessages: ResolvedIntlConfig['messages'] = {
...
+const Materialization_Create: ResolvedIntlConfig['messages'] = {
+    'materializationCreation.heading': `New Materialization`,
+    'materializationCreation.instructions': `To get started please provide a unique name, select an endpoint type, and select the Collections you want to Materialize. After that you can review the YAML before saving.`,
+};
+
+const enUSMessages: ResolvedIntlConfig['messages'] = {
+    ...CommonMessages,
+    ...Capture_Create,
+    ...Materialization_Create,
+};
```
travjenkins, line 118: This is something we've talked about before. Trying it out on some of the pages. If we like it we can use it everywhere. This does allow us to easily reference common messages and stuff like that.

src/pages/Captures.tsx
```diff
@@ -1,12 +1,35 @@
...
+import { NavLink } from 'react-router-dom';
 
 const Capture = () => {
+    const [captureList] = useState([]);
+    const [isLoading] = useState(false);
+
     return (
         <PageContainer>
-            <CaptureRoot />
-            <Outlet />
```
travjenkins, line 9: Since new capture is no longer a child we don't need the `Outlet` and we can just make the `CaptureRoot` directly handled here.

src/pages/Error.tsx
```diff
@@ -1,12 +1,14 @@
...
+import { NavLink } from 'react-router-dom';
 
 const Error = () => {
     return (
         <PageContainer>
             <Alert severity="error">
                 <AlertTitle>Error</AlertTitle>
-                This path does not exist.
+                <Box>This path does not exist.</Box>
+                <NavLink to="/">Return to homepage</NavLink>
```
travjenkins, line 11: Added a link on this page so folks don't get stuck while we dev on a Staging environment.

src/services/jsonforms.ts
```diff
@@ -146,10 +146,9 @@ const generateUISchema = (
     }
 
     const types = deriveTypes(jsonSchema);
-    if (types.length === 0) {
-        // TODO - this was returning null initially but the compiler complained so I put this in here to make it be nice.
-        return { type: 'Type Missing on UI Schema' };
-    }
+    // if (types.length === 0) {
+    //   return null;
+    // }
```
travjenkins, line 151: Instead of changing code that is being brought in from 3rd parties if we don't need to worry about something (like getting sent crappy schemas) I think commenting stuff out is best. Makes it a bit easier to compare to the original when the time comes.

package.json
```diff
@@ -15,7 +13,8 @@
         "lint": "eslint --ext js,jsx,ts,tsx .",
         "lint:fix": "eslint --fix --ext js,jsx,ts,tsx .",
         "lint-staged": "lint-staged --config .lint-staged.js",
-        "typecheck": "tsc"
+        "typecheck": "tsc",
+        "__build_deploy__": "npm run build; serve -s build"
```
travjenkins, line 17: Named weird as it is just for testing.

src/components/shared/PageContainer.tsx
```diff
@@ -5,7 +5,7 @@ import useNotificationStore, {
 } from 'stores/NotificationStore';
 
 interface PageContainerProps {
-    children: ReactNode;
+    children: ReactNode | ReactNode[];
```
travjenkins, line 8: This makes it so we don't have to wrap pages in `<></>`

src/utils/store-utils.ts
```diff
@@ -0,0 +1,9 @@
+import { devtools } from 'zustand/middleware';
+
+// TODO - add in Immer middleware
+// https://github.com/pmndrs/zustand#middleware
+// https://github.com/pmndrs/zustand/blob/main/tests/middlewareTypes.test.tsx
+
+export const devtoolsInNonProd = (process.env.NODE_ENV === 'production'
+    ? (fn: any) => fn
+    : devtools) as unknown as typeof devtools;
```
travjenkins, line 9: This is a quick and dirty wrapper for including devtools only for non prod. Eventually we should add some common global middleware and maybe even wrap the create function here. But that will take some effort from a typing standpoint.

.env
```diff
@@ -1,3 +1,12 @@
...
+PORT=8080
+
+# These endpoints can be changed live on the site by updating the window object
+# window.Estuary = { api_endpoint: "", auth_url: "" };
+REACT_APP_API_BASE_URL=https://api.estuary.dev/v1
+REACT_APP_AUTH_BASE_URL=https://auth.estuary.dev
+
+# These settings require an application restart to enable right now
+REACT_APP_SHOW_OIDC_LOGIN=true
+REACT_APP_SHOW_OIDC_LOGIN_GOOGLE=true
```
travjenkins, line 12: We'll need to eventually think of a system on how we're gonna enable 3rd party stuff in the system better than this.

src/components/capture/create/DetailsForm.tsx
```diff
@@ -100,38 +104,39 @@ function NewCaptureDetails(props: NewCaptureDetailsProps) {
 
                     return previous;
                 });
+                setHasConnectors(true);
```
travjenkins, line 107: This defaults to `false` so only have to set to true if it is true.

src/components/menus/UserMenu.tsx
```diff
@@ -9,23 +10,32 @@ import IconMenu from './IconMenu';
 
 const UserMenu = () => {
     const { logout, user } = useAuth();
+    const userName = user?.ext.displayName;
+    const email = user?.ext.email;
```
travjenkins, line 14: We are no longer just passing around a string when the user is logged in. So fetching more details.

src/components/menus/UserMenu.tsx
```diff
@@ -9,23 +10,32 @@ import IconMenu from './IconMenu';
...
                         <AccountCircleIcon fontSize="small" />
                     </ListItemIcon>
-                    {user}
+                    {userName}
+                </MenuItem>
+                <MenuItem>
+                    <ListItemIcon>
+                        <EmailIcon fontSize="small" />
+                    </ListItemIcon>
+                    {email}
```
travjenkins, line 38: Added email for now. Probably need to add Org as well in the future.

src/context/Auth.tsx
```diff
@@ -1,27 +1,44 @@
...
+}
+
+export function checkTokens(tokens: AuthTokenResponseReduced) {
+    let response;
+
+    if (isFuture(fromUnixTime(tokens.expires))) {
+        response = tokens;
+    } else {
+        response = null;
+    }
```
travjenkins, line 24: Need to check if the token expired. The new endpoint returns unit time so using the handy date-fns library again.

src/context/Auth.tsx
```diff
@@ -38,46 +55,23 @@ export const AuthProvider = (props: any) => {
...
         status,
-    } = useAsync<string | null>();
+    } = useAsync<AuthTokenResponseReduced | null>();
 
     React.useEffect(() => {
         const appDataPromise = bootstrapUser();
         run(appDataPromise);
     }, [run]);
 
-    const login = useCallback(
```
travjenkins, line 52: We no longer have local log in.

src/endpoints/auth.ts
```diff
@@ -0,0 +1,64 @@
...
+    firstName: string;
+    lastName: string;
+    locale: string;
+    orgs: string[];
+}
+
+export interface AuthTokenResponseReduced
+    extends Pick<AuthTokenResponse, 'accessToken' | 'expires' | 'IDToken'> {
+    ext: Ext;
+}
```
travjenkins, line 32: Not sure what we want to call this and I do not care too much. Basically it is for objects that are returned from the server with more stuff than we need... so I called it `Reduced`

src/endpoints/auth.ts
```diff
@@ -0,0 +1,64 @@
...
+                            orgs: ['example.org'],
+                        },
+                        expires: getUnixTime(
+                            add(new Date(), {
+                                years: 1,
+                            })
+                        ),
+                        IDToken: 'id_token_value',
+                    });
+                });
```
travjenkins, line 57: Hardcoding for now

src/services/auth.ts
```diff
@@ -1,93 +1,48 @@
...
-
-            response = { session: JSON.parse(session) };
-
-            if (user) {
-                response.user = JSON.parse(user);
-            }
-        }
-
-        return response;
-    },
```
travjenkins, line 40: We now just save the reduced token response.

src/services/auth.ts
```diff
@@ -1,93 +1,48 @@
...
-                            })
-                            .catch((accountError) => {
-                                reject(accountError);
-                            });
-                    })
-                    .catch((error) => {
-                        reject(error);
-                    });
-            }
-        );
```
travjenkins, line 87: We no longer sign users in

src/services/client.ts
```diff
@@ -27,10 +28,13 @@ export const client = <Response, Request = {}>(
 
     config.headers = headersInit;
 
+    // TODO: probably need to remove this for production
+    const API_ENDPOINT = getAPIPath();
```
travjenkins, line 32: Made a helper to go fetch the api path dynamically

src/utils/env-utils.ts
```diff
@@ -0,0 +1,30 @@
+declare global {
+    interface Window {
+        Estuary: {
+            api_endpoint: string | null;
+            auth_url: string | null;
+        } | null;
+    }
+}
```
travjenkins, line 8: Setting up settings that can be provided in the window object

src/utils/test-utils.tsx
```diff
@@ -6,38 +6,42 @@ import { AuthContext } from 'context/Auth';
 import AppContent from 'context/Content';
 import AppRouter from 'context/Router';
 import AppTheme from 'context/Theme';
+import { add, getUnixTime } from 'date-fns';
 import { ReactElement } from 'react';
-import { auth, AuthDetails } from 'services/auth';
+import { auth } from 'services/auth';
+
+const getMockTokenReduced = (username: string) => {
```
travjenkins, line 13: Updates to the tests to match the new endpoints.

jgraettinger:
Can this be landed? I'm going to work off of it for prototyping API calls

src/components/capture/create/DetailsForm.tsx
```diff
@@ -1,70 +1,69 @@
...
+                },
+                name: {
+                    description: intl.formatMessage({
+                        id: 'captureCreation.name.description',
+                    }),
+                    maxLength: 1000,
+                    minLength: 3,
+                    pattern: '^[a-zA-Z0-9_.-]*/[a-zA-Z0-9_.-]+$',
+                    type: 'string',
+                },
```
travjenkins, line 60: Instead of handling stuff "after the server call" we are just memoizing the schema and passing the connectors from the parent.

src/components/capture/create/Spec.tsx
```diff
@@ -0,0 +1,64 @@
...
+                id,
+                endpoint_spec_schema, 
+                documentation_url
+            `,
+            filter: (query) => query.eq('id', connectorImage),
+            count: 'exact',
+        },
+        [connectorImage]
+    );
+    const { data: connector, error } = useSelectSingle(tagsQuery, {});
```
travjenkins, line 40: Parent wrapper for the spec form sections. This then passes down the needed details into the children.

src/components/capture/create/Store.ts
```diff
@@ -130,13 +99,13 @@ const useCaptureCreationStore = create<CaptureCreationState>(
...
-            setHasConnectors: (val) => {
+            setConnectors: (val) => {
                 set(
                     produce((state) => {
-                        state.hasConnectors = val;
+                        state.connectors = val;
                     }),
                     false,
-                    'Form has connectors'
+                    'Caching connectors response'
```
travjenkins, line 108: This was a temporary idea... will remove later now that we have SWR

src/endpoints/auth.ts
```diff
@@ -0,0 +1,45 @@
+import { client } from 'services/client';
+import { getAuthPath } from 'utils/env-utils';
+
+export interface AuthTokenResponse {
```
travjenkins, line 4: This stuff is all probably going to go away with Supabase.

src/hooks/useAsyncHandler.ts
```diff
@@ -27,23 +27,26 @@ const useSafeDispatch = (dispatch: Function) => {
...
     setError: (setErrorResponse: any) => any;
     status: States;
 }
 
-function useAsync<T>(): UseAsyncResponse<T, undefined>;
-function useAsync<T>(initialData: T): UseAsyncResponse<T, T>;
-function useAsync<T>(initialData?: T): UseAsyncResponse<T, undefined> {
+function useAsyncHandler<T>(): Response<T, undefined>;
+function useAsyncHandler<T>(initialData: T): Response<T, T>;
+function useAsyncHandler<T>(initialData?: T): Response<T, undefined> {
```
travjenkins, line 49: Renamed because the library `reactUse` has a component called `useAsync` and the auto import kept choosing that one.

src/hooks/useConnectorTags.ts
```diff
@@ -0,0 +1,41 @@
...
+                        .select(
+                            `endpoint_spec_schema, documentation_url, connectors(image_name)`
+                        )
+                        .eq('id', imageId)
+                        .limit(1)
+                        .single()
+                ).then(generateResponse)
+            );
+        }
+    }, [imageId, run]);
```
travjenkins, line 36: This was one approach - but honestly think we'll end up just having queries in components.

src/hooks/useSupabase.ts
```diff
@@ -0,0 +1,38 @@
+import { useAsyncHandler } from 'hooks/useAsyncHandler';
+import { isEmpty } from 'lodash';
+import { useCallback, useEffect } from 'react';
+import { supabase } from 'services/supabase';
+
+function useSupabase(query: (params?: any) => any, queryParams?: any) {
```
travjenkins, line 6: This pattern will be removed now that we have supabase-swr

src/stores/SupabaseCache.ts
```diff
@@ -0,0 +1,36 @@
+import produce from 'immer';
```
travjenkins, line 1: Will get removed.

src/services/supabase.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export const supabase = createClient(
+    supabaseSettings.url,
+    supabaseSettings.anonKey
+);
+
+export const DEFAULT_INTERVAL = 500;
+
+export const callSupabase = (
```
travjenkins, line 27: Will get removed.

README.md
```diff
@@ -6,7 +6,7 @@ The web UI for Estuary Flow.
 
 You'll need to make sure the UI ahas the depencies installed. For the UI this should be straight forward and running `npm install` in the UI root directory.
```
kiahna-tucker, line 7: There is are typos in the first sentence of this paragraph: _You'll need to make sure the UI **has** the **dependencies** installed._

src/app/Authenticated.tsx
```diff
@@ -25,6 +25,7 @@ const Authenticated = () => {
             <Routes>
                 <Route element={<AppLayout />}>
                     <Route path="/" element={<Home />} />
+                    <Route path="/test/builds" element={<Builds />} />
```
kiahna-tucker, line 28: Why was this route reintroduced?

src/components/capture/create/DetailsForm.tsx
```diff
@@ -1,70 +1,69 @@
...
+}
 
 const stateSelectors = {
     formData: (state: CaptureCreationState) => state.details.data,
     setDetails: (state: CaptureCreationState) => state.setDetails,
-    setHasConnectors: (state: CaptureCreationState) => state.setHasConnectors,
+    setConnectors: (state: CaptureCreationState) => state.setConnectors,
 };
 
 function NewCaptureDetails(props: NewCaptureDetailsProps) {
```
kiahna-tucker, line 27: **Pattern Proposal:** Whenever possible, props should be destructured in the argument list of the function header.

src/app/Unauthenticated.tsx
```diff
@@ -1,7 +1,53 @@
...
+import { getLoginSettings } from 'utils/env-utils';
 
 const Unauthenticated = () => {
-    return <Login />;
+    const loginSettings = getLoginSettings();
+
+    return (
+        <Grid
+            container
+            spacing={0}
```
kiahna-tucker, line 14: The default value for the `spacing` property is 0. Is there a reason to set it explicitly here?

src/app/Unauthenticated.tsx
```diff
@@ -1,7 +1,53 @@
...
+            direction="column"
+            sx={{
+                alignItems: 'center',
+                height: '100vh',
+                justifyContent: 'center',
+            }}
+        >
+            <Topbar isNavigationOpen={false} />
```
kiahna-tucker, line 22: It is my understanding that a `Grid` container should always wrap `Grid` items. Therefore, should `Topbar` be encased in a `Grid` item?

src/app/Unauthenticated.tsx
```diff
@@ -1,7 +1,53 @@
...
+            <Grid item xs={3}>
+                <Card elevation={24} sx={{ maxWidth: 400, minHeight: 300 }}>
+                    <CardContent>
+                        <Typography
+                            gutterBottom
+                            variant="h6"
+                            component="h2"
+                            sx={{ paddingTop: '1rem', textAlign: 'center' }}
```
kiahna-tucker, line 30: Interesting to see `rem` used here. Since MUI is innately pixel-centric, I am curious if we should use pixels by default ourselves or adjust MUI's default metric in favor of another.

src/components/capture/create/Spec.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface ConnectorTag {
+    connectors: {
+        image_name: string;
+    };
+    id: string;
+    endpoint_spec_schema: string;
+    documentation_url: string;
+}
+
+interface Props {
```
kiahna-tucker, line 17: Are you leaning towards a generic prop interface name?

src/components/capture/create/Store.ts
```diff
@@ -4,13 +4,6 @@ import { isEqual } from 'lodash';
 import { devtoolsInNonProd } from 'utils/store-utils';
```
kiahna-tucker, line 4: There are anonymous actions defined in this file. For the ease of debugging, each action should be accompanied by a descriptive string.

src/components/capture/create/index.tsx
```diff
@@ -52,47 +53,73 @@ const selectors = {
         state.spec.errors,
     ],
     specFormData: (state: CaptureCreationState) => state.spec.data,
-    disoverLink: (state: CaptureCreationState) =>
-        state.links.discovered_catalog,
-    hasConnectors: (state: CaptureCreationState) => state.hasConnectors,
+    connectors: (state: CaptureCreationState) => state.connectors,
 };
 
+interface ConnectorTag {
```
kiahna-tucker, line 59: There is a `ConnectorTag` interface defined in _src/components/capture/create/Spec.tsx_ as well. Looks like you are leaning toward using generic names for interfaces that are local to files. Is that an accurate conclusion?

src/components/capture/create/index.tsx
```diff
@@ -52,47 +53,73 @@ const selectors = {
...
         selectors.errors
     );
     const specFormData = useCaptureCreationStore(selectors.specFormData);
-    const disoverLink = useCaptureCreationStore(selectors.disoverLink);
     const resetState = useCaptureCreationStore(selectors.resetState);
     const hasChanges = useCaptureCreationStore(selectors.hasChanges);
-    const hasConnectors = useCaptureCreationStore(selectors.hasConnectors);
 
     // Form props
     const [showValidation, setShowValidation] = useState(false);
```
kiahna-tucker, line 109: Is `useState` intentionally not typed in this code block?

src/components/menus/UserMenu.tsx
```diff
@@ -1,36 +1,60 @@
...
                     </ListItemIcon>
-                    {user}
+                    {userName}
                 </MenuItem>
+                <MenuItem>
+                    <ListItemIcon>
+                        <EmailIcon fontSize="small" />
+                    </ListItemIcon>
+
+                    <Stack spacing={0}>
```
kiahna-tucker, line 46: The default value for the `spacing` property is 0. Is there a reason to set it explicitly here?

src/context/Data.tsx
```diff
@@ -0,0 +1,13 @@
+import { supabase } from 'services/supabase';
```
kiahna-tucker, line 1: This is an ambiguous file name.

src/endpoints/connectors.ts
```diff
@@ -1,8 +1,9 @@
 import { client } from 'services/client';
+import { callSupabase, supabase } from 'services/supabase';
 import { BaseResponse } from 'types';
 
 export enum ConnectorTypes {
```
kiahna-tucker, line 5: Out of curiosity, if _capture_ and _materialization_ are connector types, what are _Hello World_ and  _Postgres_ classified as?

src/pages/Captures.tsx
```diff
@@ -1,14 +1,15 @@
...
 import PageContainer from 'components/shared/PageContainer';
-import { useState } from 'react';
+import CapturesTable from 'components/tables/Captures';
 import { NavLink } from 'react-router-dom';
 
-const Capture = () => {
-    const [captureList] = useState([]);
-    const [isLoading] = useState(false);
+const boxStyling: SxProps<Theme> = {
+    marginBottom: 2,
```
kiahna-tucker, line 8: I use the shorthand names for these properties. Might be worthwhile to adopt that practice universally.

src/components/capture/create/Spec.tsx
travjenkins: yeah... makes it easier and you never really export these. Will change others as we have time.

src/app/Authenticated.tsx
travjenkins: So you could hit the page if needed for any work you were doing.

src/app/Unauthenticated.tsx
```diff
@@ -1,7 +1,53 @@
...
+import Topbar from 'components/header/Topbar';
+import { FormattedMessage } from 'react-intl';
+import { supabase } from 'services/supabase';
+import { getLoginSettings } from 'utils/env-utils';
 
 const Unauthenticated = () => {
-    return <Login />;
+    const loginSettings = getLoginSettings();
+
+    return (
```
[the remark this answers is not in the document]
travjenkins: https://github.com/estuary/ui/pull/71 handled this

src/app/Unauthenticated.tsx
travjenkins: Nope... I thought the default was 1

src/app/Unauthenticated.tsx
travjenkins: We could... doesn't impact the UI from what I can tell.

src/app/Unauthenticated.tsx
travjenkins: With exception of percentages we should stick with just using the numbers. This is from some older code that I just copied.

src/components/capture/create/DetailsForm.tsx
travjenkins: This sounds good to me.

src/components/capture/create/Store.ts
travjenkins: I do not know what you're talking about.

src/components/tables/Connectors.tsx
```diff
@@ -8,33 +8,55 @@ import {
     TableRow,
     Typography,
 } from '@mui/material';
-import useConnectors from 'hooks/useConnectors';
+import Error from 'components/shared/Error';
 import { FormattedDate, FormattedMessage, useIntl } from 'react-intl';
+import { Tables } from 'services/supabase';
+import { useQuery, useSelect } from 'supabase-swr';
+
+interface ConnectorTag {
```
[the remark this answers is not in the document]
travjenkins: No reason to expose it. Each query will have it's own type to go along with it.

After a while we might see several components have the same query and we can move all that stuff out at the same time.

src/context/Data.tsx
travjenkins: That is on purpose. That way if we change out Supabase or Supabase SWR we don't need to rename this.

src/endpoints/connectors.ts
travjenkins: That's a good question... I am not totally sure. Like I normally think of them as "endpoints". We should ask this to the larger group.

src/components/capture/create/Store.ts
kiahna-tucker: There are days I wish GitHub adopted some of the functionality of Azure Repos. I am unable to leave a comment anywhere in a given file, so it is difficult to highlight a proper example. `set` can have three arguments, the third argument dictates the string that appears in Redux Dev Tools. When that third argument is undefined, a vague `anonymous` appears in the dev tools. Currently, every `hasChanges` call is ambiguous in the tool.

PR #62, opened 2022-03-31
Add an npm script to regenerate Flow types

psFried:
### Changes

The file `flow_deps/flow.d.ts` contains Typescript type definitions for all the Flow catalog types. This commit adds a slightly easier way of updating those types to the latest version. This is achieved by adding `fetch-flow.sh`, which downloads the most recent build of `flowctl`, and then calling that in an npm script to re-generate the types.

I also renamed `catalog` to `Catalog`, to use the idiom of types having upper case names. That type seems to not be used anywhere by name, so there were no other code changes.

I removed `flow_deps` from `typeRoots` in `tsconfig`. My understanding is that it wasn't necessary for it to be there in the first place, but please LMK if that's not right. When it was present, it prevented me from putting the `flow-bin/` directory under `flow_deps/` because it expected it to be a typescript module. The `package.json` was changed to make inclusion of `flow.d.ts` explicit, to avoid issues with other files being included in the typescript build accidentally.

### Tests

I manually tested this and confirmed with `diff` that the only change to `flow.d.ts` is the rename from `catalog` to `Catalog`.

### Screenshots

:framed_picture:

### DIFF
--- a/flow_deps/.gitignore
+++ b/flow_deps/.gitignore
@@ -1,0 +1,1 @@
+flow-bin/
--- a/flow_deps/README.MD
+++ b/flow_deps/README.MD
@@ -1,13 +1,13 @@
 ### What is this folder?
+
 This folder houses the files we pull in from the [estuary/flow](https://github.com/estuary/flow/) project. 
 
 The types should be used along with the server responses.
 
 ### How do we update the types?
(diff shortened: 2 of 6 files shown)


travjenkins (approved):
screenshot sus... but LGTM

PR #63, opened 2022-04-01
Preparing for Deployment

travjenkins:
For details please look at this PR: https://github.com/estuary/ui/pull/55

Initially we were trying to clean up some stuff in the UI to prepare for launching to staging. This was basic UI tweaks and making sure stuff wasn't hanging around in the UI visually, provide less deadends, etc. That eventually turned into changing how we handle auth and api calls... which blew up the scope and time to land the PR from above. 

So this branch is that PR rolled back to https://github.com/estuary/ui/pull/55/commits/9428a0a717c654e7aea5e6275ea4bf6de9d3a33a with the following two changes applied:

1. Updating prettier to no fix issues (only important for Github actions)
2. Adding a backdoor link to builds


src/stores/SchemaEditorStore.ts
```diff
@@ -11,7 +11,7 @@ export interface SchemaEditorState {
 }
 
 const useSchemaEditorStore = create<SchemaEditorState>(
```
kiahna-tucker, line 13: Should this file adopt the change in naming convention applied to the related component, _CatalogEditor.tsx_?

src/lang/en-US.ts
```diff
@@ -1,7 +1,7 @@
 // We DO NOT need this file sorted as it makes it easier for folks to update/see changes
 import { ResolvedIntlConfig } from 'react-intl/src/types';
 
-const enUSMessages: ResolvedIntlConfig['messages'] = {
+const CommonMessages: ResolvedIntlConfig['messages'] = {
```
kiahna-tucker, line 4: Was the deviation from camel case variable names intentional in this file? If so, how did you arrive at that conclusion?

src/pages/Captures.tsx
```diff
@@ -1,12 +1,35 @@
-import CaptureRoot from 'components/capture/Root';
+import AddIcon from '@mui/icons-material/Add';
+import { Button, Toolbar } from '@mui/material';
+import CatalogList from 'components/CatalogList';
```
kiahna-tucker, line 3: Would the _src/components/shared_ directory be a better location for the highlighted file?

kiahna-tucker (approved):
Approved with suggestions. Given this PR is a subset of a greater wave of changes, this review was expedited and all comments can be addressed outside its confines.

PR #77, opened 2022-04-07
Travjenkins/demo tweaks3

travjenkins:
### Changes

Clean out old files not being used anymore
Remove logs stopping themselves
Displaying connector details and not the image tag in Capture Create
Captures only displays successful captures

### Tests

Manual

### Screenshots

Captures only shows successful discovers
image

Capture create now show the details
image


kiahna-tucker (approved):
Approved.

src/components/Logs.tsx
```diff
@@ -11,11 +12,10 @@ interface Props {
 
 const NEW_LINE = '\r\n';
 
-function Logs(props: Props) {
-    const { token, defaultMessage } = props;
+function Logs({ token, defaultMessage }: Props) {
+    const supabaseClient = useClient();
```
travjenkins, line 16: Since we're running our calls through a SWR package we should probably use their client hook to be safe.

src/components/Logs.tsx
```diff
@@ -27,7 +27,7 @@ function Logs(props: Props) {
 
     useInterval(
         async () => {
-            const { data: viewLogsResponse } = await supabase
+            const { data: viewLogsResponse } = await supabaseClient
```
travjenkins, line 30: Renamed to make it clear we use the client

src/components/menus/IconMenu.tsx
```diff
@@ -23,21 +29,21 @@ const IconMenu = (props: IconMenuProps) => {
         setAnchorEl(null);
     };
 
-    const id = `${props.identifier}-button`;
```
travjenkins, line 26: This is interesting that this wasn't caught by the linters... I guess it is somewhat edge case?

src/components/tables/Captures/Empty.tsx
```diff
@@ -0,0 +1,49 @@
+import { Box, Typography } from '@mui/material';
+import ExternalLink from 'components/shared/ExternalLink';
+import { FormattedMessage, useIntl } from 'react-intl';
+
+function Empty() {
```
travjenkins, line 5: Not sure how I feel about this pattern... but might be useful in the future when we have a more fleshed out table. We could just pass an "Empty Component" to the table.

src/components/tables/Captures/Rows.tsx
```diff
@@ -0,0 +1,30 @@
+import { TableCell, TableRow } from '@mui/material';
+import { FormattedDate } from 'react-intl';
+
+interface Props {
+    discovers: any[];
+    styling: any;
+}
+
+function Rows({ discovers, styling }: Props) {
```
travjenkins, line 9: Same as the empty one... not sure how I feel about the pattern but kind of like it. Only thing that sucks is the disconnect between the rows and columns specified in the parent.

src/components/tables/Captures/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { FormattedMessage, useIntl } from 'react-intl';
+import { TABLES } from 'services/supabase';
+import { useQuery, useSelect } from 'supabase-swr';
+
+interface Discovers {
+    'capture_name': string;
+    'updated_at': string;
+    'job_status->>type': string;
+    'id': string;
+}
```
travjenkins, line 23: Need to eventually figure a pattern out for these. We need these types to help with getting safer filters and other query builders. However, when you do something like parse the JSON in the query the data that comes back does not match the query. 

For example `job_status->>type` comes back in the data just as `type`... which makes sense. This is why I am still leaving the `data` prop on the response as an any for now.

Issue #78, opened 2022-04-07
Allow for copying error messages from the UI


dyaffe:
**Background**
Currently when an error message is created on the capture page of the UI, it's not possible to copy it.

**Details**

Create a postgres capture with information that won't work to connect.  Click "test config" and note the error message.  Try to copy the error message and note that there's no way to do so.

Users will want the ability to copy / paste errors so that they can more easily communicate them with us.

travjenkins:
This is done. We changed the library to output the logs to us Monaco (same editor we use for editing discover results). There is a new issue open to represent the work needed to add more features to the editor to make it feel a lot more like a "log viewer" and less like an "editor showing logs".

PR #79, opened 2022-04-08
Logs and user menu UX fixes

travjenkins:
## Changes

Moved the logs display to a Monaco editor. This is not nearly finished yet - but it allows folks to copy and paste and interact with the logs better.

User menu was not displaying when logged in locally with email. Noticed that it was not returning the `user_metadata` object. So Added some logic to handle that and display the info we do have.

## Tests

Manual

## Issues

https://github.com/estuary/ui/issues/78 

## Content

None

## Screenshots

image


src/components/menus/UserMenu.tsx
```diff
@@ -7,16 +7,28 @@ import Divider from '@mui/material/Divider';
...
+            userName = user.user_metadata.full_name;
+            email = user.user_metadata.email;
+            emailVerified = user.user_metadata.email_verified;
+            avatar = user.user_metadata.avatar_url;
+        } else {
+            userName = user.email;
+            email = user.email;
+            emailVerified = false;
+        }
+    }
```
travjenkins, line 31: Right now this is the only place we need to fetch details. Eventually we'll probably need to make a service that'll fetch the proper details. But keeping it just inline for now.

src/components/capture/create/LogDialog.tsx
```diff
@@ -29,10 +31,14 @@ function LogDialog({ open, token, defaultMessage, actionComponent }: Props) {
             </DialogTitle>
             <DialogContent
                 sx={{
-                    height: 300,
+                    height: logHeight + 25,
```
travjenkins, line 34: For the DialogContent only the wrapper needs to be a hair bigger.

src/components/Logs.tsx
```diff
@@ -45,17 +77,22 @@ function Logs({ token, defaultMessage }: Props) {
     );
 
     return (
-        <LazyLog
-            extraLines={1}
-            stream={true}
-            text={logs
+        <Editor
+            height={`${height ?? 200}px`}
```
travjenkins, line 81: Monaco requires a height so allowing a default but really we should probably not use it. Maybe we can just end up throwing an error if no height is provided.

src/components/Logs.tsx
```diff
@@ -45,17 +77,22 @@ function Logs({ token, defaultMessage }: Props) {
...
+            defaultLanguage=""
+            theme={theme.palette.mode === 'light' ? 'vs' : 'vs-dark'}
+            options={{
+                lineNumbers: 'off',
+                readOnly: true,
+                scrollBeyondLastLine: false,
+            }}
```
travjenkins, line 88: Eventually we might need a wrapper around the editor that provides the basic default options we always want.

src/components/Logs.tsx
```diff
@@ -1,30 +1,62 @@
...
             }),
-        '...',
     ]);
 
+    const handlers = {
+        change: () => {
+            if (editorRef.current) {
+                const lineCount =
+                    editorRef.current.getModel()?.getLineCount() ?? 0;
+                editorRef.current.revealLine(lineCount);
```
travjenkins, line 52: Just trying to keep the last line loaded in view. Does not always work perfectly - but we can make it better later.

src/components/Logs.tsx
```diff
@@ -1,30 +1,62 @@
...
+const generateLogLine = (logData: any) => {
+    const parsedText = parse(logData.log_line);
+    const mergedText = parsedText.spans
+        .map((span) => {
+            return span.text;
+        })
+        .join('');
+
+    return mergedText;
+};
```
travjenkins, line 28: This will need a lot more work. The ansi parser here returns and tells us how to style specific pieces of text. We eventually need to get that stuff passed/formatted for Monaco.

kiahna-tucker (approved):

PR #81, opened 2022-04-11
Travjenkins/supa unit test

travjenkins:
## Changes / Tests

Getting very basic tests working again.


package.json
```diff
@@ -93,7 +95,11 @@
         "flow_deps"
     ],
     "jest": {
-        "resetMocks": false
+        "resetMocks": false,
+        "testTimeout": 10000,
+        "moduleNameMapper": {
+            "supabase-swr": "<rootDir>/node_modules/supabase-swr/dist/index.umd.js"
+        }
```
travjenkins, line 102: We need to manually set this up otherwise jest cannot find this module... not totally sure why.

src/utils/test-utils.tsx
```diff
@@ -16,31 +22,42 @@ const goTo = (route?: string, name?: string) => {
     );
 };
 
-const waitForLoading = async () => {
-    return Promise.resolve();
-};
```
travjenkins, line 21: This was a stand in and not working well. So removing. Just going to stick with each test having it's own wait

src/utils/test-utils.tsx
```diff
@@ -16,31 +22,42 @@ const goTo = (route?: string, name?: string) => {
...
+        draft.email = `${username}@example.org`;
+        draft.email_verified = true;
+        draft.full_name = `Full ${username}`;
+        draft.iss = 'https://api.example.org';
+        draft.name = username;
+        draft.picture = `https://example.org/picture/${username}`;
+        draft.preferred_username = username;
+        draft.provider_id = '102885624013818367764';
+        draft.sub = '102885624013818367764';
+        draft.user_name = username;
```
travjenkins, line 39: This is based on the google user metadata object.

src/utils/test-utils.tsx
```diff
@@ -16,31 +22,42 @@ const goTo = (route?: string, name?: string) => {
...
+        draft.email = `${username}@example.org`;
+        draft.email_verified = true;
+        draft.full_name = `Full ${username}`;
+        draft.iss = 'https://api.example.org';
+        draft.name = username;
+        draft.picture = `https://example.org/picture/${username}`;
+        draft.preferred_username = username;
+        draft.provider_id = '00000000000000000000_provider';
+        draft.sub = '00000000000000000000_sub';
+        draft.user_name = username;
```
travjenkins, line 39: Right now we just use the username in as many things as it makes sense to so we can check for them in the UI. Eventually we can add ability to manually set more variables in the future if it is needed.

src/utils/test-utils.tsx
```diff
@@ -16,31 +22,42 @@ const goTo = (route?: string, name?: string) => {
...
+const MockProviders = ({ children, username }: any) => {
     const mockClient = createClient('http://mock.url', 'MockSupabaseAnonKey');
+    const mockAuthSession = mockDeep<AuthSession>();
 
-    const MockClient = createContext({});
+    if (username) {
+        mockAuthSession.user = mockDeep<User>();
+        mockAuthSession.user.user_metadata = generateMockUserMetadata(username);
+        Auth.useUser = () => mockAuthSession;
+    }
```
travjenkins, line 53: Only need to mock out if a username is passed in. That's how we know if we should log the user in.

src/utils/test-utils.tsx
```diff
@@ -16,31 +22,42 @@ const goTo = (route?: string, name?: string) => {
...
-                useClient: () => mockClient,
-            }}
-        >
-            {children}
-        </MockClient.Provider>
+        <SwrSupabaseContext.Provider value={mockClient}>
+            <Auth.UserContextProvider supabaseClient={mockClient}>
+                {children}
+            </Auth.UserContextProvider>
+        </SwrSupabaseContext.Provider>
```
travjenkins, line 60: Using mocked values by setting them directly in the true providers.

src/utils/test-utils.tsx
```diff
@@ -53,27 +70,26 @@ const customRender = async (
...
+                </ThemeProvider>
+            </MockProviders>
         </AppContent>,
         {
             ...options,
         }
     );
 
-    await waitForLoading();
+    goTo(route, 'Test Page');
```
travjenkins, line 86: We wait for stuff in each test - so no need for waiting to load.

Putting the route after the router is created.

src/utils/test-utils.tsx
```diff
@@ -53,27 +70,26 @@ const customRender = async (
...
 
-    return view;
+    return {
+        user: userEvent.setup(),
+        view,
+    };
```
travjenkins, line 91: Returning a user so we can easily take user actions in tests when we need to.

kiahna-tucker (approved):

Issue #82, opened 2022-04-12
Integrate new Supabase API/Schema


travjenkins:
We have changed how the backend schema handles creation of entities. The UI needs to get wired up to this.

PR #83, opened 2022-04-13
Wiring up new API

travjenkins:
## Changes

Wiring up new API
Moving the editor a stand alone (ish) component
Some more error handling

There is a bunch more stuff to work out with new patterns. This is not really where we need to be - but it lets us use the UI now and we can figure the rest out later.

## Tests

Manual

## Issues

https://github.com/estuary/ui/issues/82

## Content

Some new messages for when the server cannot be reached

## Screenshott

N/A


src/components/draft/editor/MonacoEditor.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    .from(TABLES.DRAFT_SPECS)
+                    .update(newData)
+                    .match({
+                        draft_id: draftId,
+                        catalog_name: currentCatalog,
+                    })
+                    .then(
+                        () => {},
+                        () => {}
+                    );
```
travjenkins, line 74: Put changes in DB

src/components/draft/editor/MonacoEditor.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+                mutate()
+                    .then(() => {})
+                    .catch(() => {});
```
travjenkins, line 78: Inform the cache stuff has changed

src/components/draft/editor/Store.ts
```diff
@@ -0,0 +1,53 @@
...
+        { name: 'editor-state' }
+    )
+);
+
+export const editorStoreSelectors = {
+    draftId: (state: EditorStorState) => state.draftId,
+    setDraftId: (state: EditorStorState) => state.setDraftId,
+    currentCatalog: (state: EditorStorState) => state.currentCatalog,
+    setCurrentCatalog: (state: EditorStorState) => state.setCurrentCatalog,
+};
```
travjenkins, line 51: Selectors in the store seems to make more sense to me. Would like to add some typesctiping to handle that this should have all the same keys as the State.

src/stores/SchemaEditorStore.ts
```diff
@@ -1,47 +0,0 @@
-import { devtoolsInNonProd } from 'utils/store-utils';
```
travjenkins, line 1: No longer needed.

src/components/capture/create/index.tsx
```diff
@@ -74,21 +74,25 @@ const selectors = {
...
     notifications: {
         showNotification: (state: NotificationState) => state.showNotification,
     },
 };
 
+const notification: Notification = {
+    description: 'Your new capture is published and ready to be used.',
+    severity: 'success',
+    title: 'New Capture Created',
+};
```
travjenkins, line 86: Never changes - can be kept out of the component. Might even end up making as "library" of predefined notifications?

src/components/capture/create/index.tsx
```diff
@@ -135,130 +134,108 @@ function CaptureCreation() {
...
-
-        resetState();
-
-        navigate('/captures');
-    };
-
-    const discovers = {
-        done: (discoversSubscription: RealtimeSubscription) => {
+    const helpers = {
+        callFailed: (formState: any, subscription?: RealtimeSubscription) => {
```
travjenkins, line 142: Throwing some repeated code into helpers for now.

src/components/capture/create/index.tsx
```diff
@@ -135,130 +134,108 @@ function CaptureCreation() {
...
+                setFormState({
+                    status: CaptureCreationFormStatus.IDLE,
+                    exitWhenLogsClose: false,
+                    saveStatus: intl.formatMessage({
+                        id: 'captureCreation.status.failed',
+                    }),
+                    ...formState,
+                });
+            };
+            if (subscription) {
```
travjenkins, line 153: This allows us to use the same function for both when an AJAX call totally dies (and we don't have a chance to start a subscription) and when it comes back but with an error.

src/components/capture/create/index.tsx
```diff
@@ -135,130 +134,108 @@ function CaptureCreation() {
...
+                },
+                saveStatus: intl.formatMessage({
+                    id: 'captureCreation.status.failed',
+                }),
+            });
+        },
+    };
+
+    const waitFor = {
+        base: (query: any, success: Function, failureTitle: string) => {
```
travjenkins, line 192: Created a base waitFor that the other two share. We'll expand on this eventually.

src/components/capture/create/index.tsx
```diff
@@ -135,130 +134,108 @@ function CaptureCreation() {
...
 
-            return draftsSubscription;
+                    showNotification(notification);
+                },
+                'captureCreation.save.failedErrorTitle'
+            );
         },
     };
 
-    const prepareCatalogForSaving = () => {
```
travjenkins, line 248: We update the `draft_spec` table now and no longer need to build up a catalog to send.

src/components/capture/create/index.tsx
```diff
@@ -356,52 +328,75 @@ function CaptureCreation() {
...
+                    .from(TABLES.DRAFTS)
+                    .insert({
+                        detail: captureName,
+                    })
                     .then(
-                        (response) => {
-                            if (response.data) {
-                                setFormState({
-                                    logToken: response.data[0].logs_token,
+                        (draftsResponse) => {
```
travjenkins, line 337: Eventually stuff like below will probably get pulled out into a service where we can either pass callbacks or have the service return promises if we keep seeing this pattern.

src/components/draft/editor/FileSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const currentCatalog = useEditorStore(editorStoreSelectors.currentCatalog);
+
+    const draftSpecQuery = useQuery<DraftSpecFileList>(
+        TABLES.DRAFT_SPECS,
+        {
+            columns: DRAFT_SPECS_FILE_LIST_QUERY,
+            filter: (query) => query.eq('draft_id', draftId),
+        },
+        []
+    );
```
travjenkins, line 38: Just let the file selector query it's own list of files as the selector does not really care about the content of files.

src/components/draft/editor/MonacoEditor.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const draftSpecQuery = useQuery<DraftSpec>(
+        TABLES.DRAFT_SPECS,
+        {
+            columns: DRAFT_SPECS_QUERY,
+            filter: (query) =>
+                query
+                    .eq('draft_id', draftId)
+                    .eq('catalog_name', currentCatalog ? currentCatalog : ''),
+        },
+        [currentCatalog]
```
travjenkins, line 46: Like the file selector - just have this component go and fetch the data it needs. We'll eventually need to work on this more as I think the Monaco editor is doing some extra needless work with this approach.

src/components/draft/editor/MonacoEditor.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const { data: draftSpec, mutate } = useSelectSingle(draftSpecQuery, {
+        revalidateOnMount: true,
+    });
+
+    const editorRef = useRef<monacoEditor.editor.IStandaloneCodeEditor | null>(
+        null
+    );
+
+    const handlers = {
+        change: debounce(async () => {
```
travjenkins, line 57: Wait for the user to stop typing for a second before saving changes to the server.

src/components/draft/editor/Store.ts
```diff
@@ -0,0 +1,53 @@
+import produce from 'immer';
+import { devtoolsInNonProd } from 'utils/store-utils';
+import create from 'zustand';
+
+interface EditorStorState {
+    draftId: string | null;
+    setDraftId: (newVal: EditorStorState['draftId']) => void;
+
+    currentCatalog: string | null;
+    setCurrentCatalog: (newVal: EditorStorState['currentCatalog']) => void;
```
travjenkins, line 10: Might put the file list in here? Maybe... would just make setting the default file easier.

kiahna-tucker (approved):
Approved to merge changes ASAP. Will review tomorrow.

src/components/draft/editor/index.tsx
```diff
@@ -0,0 +1,21 @@
+import { Box } from '@mui/material';
+import EditorFileSelector from 'components/draft/editor/FileSelector';
```
kiahna-tucker, line 2: Why not go with `FileSelector`?

kiahna-tucker, line 2: Why not go with `FileSelector`?

src/lang/en-US.ts
```diff
@@ -128,6 +128,7 @@ const LoginPage: ResolvedIntlConfig['messages'] = {
 
 const CaptureCreate: ResolvedIntlConfig['messages'] = {
     'captureCreation.ctas.discover': `Test Config`,
+    'captureCreation.ctas.discoverAgain': `Regenerate Catalog`,
```
kiahna-tucker, line 131: Would `'captureCreation.ctas.rediscover'` work here?

src/components/draft/editor/FileSelector.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+function EditorFileSelector() {
+    const setCurrentCatalog = useEditorStore(
+        editorStoreSelectors.setCurrentCatalog
+    );
+    const currentCatalog = useEditorStore(editorStoreSelectors.currentCatalog);
+    const draftId = useEditorStore(editorStoreSelectors.draftId);
+    const { draftSpecs } = useDraftSpecs(draftId);
+
+    return (
```
kiahna-tucker, line 16: The conditionals in this return statement are a little challenging to follow. Could they be written as seen in the following example?

```ts
return draftId && (
  <List>
    { draftSpecs.length > 0 ? (<ComponentSeriesOne />) : (<ComponentSeriesTwo />) }
  </List>
);
```

src/lang/en-US.ts
travjenkins: I thought about that but wanted the key to feel more like it is a transformation on the key above and not a stand alone one... not sure if that even makes any sense though.

src/components/draft/editor/FileSelector.tsx
travjenkins: Nope - it is not the same thing. This is a weird side effect of out styling from linting and how it handled the null response. Since it is such a short line it will always move itself to a single line. So when I had it near the bottom it looked even more confusing.

src/components/draft/editor/index.tsx
travjenkins: The `EditorFileSelector` ? It does match as it follows the pattern of including the parent folder when the component name is pretty non-descriptive.

Issue #88, opened 2022-04-19
Reduce duplication in the routing


travjenkins:
The app currently has a lot of duplication when it comes to routing. We have the names of pages and paths hardcoded all over the place. Would be good to get those reduced.

PR #89, opened 2022-04-19
Expose route settings and cleanup

travjenkins:
## Changes

Remove lazy loading as we need to make the use of `suspense` in our codebase. We are a small enough app that it does not impact load times too much.

Added title changing ability to components.

Created a "public" object we use to populate routes.

Added capture create to the page not found list

## Tests

Manual

## Issues

[_Please list out any related issues (Here so issue is not auto closed by PR)_](https://github.com/estuary/ui/issues/88)

## Content

New content for Page titles

## Screenshots

Here are _some_ of the new titles
image

image

image

Capture Create now on page not found
image


.env
```diff
@@ -10,6 +10,9 @@ REACT_APP_SHOW_EMAIL_LOGIN=false
...
 
-# These settings control how much logging is done for SWR
+# Control how much logging is done for SWR
 REACT_APP_SWR_LOG_CACHE=false
-REACT_APP_SWR_LOG_REQUESTS=false
\ No newline at end of file
+REACT_APP_SWR_LOG_REQUESTS=false
+
+# This is the title displayed while the UI is initially loading in the browser
+REACT_APP_DEFAULT_PAGE_TITLE=Estuary Flow
```
travjenkins, line 18: Eventually we should be smart and pull a variable from the lang pack.... but this needs to be "hardcoded" into the index.html for right now

src/app/Authenticated.tsx
```diff
@@ -1,56 +1,111 @@
...
+    builds: {
+        title: 'routeTitle.builds',
+        path: '/test/builds',
+    },
+    capture: {
+        root: '/capture',
+        create: {
+            title: 'routeTitle.captureCreate',
+            path: `create`,
+            fullPath: '/capture/create',
```
travjenkins, line 28: Pages that are nested need their full path included so pages can link.

src/app/Authenticated.tsx
```diff
@@ -1,56 +1,111 @@
...
+    },
+    pageNotFound: {
+        title: 'routeTitle.error.pageNotFound',
+        path: '*',
+    },
+};
 
 const Authenticated = () => {
     return (
-        <Suspense fallback={<Skeleton animation="wave" />}>
```
travjenkins, line 25: Removing suspense as it is _around_ the `AppLayout` so the entire page disappears and that is horrible UX. So this is good enough for now.

src/app/index.tsx
```diff
@@ -1,13 +1,22 @@
...
 );
 const UnauthenticatedApp = React.lazy(() => import('./Unauthenticated'));
 
 function App() {
+    const intl = useIntl();
+    useTitle(
+        intl.formatMessage({
+            id: 'routeTitle.loginLoading',
+        })
+    );
```
travjenkins, line 18: This is how a component sets the title

src/components/navigation/Navigation.tsx
```diff
@@ -48,28 +51,43 @@ const Navigation = ({ onNavigationToggle, open, width }: Props) => {
                 <List aria-label="main application navigation">
                     <ListItemLink
                         icon={<ExploreIcon />}
-                        title="Dashboard"
-                        link="/dashboard"
-                        key="Dashboard"
+                        title={intl.formatMessage({
```
kiahna-tucker, line 54: Given the repetition, I am curious if it would be possible to pass `routeDetails.focalRoute` to the `ListItemLink` component and have component set `title`, `link`, and `key` properties internally.

src/pages/Home.tsx
```diff
@@ -1,8 +1,16 @@
 import { Box, Toolbar, Typography } from '@mui/material';
```
kiahna-tucker, line 1: Should this component be renamed to `Dashboard`?

src/pages/error/PageNotFound.tsx
```diff
@@ -47,8 +73,6 @@ const PageNotFound = () => {
 
     return (
         <PageContainer>
-            <Topbar isNavigationOpen={false} hideNavigationMenu />
```
kiahna-tucker, line 50: Do you want to remove the `hideNavigationMenu` prop from `Topbar`?

kiahna-tucker (approved):
Approved with suggestions. It would be great if the active discussion threads could be addressed before merging, but a follow-up PR for any changes is fine.

src/components/navigation/Navigation.tsx
travjenkins: I think so - I just didn't want to take the time to refactor this stuff. But it's a good place for something like that.

src/pages/Home.tsx
travjenkins: I think so eventually.

src/pages/error/PageNotFound.tsx
travjenkins: No... I think we should keep it for now. It is useful and I think we could use it for certain error/loading pages where we don't want the user leaving.

PR #90, opened 2022-04-19
Entity table tweaks and new capture create entrypoint

travjenkins:
## Changes

1. Moved stuff around to make entity table more reusable. We need to move away form the current pattern of having to repeat a lot of state stuff and probably go with something like [React Table](https://react-table.tanstack.com/)
2. Added ability for user to start capture creation flow from entity table
3. Remove builds page as Captures is now using Entity table
4. Removing some snapshot test... need to look into why tests take a long time. I think it is cause Jest has to transpile TS? Just a guess at this point.

## Tests

Manual

## Issues

## Content

New content for the table headers

## Screenshots

Connectors table
image

Sorting
image

Filtering
image

Creating capture from connectors list
image

image


.github/workflows/main.yml
```diff
@@ -1,142 +1,139 @@
...
-          node-version: 16
-
-      - name: Get Deps
-        uses: bahmutov/npm-install@v1
+            - name: Checkout
+              uses: actions/checkout@v2
 
-      # TODO(johnny): Should this build step be verifying that the repo
-      # isn't dirty? Seems like this is just writing out formatted results
-      # which are then ignored?
```
travjenkins, line 31: Removing this comment was the only change besides formatting.

src/app/Authenticated.tsx
```diff
@@ -26,6 +26,9 @@ export const routeDetails = {
             title: 'routeTitle.captureCreate',
             path: `create`,
             fullPath: '/capture/create',
+            params: {
+                connectorID: 'connectorID',
+            },
```
travjenkins, line 31: You can pass this route params.

src/components/capture/DetailsForm.tsx
```diff
@@ -29,13 +31,29 @@ const stateSelectors: StoreSelector<CaptureCreationState> = {
 
 function NewCaptureDetails({ connectorTags }: Props) {
     const intl = useIntl();
+    const [searchParams] = useSearchParams();
+    const connectorID = searchParams.get(
+        routeDetails.capture.create.params.connectorID
+    );
```
travjenkins, line 37: See if there is a search param with a connector ID

src/components/capture/DetailsForm.tsx
```diff
@@ -29,13 +31,29 @@ const stateSelectors: StoreSelector<CaptureCreationState> = {
...
+    useEffect(() => {
+        if (connectorID) {
+            setDetails({
+                data: {
+                    name: '',
+                    image: connectorID,
+                },
+            });
+        }
+    }, [connectorID, setDetails]);
```
travjenkins, line 55: Initially I put this into the index file. However, it feels better in this component. Mainly because this is the component that "owns" updating details... so I think it makes sense here.

src/components/tables/Captures/Rows.tsx
```diff
@@ -1,25 +1,105 @@
...
-                    <TableCell style={styling}>{row.type}</TableCell>
-                    <TableCell style={styling}>
-                        <FormattedDate
-                            day="numeric"
-                            month="long"
-                            year="numeric"
-                            value={row.updated_at}
-                        />
+            {data.map((publication: any) => (
+                <TableRow key={`Entity-${publication.id}`}>
```
travjenkins, line 14: Moved the rows from entity table into here

src/components/tables/Captures/index.tsx
```diff
@@ -1,7 +1,84 @@
...
+];
 
 function CapturesTable() {
+    const rowsPerPage = 10;
+    const [pagination, setPagination] = useState<{ from: number; to: number }>(
+        getPagination(0, rowsPerPage)
+    );
+    const [searchQuery, setSearchQuery] = useState<string | null>(null);
+    const [sortDirection, setSortDirection] = useState<SortDirection>('desc');
+    const [columnToSort, setColumnToSort] = useState<any>('updated_at');
```
travjenkins, line 53: This is the stuff that is currently duplicated. Not the best approach but not the worst.

src/components/tables/Connectors/Rows.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    </TableCell>
+                    <TableCell style={columnStyling}>
+                        {row.connector_tags[0].documentation_url ? (
+                            <ExternalLink
+                                link={row.connector_tags[0].documentation_url}
+                            >
+                                Docs
+                            </ExternalLink>
+                        ) : (
+                            <>N/A</>
```
travjenkins, line 47: I think this would make a good component. We now have two spots where links to docs are put. Also - this will be an important UX for users.

src/components/menus/__tests__/UserMenu.tsx
```diff
@@ -43,13 +43,4 @@ describe('When there is user metadata', () => {
             });
         });
     });
-
-    it('the menu should look right', async () => {
```
travjenkins, line 47: This snapshot wasn't really needed. Mainly there to prove it would work.

src/components/tables/Connectors/Rows.tsx
```diff
@@ -0,0 +1,90 @@
...
+    width: '20%',
+};
+
+function Rows({ data }: Props) {
+    const navigate = useNavigate();
+
+    return (
+        <>
+            {data.map((row) => (
+                <TableRow key={`Connector-${row.id}`}>
```
travjenkins, line 23: Most of the row renderer is taken from the old connector table

kiahna-tucker (approved):

Issue #73, opened 2022-04-06
UI Updates for Usability


dyaffe:
**Background**
This ticket intends to serve as a catch all for several UI updates that would improve it's overall usability.

**Details**

1.  Whenever a user chooses to use SSH, there are certain things that are always going to be the case and it would be good if we added checks and logical defaults to help them:

- Local port and port are always the same thing.  Is there a way for us to change them together or just present one form?  I believe it also doesn't matter what is entered in the values?  As such, can we auto fill them out?  @psFried what if we just always used 1 greater than the port that's used in other sections for both of these and eliminated them from the UI?
- When the network proxy is used, the host is actually not the host of the DB anymore, but instead it's always localhost.  We should default it to localhost and potentially grey the field out not to be used.
- We should update the description of SSH Endpoint to show it's format which is ssh://hostname:port.  It would also be ideal to validate that the user enters information in that fashion and show a validation error in the UI before saving if they don't
- We currently require that the user types out sshForwarding as the proxy type.  We should either not require that or just make it a dropdown form since there are solely preset options

2. Is there a reason that we show failed attempts at creating captures?  They're not particularly relevant after the fact and likely should be excluded from the capture view.
3. Can we link to the documentation for each capture when showing the link to them?
4. We should replace the image when showing a connector with the human readable name (which I believe is the description)
5. Would it be possible to use the wording that we've already created for the credential manager around help messages?  That wording had nice links and seemed more useful.
6. Can we display links instead of anchor tags for airbyte connectors?
7. I'm getting a silent error with nothing in supabase for airbyte connectors.  Is there anything we can do to at least get some feedback?  It says "queued".  Looks like this broke the UI and stopped my testing for now

travjenkins:
1. I think pretty much most of those would normally be in the schema. Since we'l always have SSH stuff on connectors we can also look into just writing some custom handlers for SSH in the UI.
2. We just fetch everything in the table to show lots of data. 
3. Not sure what this is requesting.
4. This is in an upcoming PR
5. Sure. Where do you want them to go?
6. Not sure what this is requesting.
7. This should be kinda fixed now. The last error we need to handle is when the control plane crashes/ just doesn't respond. It would display a general error on the page but nothing to inform the user that "Hey - the server is down"
8. Yeah... that is needed. It also does not show validation errors properly.

travjenkins:
2. Yup... I'll update the query today
3. Ahh - gotcha. Yeah we can do that. Each connector needs to add their docs to their schema.
[...]

PR #93, opened 2022-04-22
Changing capture name display and table filter label

travjenkins:
## Changes

Can now pass in a label to the entity table filter
Using the Schema title as the name for connectors if available
Add Connector title to Admin table

## Tests

Manual

## Issues

Wrapping up: https://github.com/estuary/ui/issues/73

## Content

New label for the filter

## Screenshots

Label for Connector table
image

Connectors list
image

Added Connector Title to admin table
image

### DIFF
--- a/src/utils/misc-utils.ts
+++ b/src/utils/misc-utils.ts
@@ -1,8 +1,18 @@
+import { ConnectorTag } from 'components/capture/create';
+
 export const stripPathing = (stringVal: string) => {
     return stringVal.substring(
         stringVal.lastIndexOf('/') + 1,
         stringVal.length
     );
+};
+
+export const getConnectorName = (connectorTag: ConnectorTag) => {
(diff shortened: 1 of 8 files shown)


src/utils/misc-utils.ts
```diff
@@ -1,10 +1,20 @@
...
     );
 };
 
+export const getConnectorName = (connectorTag: ConnectorTag) => {
+    if (connectorTag.title) return connectorTag.title;
+    if (connectorTag.connectors.detail) return connectorTag.connectors.detail;
+    if (connectorTag.image_tag) return stripPathing(connectorTag.image_tag);
+
+    throw new Error('Could not figure out Connector Name');
+};
```
travjenkins, line 16: This will probably slowly grow - but this is the current order of what we want to show for capture name when querying the `Connectors_Tag` table.

kiahna-tucker (approved):

Issue #92, opened 2022-04-22
Get Monaco editor to better handle resizing


travjenkins:
The Monaco editor will grow to fit the container it is in - but it will not shrink to fit the container. This means that if you widen your screen for a moment and then try to make it small again the editor is stuff at the larger size.

This is known as the Monaco Editor is made to be hyper flexible and they cannot know how to resize stuff for you. Here is a Github link on the issue: https://github.com/react-monaco-editor/react-monaco-editor/issues/53 

Starts small
![image](https://user-images.githubusercontent.com/270078/164744738-75c3b94b-c412-4b1d-9ad4-0ba1ebadbec6.png)

Grows when you fullscreen the window
![image](https://user-images.githubusercontent.com/270078/164744789-b875789b-10bc-4078-8082-604e751bf76f.png)

Does not shrink again when you make the window smaller
![image](https://user-images.githubusercontent.com/270078/164744886-a26eac30-4ded-443e-baed-4abfab298c6e.png)

travjenkins:
This has gotten a lot better but there are still some rough edges. Sometimes only the scroll bar of the editor poke out.

travjenkins:
We'll close this for now and open a new one if other users notice anything.

PR #101, opened 2022-04-28
Travjenkins/update supabase stuff

travjenkins:
## Changes

Wired up new stuff from Supabase to the UI.

Updated the file selector to use MUI data grid. This enables things like filtering and sorting. There is still some work/tweaks we need to do in there. However, I think the extra functionality is worth it given we might end up with a lot of files to select.

Added new details to the connectors page in admin. This is looking cluttered but I just wanted to get the new Open Graph stuff on the page. We need to take some time and really design what that table should look like and figure out where to put extra stuff.

## Tests

Manual

## Issues

progression: https://github.com/estuary/ui/issues/92

## Content

New content for Connector "Home Page" link

## Screenshots

Mode details on connectors (could probably combine both links into a single column)
image

Filtering on file selector
image

Can filter on type or name
image

Editor resizes better now (still not done)
image
image


src/components/editor/DraftSpec.tsx
```diff
@@ -26,19 +24,18 @@ function DraftSpecEditor() {
     const [draftSpec, setDraftSpec] = useState<DraftSpecQuery | null>(null);
 
     const handlers = {
-        change: (newVal: any) => {
+        change: (newVal: any, catalogName: string) => {
```
travjenkins, line 27: Makes sense to pass the catalog name so it can better handle async updates (these are not working 100% yet.

src/components/editor/DraftSpec.tsx
```diff
@@ -57,28 +54,18 @@ function DraftSpecEditor() {
     }, [draftSpecs, setSpecs]);
 
     useEffect(() => {
-        setDraftSpec(
-            draftSpecs[currentCatalog] ? draftSpecs[currentCatalog] : null
-        );
-    }, [currentCatalog, draftSpecs, setSpecs]);
+        if (currentCatalog) {
+            setDraftSpec(currentCatalog);
+        }
```
travjenkins, line 59: We just store the current catalog now and not the index. So we don't want to set the draft spec unless there is actually a catalog to display.

src/components/editor/DraftSpec.tsx
```diff
@@ -57,28 +54,18 @@ function DraftSpecEditor() {
...
-                    value={draftSpec.spec_patch}
-                    path={draftSpec.catalog_name}
-                    onChange={handlers.change}
-                />
-            </Box>
+            <EditorAndList
+                value={draftSpec.spec}
+                path={draftSpec.catalog_name}
+                onChange={handlers.change}
+            />
```
travjenkins, line 68: Reduced some duplication.

src/components/editor/EditorAndList.tsx
```diff
@@ -0,0 +1,41 @@
...
+    height?: number;
+}
+
+const DEFAULT_HEIGHT = 350;
+
+function EditorAndList(props: Props) {
+    const { height } = props;
+    const heightVal = height ?? DEFAULT_HEIGHT;
+
+    return (
```
travjenkins, line 17: Most of this is the same as before. However, we are now using a grid to lay stuff out. This is much nicer than before and helps make Monaco editor behave.

travjenkins: I think we still need to do some work on the "responsiveness" of the application.

src/components/editor/FileSelector.tsx
```diff
@@ -1,51 +1,97 @@
...
+const getRowId = (spec: any) => {
+    let newSelectionModel;
+
+    if (spec.id) {
+        newSelectionModel = spec.id;
+    } else if (spec.draft_id) {
+        newSelectionModel = spec.draft_id;
+    }
+
+    return `${newSelectionModel}-${spec.catalog_name}`;
```
travjenkins, line 31: Supporting both LiveSpec and DraftSpec is kinda annoying cause DraftSpec does not have a unique id. I assume this is probably good enough but we shall see as we test.

src/components/editor/FileSelector.tsx
```diff
@@ -1,51 +1,97 @@
...
+            <ListItemText
+                primary={params.row.catalog_name}
+                secondary={params.row.spec_type}
+            />
+        ),
+    },
+    {
+        field: 'spec_type',
+        headerName: 'Type',
+    },
```
travjenkins, line 49: We only display one column. Including type as a column so that the user can filter on it.

src/components/editor/FileSelector.tsx
```diff
@@ -1,51 +1,97 @@
...
-        </List>
+    const [selectionModel, setSelectionModel] = useState<GridSelectionModel>(
+        []
     );
+
+    useEffect(() => {
+        if (specs) {
+            setSelectionModel(getRowId(specs[0]) as any);
+        }
+    }, [specs]);
```
travjenkins, line 71: This basically sets the first spec as the auto selected one. This does not _really_ work in all cases so some work needs to be done. Probably just need to add hacky "initialSettingDone" sort of thing so this does not happen repeatedly.

src/components/editor/Store.tsx
```diff
@@ -54,7 +55,14 @@ const getInitialState = <T,>(
...
                 produce((state) => {
-                    state.specs = newVal;
+                    if (newVal && newVal.length > 0) {
+                        if (state.specs === null) {
+                            console.log('updating current catalog');
+
+                            state.currentCatalog = newVal[0];
+                        }
+                        state.specs = newVal;
+                    }
```
travjenkins, line 65: When specs are set the first time we want to set the current catalog to the first one in the list.

src/components/tables/Connectors/Rows.tsx
```diff
@@ -1,18 +1,24 @@
...
+import Link from 'components/tables/Link';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 import { useNavigate } from 'react-router';
 
 interface Props {
     data: Connector[];
 }
 const columnStyling = {
-    maxWidth: '20%',
+    minWidth: '10%',
```
travjenkins, line 20: Just trying to make things less crappy... still looks bad on a smaller view.

src/components/tables/Connectors/Rows.tsx
```diff
@@ -54,14 +64,22 @@ function Rows({ data }: Props) {
             {data.map((row) => {
                 return (
                     <TableRow key={`Connector-${row.id}`}>
-                        <TableCell style={columnStyling}>
-                            <ConnectorName connector={row} />
+                        <TableCell align="left" style={columnStyling}>
+                            <ConnectorName
+                                path={row.open_graph['en-US'].image}
+                                connector={row}
+                            />
```
travjenkins, line 71: Just hardcoding for now. I think we'll need to make a util to grab the proper localized stuff from the open graph. I am thinking we should end up making the open graph stuff fetched on load of the application with SWR and just keep it as a global store.

src/components/tables/Connectors/Rows.tsx
```diff
@@ -75,17 +93,16 @@ function Rows({ data }: Props) {
...
-                                >
-                                    <FormattedMessage id="captureCreation.config.source.doclink" />
-                                </ExternalLink>
-                            ) : (
-                                <FormattedMessage id="common.missing" />
-                            )}
+                            <Link
+                                path={row.connector_tags[0].documentation_url}
+                                messageId="captureCreation.config.source.doclink"
+                            />
```
travjenkins, line 99: Tiny bit of duplication reduced.

src/components/editor/DraftSpec.tsx
kiahna-tucker, line 27: Since there is not an existing variable named `value`, prefixing the first argument with `new` seems odd since a comparison is not performed.

As for the second argument, is the proper term for that segment of data _catalog name_ or _catalog namespace_? Even if both terms are suitable, we should select one and use that throughout the codebase to prevent viewers/contributors from believing there is a distinction between instances of `catalogNamespace` and `catalogName`. Whichever term will appear in Estuary's documentation and used in communication should be the term we move forward with in the codebase.

src/components/editor/DraftSpec.tsx
```diff
@@ -26,19 +24,18 @@ function DraftSpecEditor() {
     const [draftSpec, setDraftSpec] = useState<DraftSpecQuery | null>(null);
 
     const handlers = {
-        change: (newVal: any) => {
+        change: (newVal: any, catalogName: string) => {
             if (draftSpec) {
                 const newData = {
```
kiahna-tucker, line 29: For the reason described in the related comment in this file, I believe prefixing this constant with `new` is a little odd.

src/components/editor/MonacoEditor.tsx
```diff
@@ -5,24 +5,28 @@ import * as monacoEditor from 'monaco-editor/esm/vs/editor/editor.api';
 import { useRef } from 'react';
 import { useIntl } from 'react-intl';
 
-interface Props {
+export interface Props {
```
kiahna-tucker, line 8: I am an advocate for grouping optional properties and having all required properties listed first.

src/components/editor/MonacoEditor.tsx
```diff
@@ -33,13 +37,15 @@ function MonacoEditor({ disabled, value, path, onChange }: Props) {
...
-                height="300px"
+                height={`${heightVal}px`}
                 defaultLanguage="json"
                 theme={theme.palette.mode === 'light' ? 'vs' : 'vs-dark'}
                 defaultValue={intl.formatMessage({ id: 'common.loading' })}
                 value={JSON.stringify(value, null, 2)}
                 path={path}
-                options={{ readOnly: disabled ? disabled : false }}
+                options={{
+                    readOnly: disabled ? disabled : false,
```
kiahna-tucker, line 47: Could this not be rewritten as `disabled || false`?

src/components/editor/Store.tsx
```diff
@@ -54,7 +55,14 @@ const getInitialState = <T,>(
         setSpecs: (newVal) => {
```
kiahna-tucker, line 55: Whenever possible, I believe function arguments should be descriptive as opposed to generic. Is there a reason why this argument has a generic name?

src/components/tables/Link.tsx
```diff
@@ -0,0 +1,21 @@
+import ExternalLink from 'components/shared/ExternalLink';
+import { FormattedMessage } from 'react-intl';
+
+interface Props {
+    path?: string;
```
kiahna-tucker, line 5: The parameter note made in an earlier comment applies here as well.

kiahna-tucker (approved):
Approved with suggestions.

src/components/editor/DraftSpec.tsx
travjenkins: Since sometimes we need new names for variables inside functions due to scoping issues I try to keep them consistent. Hence `newVal`.

Yeah I agree but we can worry about that when we're done.

src/components/editor/EditorAndList.tsx
```diff
@@ -0,0 +1,41 @@
...
+    Props as MonacoEditorProps,
+} from 'components/editor/MonacoEditor';
+
+export interface Props extends MonacoEditorProps {
+    height?: number;
+}
+
+const DEFAULT_HEIGHT = 350;
+
+function EditorAndList(props: Props) {
```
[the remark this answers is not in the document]
travjenkins: None of those things help actually describe what this is though. `EditorAndList` is weird but it is exactly what it is.

PR #108, opened 2022-05-04
Updating the name display for connectors

travjenkins:
## Changes

- The UI will always try to pull the name of a connector from the `open_graph` column. Right now this will always pull the US English version but eventually we should make this translatable.

## Tests

Manual

## Issues

Wrapping Up: https://github.com/estuary/ui/issues/73


src/types/index.ts
```diff
@@ -34,3 +34,10 @@ export interface BaseResponse {
 export interface BaseComponentProps {
     children?: ReactNode;
 }
+
+export interface OpenGraph {
+    'en-US': {
+        image: string;
+        title: string;
+    };
+}
```
travjenkins, line 43: Eventually we should probably get this from Supabase somehow... but just hardcoding for now.

kiahna-tucker (approved):

PR #110, opened 2022-05-05
Travjenkins/random tweaks3

travjenkins:
## Changes

Created shared table cells for some of the cells in the tables
Displaying the reads from and writes to data for Captures/Materializations
Added details to the publishing table
Added an Access Grants table to Admin
If logs are told not to check for new logs with an interval the "restart" CTA is not shown

## Tests

Manual

## Issues

## Content

None

## Screenshots

Writes to and Reads from display their items as a list of chips
image

If they are over a max width when use hovers they widen
image

We now as a note that the entity was published with the UI
image

Roles Table
image

Some logs working / they don't show the restart button
image


src/components/capture/Details.tsx
```diff
@@ -1,21 +1,22 @@
...
-    };
-    logs_token: string;
+    pub_id: string;
+    live_spec_id: string;
+    published_at: string;
+    // publications: {
+    //     id: string;
+    //     job_status: JobStatus;
+    //     logs_token: string;
+    // }[];
```
travjenkins, line 19: Commenting out for now to make sure viewing the drafts at least work

src/components/capture/Create.tsx
```diff
@@ -243,6 +243,7 @@ function CaptureCreate() {
                     {
                         draft_id: id,
                         dry_run: false,
+                        detail: 'Published via UI',
```
travjenkins, line 246: If this stays we should move this to a common place.

src/components/tables/Connectors/index.tsx
```diff
@@ -71,10 +71,10 @@ function ConnectorsTable() {
...
             <EntityTable
                 noExistingDataContentIds={{
-                    header: 'admin.connectors.main.message1',
-                    message: 'admin.connectors.main.message2',
-                    docLink: 'admin.connectors.main.message2.docLink',
-                    docPath: 'admin.connectors.main.message2.docPath',
+                    header: 'connectors.main.message1',
+                    message: 'connectors.main.message2',
+                    docLink: 'connectors.main.message2.docLink',
+                    docPath: 'connectors.main.message2.docPath',
```
travjenkins, line 77: typo

src/components/tables/cells/ChipList.tsx
```diff
@@ -0,0 +1,63 @@
+import { Box, Chip, styled, TableCell } from '@mui/material';
```
travjenkins, line 1: New call that can display lists of strings in chips. I think eventually this might expand and they become elements that the user can take action on, select some, select all, deselect all, etc. But right now they just display the text.

src/components/tables/cells/EntityName.tsx
```diff
@@ -0,0 +1,45 @@
+import { Box, TableCell, Tooltip } from '@mui/material';
+import { getDeploymentStatusHexCode } from 'utils/misc-utils';
+
+interface Props {
+    name: string;
+}
+
+function EntityName({ name }: Props) {
```
travjenkins, line 8: Display the name with the status.

src/components/tables/cells/TimeStamp.tsx
```diff
@@ -0,0 +1,37 @@
+import { Box, TableCell, Tooltip } from '@mui/material';
+import { formatDistanceToNow } from 'date-fns';
+import { FormattedDate } from 'react-intl';
+
+interface Props {
+    time: string;
+}
+
+function TimeStamp({ time }: Props) {
```
travjenkins, line 9: Display the time

src/types/index.ts
```diff
@@ -41,3 +41,7 @@ export interface OpenGraph {
         title: string;
     };
 }
+
+export interface JobStatus {
+    type: string;
+}
```
travjenkins, line 47: Slowly adding more common types that are used in several tables.

src/lang/en-US.ts
```diff
@@ -187,6 +189,8 @@ const LogsDialog: ResolvedIntlConfig['messages'] = {
 
 const AdminPage: ResolvedIntlConfig['messages'] = {
     'admin.header': `Administration`,
+    'admin.roles': `Your ?Roles?`,
```
kiahna-tucker, line 192: Is the use of question marks in `admin.roles` and `admin.roles.message` intentional?

kiahna-tucker (approved):
Approved with question.

src/components/shared/UserAvatar.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+const DEFAULT_SIZE = 25;
+
+function UserAvatar({ avatarUrl, size, userName }: Props) {
+    return (
+        <Avatar
+            src={avatarUrl ?? ''}
+            sx={{
+                fontSize: (size ?? DEFAULT_SIZE) / 1.5,
```
kiahna-tucker, line 16: It may be worthwhile to save the result of the common evaluation in a constant and use that value to set `fontSize`, `height`, and `width`.

src/lang/en-US.ts
travjenkins: Yeah it was. I was trying to think of a way to make it stand out as a term we were not sure about yet.

src/components/shared/UserAvatar.tsx
travjenkins: I will... I wasn't sure if it was worth it wile doing but if you noticed it as well I will change it.

PR #112, opened 2022-05-05
.env: enable email auth, change port to 3000, and add commented local endpoint

jgraettinger:
With these changes, and some changes on the control-plane side, using the UI locally works great 👍 

We're even able to login as different people, and "sign up" new accounts as well to test out onboarding flows.


travjenkins (approved):
lgtm

PR #113, opened 2022-05-06
Allow rendering JSON forms from wacky schemas

### LINKED ISSUE #107: Unable to render a subset of materialization config forms.
Currently, the following materialization endpoint and resource configuration schemas cannot be rendered: PostgreSQL, BigQuery, and Snowflake.

### Notes

When evaluating the issue, I noticed a pattern amongst the schemas whose refs were able to be resolved: the `$ref` property was present. When a schema containing a defined `$ref` is passed to `JsonRefs.resolveRefs()`, the `$schema` property in the original schema is placed in the object of the `refs` property in the resulting schema.

For example, consider the following sample schema:

```ts
{
  "$ref": "#/definitions/Config",
  "$schema": "http://json-schema.org/draft-04/schema#",
  "definitions": { ... }
  // additional properties
}
```

For the purpose of this example, let's refer to it as _Origin Schema A_. When _Origin Schema A_ is passed to `JsonRefs.resolveRefs()`, it creates _Resolved Schema A_ that **can** be rendered and is defined as follows:

```ts
{
  "refs": {
    "#": {
      "def": {
        "$ref": "#/definitions/Config",
        "$schema": "http://json-schema.org/draft-04/schema#",
        "definitions": { ... },
        },
        // additional properties
      },
    // additional property
  },
  "resolved": {
    "type": "object",
    "required": ["database", "host", "password", "port", "user"],
    // additional properties
}
```

Now, say there is an _Origin Schema B_ that is defined as follows:

```ts
{
  "type": "object",
  "$schema": "http://json-schema.org/draft-04/schema#",
  "required": ["host", "password", "user"],
  // additional properties
```

When _Origin Schema B_ is passed to `JsonRefs.resolveRefs()`, it creates _Resolved Schema B_ that **cannot** be rendered and is defined as follows:

```ts
{
  "refs": {},
  "resolved": {
    "type": "object",
    "$schema": "http://json-schema.org/draft-04/schema#",
    "required": ["host", "password", "user"],
    // additional properties
   },
```

I am curious if the presence of the `$schema` property in the resolved schema (see _Resolved Schema B_ reference) causes the schema validation error. Sharing this context in the event it helps find a path to a solution.

### Reference
app__materialization-create-error__postgres__no-schema-with-key-or-ref

psFried:
## Changes

Fixes #107 

The EndpointConfigForm and ResourceConfigForm previously would barf
whenever they encountered a non-compliant JSON schema. This commit makes
the AJV validator much more permissive in the schemas that it will
accept.

This also removes the `json-refs` dependency and assumes that all json
schemas are fully self-contained. This property is ensured by the build
process for flow collection schemas. For endpoint and resource schemas,
it seems to also generally hold.

This seems to work to get all of our forms rendering, and so I'm submitting it as a minimal PR just to address #107. There's a number of other less major issues remaining with the forms, and I'll plan on submitting a follow-on PR to address those.

## Tests

I tested manually with all the connectors in TW and it seemed to work ok. IMO we should hold off on automated tests until we have greater certainty that JSON Forms will work for us.

## Content

None

## Screenshots

image

### DIFF
--- a/src/components/materialization/ResourceConfigForm.tsx
+++ b/src/components/materialization/ResourceConfigForm.tsx
@@ -1,4 +1,3 @@
-import { createAjv } from '@jsonforms/core';
 import { materialCells } from '@jsonforms/material-renderers';
 import { JsonForms } from '@jsonforms/react';
 import { StyledEngineProvider } from '@mui/material';
@@ -9,9 +8,12 @@ import useEntityStore, {
     fooSelectors,
     FormStatus,
 } from 'components/shared/Entity/Store';
-import JsonRefs from 'json-refs';
+import {
+    createJSONFormDefaults,
+    setDefaultsValidator,
+} from 'components/shared/Entity/EndpointConfigForm';
 import { isEmpty } from 'lodash';
-import { useEffect, useState } from 'react';
+import { useEffect } from 'react';
 import {
     defaultOptions,
     defaultRenderers,
@@ -24,8 +26,6 @@ type Props = {
     resourceSchema: any;
 };
 
-const defaultAjv = createAjv({ useDefaults: true });
-
 const stateSelectors: StoreSelector<CreationState> = {
     formData: (state) => state.resourceConfig.data,
     setConfig: (state) => state.setResourceConfig,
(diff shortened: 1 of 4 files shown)


src/components/materialization/ResourceConfigForm.tsx
```diff
@@ -37,63 +37,42 @@ function NewMaterializationResourceConfigForm({ resourceSchema }: Props) {
     const displayValidation = useEntityStore(fooSelectors.displayValidation);
     const status = useEntityStore(fooSelectors.formStateStatus);
 
-    const [deReffedSchema, setDeReffedSchema] = useState<any | null>(null);
-
     // Resolve Refs & Hydrate the object
     //  This will hydrate the default values for us as we don't want JSONForms to
     //  directly update the state object as it caused issues when switching connectors.
     useEffect(() => {
-        async function resolveSchemaRefs(endpointResponse: any) {
```
travjenkins, line 46: 🎉🎉🎉

travjenkins (approved):
lgtm

kiahna-tucker (approved):

PR #114, opened 2022-05-06
Travjenkins/random tweaks5

travjenkins:
## Changes

- [x] Show draft errors on the creation page by the logs
- [x] Collapse logs by default so the error section does not take up too much room
- [x] Fix some styling with the file/editor 
- [x] Disable the "diff button" in the editor as the diff viewer is really broken
- [x] Check if the user has access to any object for creation
- [x] Added materialize button to captures table (disabled)
- [x] Updated table to handle columns without titles
- [x] Offer autocomplete for entity name with the proper prefixes
     - This is partially done with a second field. This should be temporary.

## Tests

Manual

## Issues

Working off Dave's document

## Content

Tweaked some messages

## Screenshots

Showing draft errors
image

Getting editor to stay inside the table lines
image

Handling edge case where user does not have access to any grants
image

Actions are closer to items that action is taken on. Details no longer looks like button
image


src/components/editor/EditorAndList.tsx
```diff
@@ -17,22 +18,27 @@ function EditorAndList(props: Props) {
     return (
         <Grid
             container
-            spacing={2}
             sx={{
                 bgcolor: 'background.paper',
-                height: heightVal,
+                height: `${heightVal + DEFAULT_TOOLBAR_HEIGHT}px`,
+                mb: 2,
```
travjenkins, line 24: Need to make sure the things using editor know about the "toolbar" at the top now. Also adding some margin below to make sure there is a little empty space around the editor. Having editor near the button of the page without extra space just looked odd to me.

src/components/materialization/create/CollectionConfig.tsx
```diff
@@ -17,21 +17,20 @@ function CollectionConfig() {
...
-                                    collectionName={collection}
-                                    id={imageTag.id}
-                                    key={`CollectionResourceConfig-${index}`}
-                                />
-                            );
-                        })}
-                    </>
-                }
-            />
+            >
```
travjenkins, line 20: Having a `prop` named body does not really make sense when you can just pass children in.

src/components/shared/Entity/DetailsForm.tsx
```diff
@@ -57,40 +62,69 @@ function DetailsForm({ connectorTags, messagePrefix }: Props) {
         }
     }, [connectorID, setDetails]);
 
+    const accessGrantsOneOf = useMemo(() => {
```
travjenkins, line 65: We need to "post process" some data so throwing those into memos. Keeping them apart since they can change independently.

src/components/shared/Entity/DetailsForm.tsx
```diff
@@ -57,40 +62,69 @@ function DetailsForm({ connectorTags, messagePrefix }: Props) {
...
                         id: 'entityName.description',
                     }),
-                    maxLength: 1000,
-                    minLength: 3,
+                    // TODO (prefix) Make prefix a part of the name field
                     // This pattern needs to match https://github.com/estuary/animated-carnival/blob/main/supabase/migrations/03_catalog-types.sql
-                    pattern: `^([a-zA-Z0-9-_.]+/)+[a-zA-Z0-9-_.]+$`,
+                    // Right now with prefix broken out it means the first part is a bit different
+                    // `^([a-zA-Z0-9-_.]+/)+[a-zA-Z0-9-_.]+$`
+                    pattern: `^([a-zA-Z0-9-_./])+[^/]$`,
```
travjenkins, line 127: Updating pattern but need to change it a bit right now since the prefix is in a stand alone input.

src/components/shared/Entity/Error/DraftErrors.tsx
```diff
@@ -0,0 +1,22 @@
+import KeyValueList, { KeyValue } from 'components/shared/KeyValueList';
+import useDraftSpecErrors from 'hooks/useDraftSpecErrors';
+
+export interface DraftErrorProps {
+    draftId?: string | null;
+}
+
+function DraftErrors({ draftId }: DraftErrorProps) {
```
travjenkins, line 8: Showing errors from the draft errors table

src/components/shared/Entity/Error/Logs.tsx
```diff
@@ -0,0 +1,45 @@
...
+    Typography,
+} from '@mui/material';
+import Logs from 'components/Logs';
+import ErrorBoundryWrapper from 'components/shared/ErrorBoundryWrapper';
+
+export interface ErrorLogsProps {
+    logToken?: string | null;
+}
+
+function ErrorLogs({ logToken }: ErrorLogsProps) {
```
travjenkins, line 16: Logs in the error panel can be collapsed by default to make a bit more room for the main view. Also, it allows users to control when they are fetching logs a bit better.

src/components/shared/Error.tsx
```diff
@@ -29,64 +27,66 @@ function Error({ error, hideTitle }: Props) {
...
+                            transition: 'all 250ms ease-in-out',
+                        }}
+                    >
+                        <ExpandMore />
+                    </IconButton>
+                    <Collapse in={expanded} timeout="auto" unmountOnExit>
+                        <Paper variant="outlined" square>
+                            <KeyValueList data={details} />
+                        </Paper>
+                    </Collapse>
```
travjenkins, line 83: Will handle switching this over to accordion with the next PR.

src/components/tables/cells/EntityName.tsx
```diff
@@ -9,35 +9,19 @@ function EntityName({ name }: Props) {
...
-                        style={{
-                            verticalAlign: 'middle',
-                        }}
-                    >
-                        {name}
-                    </span>
-                </Box>
-            </Tooltip>
+            <>
+                <EntityStatus />
```
travjenkins, line 16: Broke the entitystatus into a component and do not need tooltip as we always show the full entity name.

src/components/capture/Create.tsx
```diff
@@ -58,6 +59,10 @@ function CaptureCreate() {
 
     // Supabase stuff
     const supabaseClient = useClient();
+    const { combinedGrants } = useCombinedGrantsExt({
+        onlyAdmin: true,
```
kiahna-tucker, line 63: I believe `adminOnly` is the more common way to phrase this. A change is not necessary, but calling it out in the event you'd like to update it.

src/components/capture/Create.tsx
```diff
@@ -301,7 +307,7 @@ function CaptureCreate() {
                 supabaseClient
                     .from(TABLES.DRAFTS)
                     .insert({
-                        detail: entityName,
+                        detail: `${entityPrefix.title}${entityName}`,
```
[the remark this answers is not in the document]
kiahna-tucker: Since this concatenation is performed in more than one place, would it be beneficial to save it off in a constant?

If you plan to replace this code fragment soon and in a fashion that renders the constant unnecessary, there is no reason to change the logic. If the code fragment will linger for awhile, it may be worth considering.

kiahna-tucker (approved):
Approved with suggestions and comments. No issue with any logical changes or amendments. I do have a few lingering questions about some of the names I came across, but I will pose those in our daily meeting tomorrow. I imagine certain items are merely in flux.

src/components/editor/EditorAndList.tsx
```diff
@@ -8,7 +8,8 @@ export interface Props extends MonacoEditorProps {
     height?: number;
 }
 
-const DEFAULT_HEIGHT = 350;
+const DEFAULT_TOOLBAR_HEIGHT = 20;
```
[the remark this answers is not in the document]
travjenkins: Yeah I thought about it. Given we have import/export I am slowly feeling like moving constants to a shared location is not needed. I am starting to feel like the place where the assumption/limitation is causing the constant should house the constant. How sure though.

PR #115, opened 2022-05-06
Johnny/avatar fixes

jgraettinger:
## Changes

User avatars should expect an email, but a name may not be available.
Update accordingly.

## Tests

Updated UserAvatar types and dealt with fallout from there.

## Screenshots

image


travjenkins (approved):
lgtm

PR #117, opened 2022-05-09
Connectors Table Temporary Patch

kiahna-tucker:
## Changes

Control the connectors table display so that a row is not rendered for connectors without tag information (i.e., no entries within the `connector_tags` table).

**NOTE:** This is a temporary patch. Pending an official UX decision.

## Tests

Only manual testing is performed.


travjenkins (approved):
lgtm

PR #121, opened 2022-05-10
Enable LogRocket

travjenkins:
## Changes

- [x] Adding init call for LogRocket
- [x] Added `env` props to enable/disable anything
- [x] Added sanitization and user identification settings

## Tests

Manual testing

## Issues

None... discussed this in a meeting

## Content

None

## Screenshots

New settings
image

LogRocket Dashboard
image

Default settings are to not show anything typed into the UI
image


.env
```diff
@@ -22,4 +22,12 @@ REACT_APP_SWR_LOG_CACHE=false
...
\ No newline at end of file
+REACT_APP_DEFAULT_PAGE_TITLE=Estuary Flow
+
+# LogRocket settings
+REACT_APP_LOGROCKET_APP_ID=tue20y/estuarytest
+REACT_APP_LOGROCKET_SANITIZE_INPUTS=true
+REACT_APP_LOGROCKET_SANITIZE_TEXT=false
+REACT_APP_LOGROCKET_ID_USER=true
+REACT_APP_LOGROCKET_ID_USER_INCLUDE_NAME=true
+REACT_APP_LOGROCKET_ID_USER_INCLUDE_EMAIL=true
```
travjenkins, line 33: Trying to make configuring this as granular as possible. Important for anyone running their own version of the UI.

src/app/index.tsx
```diff
@@ -13,6 +15,13 @@ function App() {
 
     const { user } = Auth.useUser();
 
+    useEffect(() => {
+        if (user) {
+            console.log('this was called');
+            identifyUser(user);
+        }
+    }, [user]);
```
travjenkins, line 23: Only need to run this when user changes.

src/app/index.tsx
```diff
@@ -13,6 +15,12 @@ function App() {
 
     const { user } = Auth.useUser();
 
+    useEffect(() => {
+        if (user) {
+            identifyUser(user);
+        }
+    }, [user]);
+
```
travjenkins, line 23: Only run when user changes (so basically once a session)

src/index.tsx
```diff
@@ -17,6 +18,8 @@ import AppProviders from './context';
 //     console.log('  Deet > ', { inter, phase });
 // };
 
+initLogRocket();
```
travjenkins, line 21: Not sure the best place to call this but decided as early as possible was probably best.

src/services/logrocket.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (logRocketSettings.appID) {
+        LogRocket.init(logRocketSettings.appID, {
+            dom: {
+                inputSanitizer: logRocketSettings.sanitize.inputs,
+                textSanitizer: logRocketSettings.sanitize.text,
+            },
+        });
+        setupLogRocketReact(LogRocket);
+    }
+};
```
travjenkins, line 24: Startup LogRocket. If there is no appID then we assume the functionality is disabled. The `dom` setting is where you can control what is shown or hidden.

src/services/logrocket.ts
```diff
@@ -0,0 +1,43 @@
...
+            traits.name = userDetails.userName;
+        }
+
+        if (logRocketSettings.idUser.includeEmail) {
+            traits.email = userDetails.email;
+        }
+
+        LogRocket.identify(user.id, traits);
+    }
+};
```
travjenkins, line 43: ID user if that is enabled and only send the data that is enabled to be sent.

src/utils/env-utils.ts
```diff
@@ -7,6 +7,8 @@ declare global {
     }
 }
 
+const ENABLED = 'true';
```
travjenkins, line 10: Thought about making an `enum` but since this isn't _really_ going to check for multiple values and only cares if it is `true` decided just a plain old const was good.

.env
```diff
@@ -22,4 +22,12 @@ REACT_APP_SWR_LOG_CACHE=false
 REACT_APP_SWR_LOG_REQUESTS=false
 
 # This is the title displayed while the UI is initially loading in the browser
-REACT_APP_DEFAULT_PAGE_TITLE=Estuary Flow
\ No newline at end of file
+REACT_APP_DEFAULT_PAGE_TITLE=Estuary Flow
+
+# LogRocket settings
+REACT_APP_LOGROCKET_APP_ID=tue20y/estuarytest
```
travjenkins, line 28: This is still my personal test account. Once we see how this works with more than just a local we can make a decision on licensing.

kiahna-tucker (approved):
Approved without review.

PR #122, opened 2022-05-10
UI Tweaks 7

travjenkins:
## Changes

- [x] Allow file selector to be resized
- [x] Add loading indicator to the create workflow
- [x] Keep collections selector open when item is selected
- [x] Fix materializations sending bindings in wrong format

## Tests

Manual

## Issues

Working off Google Doc

## Content

None

## Screenshots

Small loading bar shows when doing something
image

You adjust the width with the middle bar
image

Min width so you don't lose a panel
image

Removing log title in details to make things line up a bit nicer
image
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
                 "react-dom": "^17.0.2",
                 "react-error-boundary": "^3.1.4",
                 "react-intl": "^5.25.1",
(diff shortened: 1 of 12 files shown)


kiahna-tucker (approved):
Approved to reduce EOD delay.

PR #123, opened 2022-05-10
UI Tweaks 8

travjenkins:
## Changes

- [x] Turn off LogRocket for locals
- [x] Fix materializations name
- [x] Make catalog more visible once it is generated
- [x] Handle JWT expiration better
- [x] Show message when forms have validation issues

## Tests

## Issues

## Content

## Screenshots

When JWT expire error happens we auto logout user and notify them
image 

Page will scroll to catalog when it is generated
image

Display a general message when there are problems
image


kiahna-tucker (approved):
Approved in advance.

PR #131, opened 2022-05-12
Travjenkins/random tweaks11

travjenkins:
## Changes

- [x] Encrypt endpoint config
- [x] Handle empty table view which is good enough for now. It shows the same as when there is NO data... I think it would be a better UX for the table to know the difference between no data and nothing is returned with the current filter
- [x] Fix materialization catalog editor not showing up properly
- [x] Handle state of form better
- [x] Fix materialize from capture create
- [x] Update how the forms are displayed
- [x] Update form UX for creation

## Tests

Manual

## Issues

Working off Google Doc

## Content

Two basic error messages for error handling

## Screenshots

The connection config is encrypted
image

Basic error when encryption fails
image

Handle empty table
image

New layout for the resource config
image

Made the "nested" forms take up less space for endpoint config
image

Making forms collapsible
image

Break up the form so there are multiple tabs
image


src/components/capture/Create.tsx
```diff
@@ -121,15 +119,17 @@ function CaptureCreate() {
         EditorStoreState<DraftSpecQuery>['id']
     >((state) => state.id);
 
+    // Reset the cataog if the connector changes
```
kiahna-tucker, line 122: > Reset the cataog if the connector changes

There is a typo in the word _catalog_ in this comment. This does not need to be changed. Calling this out for general awareness.

src/components/editor/MonacoEditor.tsx
```diff
@@ -39,39 +39,36 @@ function MonacoEditor({
...
+    //     }
 
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [serverUpdate]);
+    //     // eslint-disable-next-line react-hooks/exhaustive-deps
+    // }, [serverUpdate]);
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    const bar = useCallback(
+    const editorUpdateDebounce = useCallback(
```
kiahna-tucker, line 70: Since this function is not being exported elsewhere, what do you think about dropping the word _editor_ in the name? Not a critical change, but `updateDebounce` aligns with the function convention where names start with an active verb and it is slightly more concise.

src/components/materialization/create/index.tsx
```diff
@@ -277,6 +272,7 @@ function MaterializationCreate() {
                 setFormState({
                     status: FormStatus.GENERATING_PREVIEW,
                 });
+                setDraftId(null);
 
                 // TODO: Use connector_tags.resource_spec_schema as the value of bindings.resource when the
```
kiahna-tucker, line 277: This TODO comment can be removed.

src/context/Theme.tsx
```diff
@@ -126,6 +126,13 @@ export const outlineSx: SxProps<Theme> = {
     border: `1px solid ${slate[200]}`,
 };
 
+// TODO (theme) Figure out how to make these composable
```
kiahna-tucker, line 129: Could you clarify this comment for me? I am not sure I understand the task it is identifying.

src/components/editor/MonacoEditor.tsx
travjenkins: I'm going to leave it. The name has a label because later on I feel we're going to need other debouces/throttles in there that are not for the editor itself.

kiahna-tucker (approved):
Approved in the interest of time.

Issue #134, opened 2022-05-13
Display shard status in entity table.


kiahna-tucker:
An API supplied by the `data-plane-gateway` surfaces an entity's shard status. This piece of data should be used to dictate the entity status indicator within the UI, replacing the hardcoded value.

PR #139, opened 2022-05-18
Display Shard Status in Entity Table.

kiahna-tucker:
## Changes

Fetch and display an indicator of the shard status from the shards list API provided by the data plane gateway.

## Tests

## Issues

#134 Display shard status in entity table.

## Screenshots

**Capture | Entity Table with Status**
pr-screenshot__140__auth-gateway-integration__captures

<br />

**Materialization | Entity Table with Status**
pr-screenshot__140__auth-gateway-integration__materializations


src/hooks/useGatewayAuthToken.ts
```diff
@@ -0,0 +1,73 @@
...
+    );
+    const { data: grants } = useSelect(combinedGrantsQuery);
+
+    const prefixes: string[] =
+        grants?.data.map(({ object_role }) => object_role) ?? [];
+
+    const gatewayConfig = getStoredGatewayAuthConfig();
+
+    return useSWR(
+        !gatewayConfig?.token && prefixes.length > 0
```
kiahna-tucker, line 57: This condition needs to be expanded. The `gateway_auth_token` API should be hit if there is no gateway configuration, no authentication token, or no valid authentication token in local storage and object role grants exist for this user. A JWT library will need to be used to determine whether an authentication token has expired. Any preferences?

src/services/shard-client.ts
```diff
@@ -0,0 +1,64 @@
...
+
+            if (shards.length > 0) {
+                setShards(shards);
+            }
+        })
+        .catch((error: string) => {
+            if (
+                error.includes(ErrorFlags.TOKEN_INVALID) ||
+                error.includes(ErrorFlags.TOKEN_NOT_FOUND) ||
+                error.includes(ErrorFlags.OPERATION_INVALID)
```
kiahna-tucker, line 45: This conditional statement needs to be narrowed, removing the check for invalid operation flag. The check was included as a temporary patch due to the deployment of a stale version of the `data-plane-gateway`.

src/hooks/useGatewayAuthToken.ts
travjenkins: Not really - we should start by looking at Supabase and if they don't have one find one.

src/services/shard-client.ts
travjenkins: Anything bad happen by leaving it in?

src/utils/env-utils.ts
```diff
@@ -105,3 +111,49 @@ export const getEncryptionSettings = () => {
...
+export const getStoredGatewayAuthConfig =
+    (): GatewayAuthTokenResponse | null => {
+        const config = localStorage.getItem(LocalStorageKeys.GATEWAY);
+
+        return config ? JSON.parse(config) : null;
+    };
+
+export const removeGatewayAuthConfig = (): void => {
+    localStorage.removeItem(LocalStorageKeys.GATEWAY);
+};
```
travjenkins, line 147: Post PR: This should not be in this utils as it is not related to .env file. We should start making a local storage utils or hooks... depending on needs.

src/services/shard-client.ts
kiahna-tucker: No, it will just result in the regeneration of a valid token. Since the `data-plane-gateway` was redeployed, I will evaluate whether it contains the changes needed to remove the extra condition.

PR #147, opened 2022-05-19
Add nginx.conf so that page refreshes will work

psFried:
## Changes

React-router wasn't playing nicely with the default nginx config.
This adds a config file, based on the default, which has a `try_files`
that falls back to index.html, which is what react-router needs.
With this, you should be able to refresh on any page and stay where
you're at in the UI.

### DIFF
--- a/public/Dockerfile
+++ b/public/Dockerfile
@@ -1,3 +1,8 @@
 FROM nginx:alpine
 
+COPY ./nginx.conf /etc/nginx/nginx.conf
 COPY ./ /usr/share/nginx/html/
+# Delete the nginx.conf file so that we don't serve our config via nginx
+# This is because I don't know an easier way to COPY all files _except_ that one.
+RUN rm /usr/share/nginx/html/nginx.conf
+
--- a/public/nginx.conf
+++ b/public/nginx.conf
@@ -1,0 +1,61 @@
+# This config file was taken from the default nginx image by running:
+# $ docker run --rm --detach --name tmp nginx:alpine
+# $ docker cp tmp:/etc/nginx/nginx.conf
+# Any modifications from the default should have comments.
+user  nginx;
+worker_processes  auto;
+
+error_log  /var/log/nginx/error.log notice;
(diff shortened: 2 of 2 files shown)


public/Dockerfile
```diff
@@ -1,3 +1,8 @@
 FROM nginx:alpine
 
-COPY ./ /usr/share/nginx/html/
\ No newline at end of file
+COPY ./nginx.conf /etc/nginx/nginx.conf
+COPY ./ /usr/share/nginx/html/
+# Delete the nginx.conf file so that we don't serve our config via nginx
+# This is because I don't know an easier way to COPY all files _except_ that one.
+RUN rm /usr/share/nginx/html/nginx.conf
```
travjenkins, line 7: Could we move this file over to a `.docker` folder instead of having it in the `public` folder?

travjenkins (approved):
lgtm

public/Dockerfile
psFried: Maybe? I was pretty unclear on how else to end up with the file inside of `build/`, which is the docker context for the image build. But it occurs to me now that the docker build context can be whatever we want it to be, so I can definitely move that conf file so we don't have to delete it later during the image build. Is there some special handling with `.docker` and create-react-app? I couldn't immediately find and docs on that.

travjenkins: That's a fair point and I think we can leave it as this is the easiest. In other setup docs I've read they just tell you to copy the build into the folder you want... and that's basically the same amount of work/coding as this is.

psFried: Ok, I just tried a quick test, and I don't see an out-of-the-box way to have other stuff end up in that `build/` dir. I also looked at what it would take to just change the docker build context, and decided that's not worth it now, either, because `gcloud run deploy` doesn't support separate arguments for the source Dockerfile and the context. We can revisit that later, though, if we want to change to build images directly instead of through `gcloud run deploy`, but :shrug:

PR #158, opened 2022-05-25
Travjenkins/random tweaks15

travjenkins:
## Changes

- [ ] Add support for forms to display secrets correct

## Tests

Manual

## Issues

None

## Content

None

## Screenshots

image

### DIFF
--- a/src/components/editor/FileSelector.tsx
+++ b/src/components/editor/FileSelector.tsx
@@ -9,7 +9,7 @@ import { EditorStoreState } from 'components/editor/Store';
 import { DraftSpecQuery } from 'hooks/useDraftSpecs';
 import { PublicationSpecQuery } from 'hooks/usePublicationSpecs';
 import { useZustandStore } from 'hooks/useZustand';
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { FormattedMessage } from 'react-intl';
 
 const initialState = {
@@ -51,7 +51,7 @@ const columns: GridColDef[] = [
 ];
 
 function EditorFileSelector() {
-    const [initDone, setInitDone] = useState(false);
+    const initDone = useRef(false);
 
     const isSaving = useZustandStore<
         EditorStoreState<DraftSpecQuery>,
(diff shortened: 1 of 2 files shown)


src/components/editor/FileSelector.tsx
```diff
@@ -51,7 +51,7 @@ const columns: GridColDef[] = [
 ];
 
 function EditorFileSelector() {
-    const [initDone, setInitDone] = useState(false);
+    const initDone = useRef(false);
```
travjenkins, line 54: Using a ref makes more sense here

src/services/jsonforms.ts
```diff
@@ -355,10 +368,13 @@ const generateUISchema = (
...
+    // just create a default Control, but we first check if we should turn the input into a password
+    // input. This overrides other settings to keep things secure. Then we check if we need to set
+    // the `multi` option based on whether the json schema contains a `multiline` annotation.
     const controlObject: ControlElement = createControlElement(currentRef);
-    if (isMultilineText(jsonSchema)) {
+    if (isSecretText(jsonSchema)) {
+        addOption(controlObject, 'format', 'password');
+    } else if (isMultilineText(jsonSchema)) {
         addOption(controlObject, 'multi', true);
     }
```
travjenkins, line 379: Decided to make the check for secret first as that should never be overwritten

psFried: I'm curious whether these two options need to be mutually exclusive. Would it work to have a field that is a multi-line secret? The use case I'm thinking of is for ssh tunneling, where you need to paste in a private key. IIRC it actually posed a problem for that field to _not_ be multi-line because it would strip out all newline characters, which causes the certificate to be malformed. This is the only use case that I know of currently for multi-line secrets, so we can talk though some other options if that seems unworkable.

src/components/editor/FileSelector.tsx
psFried: I'll just trust you on that one :upside_down_face:

psFried (commented):
LGTM % the one conversation we should resolve

kiahna-tucker (approved):
Approved in the interest of time. The comment Phil left behind in _src/services/jsonforms.ts_ should be addressed before merging.

src/services/jsonforms.ts
travjenkins: You cannot do that with HTML. So if we ever want a multi line secret then we'll have to build an input for it.

jgraettinger: Google service accounts may be motivating here. They're both big and also secret.

PR #164, opened 2022-05-26
Various content nits and tweaks

oliviamiannone:
## Changes

First run-through of content with proposed edits. Suggestions include capitalization, punctuation, and re-wording, which you may want to take or leave. The intention was to not majorly change the meaning of anything.

I also added a couple of doc links, which I wanted to call out specifically to make sure I did them right. (It was easy enough to copy the syntax but not sure if anything else needs to get wired in for them to work).


travjenkins (approved):
lgtm

Issue #161, opened 2022-05-26
Display reported replica processing errors in details panel.


kiahna-tucker:
The `shards-list` API reports encountered replica processing errors for shards with a FAILED status. These errors should be bundled by an alert and displayed in the details panel of an entity table row.

PR #170, opened 2022-05-31
Display Shard ID and Replica Processing Errors.

kiahna-tucker:
## Changes

Display the shard ID and replica processing errors in the details panel of the entity table.

## Tests

## Issues

Display reported replica processing errors in details panel. #161 

## Content

The label displayed next to the shard ID (i.e., `CaptureDetails.captureDetails.shardDetails.title`) and the text displayed on the replica processing error accordion (i.e., `CaptureDetails.captureDetails.shardDetails.errorTitle`).

## Screenshots

_Pending..._


src/components/capture/Details.tsx
```diff
@@ -46,11 +82,106 @@ function CaptureDetails({ lastPubId, disableLogs }: Props) {
...
+                                                    }}
+                                                    onMount={handlers.mount}
+                                                    value={shardDetails.errors
+                                                        .join(NEW_LINE)
+                                                        .split(/\\n/)
+                                                        .join(NEW_LINE)
+                                                        .replaceAll(
+                                                            /\\"/g,
+                                                            '"'
+                                                        )}
```
travjenkins, line 155: Since this is done else where we should probably make this a service and share this.

src/components/capture/Details.tsx
```diff
@@ -46,11 +82,106 @@ function CaptureDetails({ lastPubId, disableLogs }: Props) {
...
+                                                    options={{
+                                                        lineNumbers: 'off',
+                                                        readOnly: true,
+                                                        scrollBeyondLastLine:
+                                                            false,
+                                                        minimap: {
+                                                            enabled: false,
+                                                        },
+                                                    }}
+                                                    onMount={handlers.mount}
```
travjenkins, line 147: Don't know what the solution is here... but I feel we are repeating some of the editor stuff. This might be okay because having the react monaco editor already reduces duplication a lot. However, I am thinking we might be able to share the theme and default options somehow.

travjenkins (approved):
Looks good. Comments can be thought and worked on later.

oliviamiannone (commented):
Labels look good to me! My one note is that, since this is (to my knowledge) the only place a UI user is exposed to the concept of shards, there should be a link to the docs. 
I'd recommend where it says "Shard ID" linking to https://docs.estuary.dev/concepts/advanced/shards/ or alternatively adding some simple hover text saying something like "A shard is a unit of compute assigned to your task by the Flow runtime." Whatever makes the most sense for your side. 
As for the list of replica processing errors - I'm not sure what the error text would look like so I want to call this out just in case. I assume the errors themselves are fairly explicit with next steps for troubleshooting for the user. But if they are somewhat ambiguous, we may need to add docs here as well. I can circle back on this one.

PR #177, opened 2022-06-01
Add local .env so npm start works OOTB

psFried:
Adds `.env.development.local`, which is used to override properties from
`.env` when running `npm start`. The purpose of this is to allow
developers to run `npm start` without needing to modify the `.env` file
to run locally. The approach was taken from:
https://create-react-app.dev/docs/adding-custom-environment-variables/#adding-development-environment-variables-in-env

## Tests

Works on my machine :wink:

### DIFF
--- a/.env
+++ b/.env
@@ -2,13 +2,8 @@ EXTEND_ESLINT=true
 
 PORT=3000
 
-# Uncomment me for local development:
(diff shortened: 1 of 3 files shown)


travjenkins (approved):
lgtm

PR #180, opened 2022-06-02
Repair Entity Table State Defects.

kiahna-tucker:
## Changes

1. Resolve the component nesting error encountered when the entity table _appears_ to be in a loading state.

1. Display the _unmatched filter_ content on the entity table when a catalog namespace query string cannot be found.

1. Remove the authentication token from local storage when a user is unauthenticated.

**NOTE:** I am convinced `Typography` should wrap all instances of text in this application. Without supplying props, the component applies the `body1` styling to its children, which adjusts the font size, kerning, line height, etc. of a given string. Raw text (i.e., text without a `Typography` parent) is noticeably different and reduces the effectiveness of type manipulation in the global theme. More than happy to discuss this proposal in a more effective medium.

## Tests

## Screenshots

**Entity Table | Unmatched Filter**

pr-screenshot__145__entity-table-state-bugs__unmatched-filter


src/hooks/useShardsList.ts
```diff
@@ -56,8 +57,7 @@ const useShardsList = <T extends LiveSpecsExtBaseQuery>(specs: T[]) => {
                 .catch((error: string) => {
                     if (
                         error.includes(ErrorFlags.TOKEN_INVALID) ||
-                        error.includes(ErrorFlags.TOKEN_NOT_FOUND) ||
-                        error.includes(ErrorFlags.OPERATION_INVALID)
```
kiahna-tucker, line 60: Invalid operations should never result in the generation of a new auth gateway URL and token. This line was introduced as a temporary patch while an update to the `data-plane-gateway` was underway and can safely be removed now.

src/components/tables/EntityTable.tsx
```diff
@@ -353,7 +353,7 @@ function EntityTable({
                         </TableHead>
 
                         <TableBody>
-                            {selectData ? (
+                            {selectData && selectData.length > 0 ? (
```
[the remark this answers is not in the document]
travjenkins: Ahhh yeah - good catch. Why do we ever trust falsy 🤦

src/pages/Login.tsx
```diff
@@ -20,6 +21,8 @@ const urls = getUrls();
 const Login = () => {
     useBrowserTitle('browserTitle.login');
 
+    removeGatewayAuthConfig();
```
travjenkins, line 24: This should also use the useLocalStorage here.

travjenkins (commented):
please switch that last function to the ls hook

src/pages/Login.tsx
kiahna-tucker: Updated.

travjenkins (approved):
lgtm

Issue #162, opened 2022-05-26
Add new input type for Multi Line Secrets


travjenkins:
**Request**
Enable the UI to render a multi line secret (password textarea) for connector schemas.

**Reasoning**
There are some connectors that would be better served by this. An example is Google service account keys.

**Links**
https://github.com/estuary/ui/pull/158 : The PR that kicked off this discussion

travjenkins:
Multiline Secrets show as a normal password input with an extra button
![image](https://user-images.githubusercontent.com/270078/171741730-3bd5d178-8def-4ff3-b667-e4e402236d4c.png)

Upload opens dialog
![image](https://user-images.githubusercontent.com/270078/171741792-a75fa0f6-626e-40e7-9e9c-5af2ff983fd7.png)

Click opens file selector
![image](https://user-images.githubusercontent.com/270078/171741828-c34a02b3-3d8f-46a5-a45b-a75c8406aaaf.png)

You can also drag file
![image](https://user-images.githubusercontent.com/270078/171741875-74fea5ca-4923-4b65-b430-8bc28a970d2b.png)

Content of files (including new lines) are put into the inpu
![image](https://user-images.githubusercontent.com/270078/171741906-1fe36d9c-cef6-42d4-9056-6af26477983c.png)

Input saves values correctly (via upload OR pasting in the value)
![image](https://user-images.githubusercontent.com/270078/171741988-e4767071-2b24-4b59-8487-d39346364b1f.png)

PR #185, opened 2022-06-02
162: Adding support for multi line secrets in schemas

travjenkins:
## Changes

- Added support for a custom renderer for multiline secrets
- Handle pasting contents to no longer strip newlines
- Can upload files

## Tests

Manual

## Issues

Fixes: https://github.com/estuary/ui/issues/162

## Content

Content is hard coded as it is part of JSON Forms... but looking into that

## Screenshots

See [Issue 162](https://github.com/estuary/ui/issues/162)


src/forms/renderers/MultiLineSecret.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    '.cer',
+                    '.cr1',
+                    '.crt',
+                    '.p7c',
+                    '.p7b',
+                    '.p7r',
+                    '.der',
+                    '.pfx',
+                    '.spc',
+                ]}
```
travjenkins, line 105: @psFried - I'd like to chat with you above this before we launch MVP. I'm not totally sure what list of files we should  allow, should this be in the UI or the schema, should we add an option to the schema for "fileUpload", etc.

mdibaiee: @travjenkins AFAIK private ssh keys generated using `ssh-keygen` do not have an extension, they are just called something like `id_rsa`, so I'm not sure, but would this force the user to rename their private key to add an extension?

travjenkins: That's a good point. It would. So I think I will add no file type to the list as well.

kiahna-tucker (approved):

Issue #183, opened 2022-06-02
Total App Crash in Certain Workflows


dyaffe:
**Problem**
I'm seeing this in a few places today -- notably on collections details and in a few places when I've navigated around materializations.  The UI refreshes and the whole page goes white when doing certain workflows.

The easiest place to recreate it is going to collections and clicking on any of the details for collections that Will recently created.

travjenkins:
Simple fix - we are trying to use an object that does not exist

PR #186, opened 2022-06-03
Repair Issue Blocking Collections Details Panel Display

kiahna-tucker:
## Changes

1. Move details panel-related components to a shared location (i.e., _src/components/tables/Details_).

1. Rename the `CaptureDetails` component that was used across tables.

1. Create a component specifically for the logic that surfaces shard information in the details panel.

1. Resolve blocker for the collections details panel.

1. Added dynamic browser titles for the details panel. Previously, the opening of the details panel on the collections and materializations tables showed captures messaging.

**NOTE:** It appears that the replica processing errors no longer contain new lines. Therefore, they show up on one _long_ line in the editor. An issue should be created for that.

## Tests

## Issues

Fixed: #183 

## Content

Yes, the browser titles were expanded to include collections. Outside of that addition, only update to content IDs were adjusted which only impacts the codebase -- not the consumer-facing UI.

## Screenshots

**Details Panel | Collections**

pr-screenshot__146__details-panel-bugs__collections

<br />

**Details Panel | Captures**

pr-screenshot__146__details-panel-bugs__captures

<br />

**Details Panel | Materializations**

pr-screenshot__146__details-panel-bugs__materializations

<br />


src/components/tables/Captures/Rows.tsx
```diff
@@ -135,6 +135,7 @@ function Row({ isSelected, setRow, row, showEntityStatus }: RowProps) {
                 detailsExpanded={detailsExpanded}
                 id={row.last_pub_id}
                 colSpan={tableColumns.length}
+                browserTitleKey="captureDetails"
```
travjenkins, line 138: We should take this out in a follow up PR. I noticed the browserTitle when looking at details page and forgot that details started as a page. Now that it is a panel we don't need it.

src/components/tables/Details/ShardInformation.tsx
```diff
@@ -0,0 +1,139 @@
...
+                )
+            );
+        }
+    }, [specs, getShardDetails, setShardDetails, entityType]);
+
+    const handlers = {
+        mount: (editor: monacoEditor.editor.IStandaloneCodeEditor) => {
+            editorRef.current = editor;
+        },
+    };
```
travjenkins, line 63: Weird thought that never popped into my head until now... I'm not sure we need to create this ref since we don't use it here. I think we might be fine not handling mount when the editors are not used for editing.

travjenkins (approved):
lgtm - there are some tweaks we can handle later.

src/components/tables/Captures/Rows.tsx
kiahna-tucker: Follow up PR: #187

src/components/tables/Details/ShardInformation.tsx
kiahna-tucker: Follow up PR: #187

Issue #191, opened 2022-05-12
Date field not validating correct format - FB marketing


oliviamiannone:

travjenkins:
Now that the Facebook schema is updated pulling this into UI so we can fix it

PR #192, opened 2022-06-03
191: Adding in date formatting

travjenkins:
## Changes

Add formatting for JSON Schema forms that use date-time

## Tests

Manual

## Issues

https://github.com/estuary/ui/issues/191

## Content

None

## Screenshots

In the comments of the issue

### DIFF
--- a/src/services/jsonforms.ts
+++ b/src/services/jsonforms.ts
@@ -80,12 +80,23 @@ const addOption = (elem: ControlElement | Layout, key: string, value: any) => {
         elem.options = {};
     }
     elem.options[key] = value;
+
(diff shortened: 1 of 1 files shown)


src/services/jsonforms.ts
```diff
@@ -80,6 +80,8 @@ const addOption = (elem: ControlElement | Layout, key: string, value: any) => {
         elem.options = {};
     }
     elem.options[key] = value;
+
+    return elem;
```
travjenkins, line 84: Return object so we can throw more stuff in it

kiahna-tucker (approved):

PR #199, opened 2022-06-08
init: json forms test page

travjenkins:
## Changes

Add a JSON Forms test page to non-production 
Stopped using a function for `env` stuff since that won't change after start up

## Tests

Manual

## Issues

N/A

## Content

None - all content is hardcoded since it is a test page

## Screenshots

image

### DIFF
--- a/src/app/Authenticated.tsx
+++ b/src/app/Authenticated.tsx
@@ -13,12 +13,14 @@ import Admin from 'pages/Admin';
 import Captures from 'pages/Captures';
 import Collections from 'pages/Collections';
 import Connectors from 'pages/Connectors';
+import TestJsonForms from 'pages/dev/TestJsonForms';
 import PageNotFound from 'pages/error/PageNotFound';
 import Home from 'pages/Home';
 import Materializations from 'pages/Materializations';
 import Registration from 'pages/Registration';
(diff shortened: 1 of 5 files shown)


src/app/Authenticated.tsx
```diff
@@ -192,6 +194,12 @@ const Authenticated = () => {
                             path={routeDetails.admin.path}
                             element={<Admin />}
                         />
+                        {!isProduction && (
+                            <Route
+                                path="test/jsonforms"
+                                element={<TestJsonForms />}
+                            />
+                        )}
```
travjenkins, line 202: Making sure this does not show in Production... even though there isn't much risk if it did.

src/utils/env-utils.ts
```diff
@@ -10,13 +10,8 @@ declare global {
...
 
-export const isProduction = () => {
-    return process.env.NODE_ENV === 'production';
-};
-
-export const isTest = () => {
-    return process.env.NODE_ENV === 'test';
-};
+export const isProduction = process.env.NODE_ENV === 'production';
+export const isTest = process.env.NODE_ENV === 'test';
```
travjenkins, line 15: These don't need to be functions - variables are fine.

Issue #198, opened 2022-06-07
Cannot reset password


travjenkins:
Issue:
When someone tries to reset their password they are sent to the app but get the "Welcome" message because they are missing access grants.

Fix:
They need to land on a password reset page.

PR #200, opened 2022-06-08
[198] Add password reset... and add magic link auth

travjenkins:
## Changes

Initially this wast to add password reset. As I dove into handling the edge cases it seems like just using magic link auth is MUCH easier. I will leave the password reset stuff in there on the random off chance someone gets to it... somehow.

Moving towards the `notistack` library for notifications. Right now just login is using it. But next we'll convert the notification store to fire off requests to notistack.

Added an image to the Magic Link email

## Tests

Manual

## Issues

Fixes: https://github.com/estuary/ui/issues/198

## Content

Some new messages around logging in

## Screenshots

Edge case... someone gets password reset link
image

Missing email
image

Supabase returns error trying to send email
image

OIDC for Google failed
image

Not allowed to user magic link multiple times. We'll log you out if you try
image

Limit on number of magic link requests
image

Magic Link Email
image


src/app/Authenticated.tsx
```diff
@@ -76,9 +76,14 @@ export const routeDetails = {
...
-        title: 'routeTitle.registration',
-        path: '/register',
+    user: {
+        title: 'routeTitle.user',
+        path: '/user',
+        registration: {
+            title: 'routeTitle.registration',
+            path: 'register',
+            fullPath: '/user/register',
+        },
```
travjenkins, line 86: Initially this contained password reset but that was removed. I still think we'll need some kind of "user" path so leaving it

src/app/Authenticated.tsx
```diff
@@ -98,15 +103,21 @@ const Authenticated = () => {
...
+            <Routes>
+                <Route path={routeDetails.user.path}>
                     <Route
-                        path={routeDetails.registration.path}
+                        path={routeDetails.user.registration.path}
                         element={<Registration />}
                     />
+                </Route>
+                <Route path="*" element={<NoGrantsFound />} />
+            </Routes>
```
travjenkins, line 115: Making folks that hit "No Grants" able to land on the registration page as that will be who we need to "register"

src/app/Unauthenticated.tsx
```diff
@@ -2,19 +2,12 @@ import Login from 'pages/Login';
 import { Route, Routes } from 'react-router';
 
 export const logoutRoutes = {
-    path: 'logout',
-    params: {
-        reason: 'reason',
-    },
+    path: '',
```
travjenkins, line 5: Redid the reason thing cause it was a major pain and kinda weird to make the login page a notification controller.

src/components/login/MagicLink.tsx
```diff
@@ -0,0 +1,154 @@
...
+import React, { useState } from 'react';
+import { FormattedMessage, useIntl } from 'react-intl';
+import {
+    defaultOptions,
+    defaultRenderers,
+    hideValidation,
+    showValidation,
+} from 'services/jsonforms';
+
+const MagicLink = () => {
```
travjenkins, line 17: New form to handle submitting magic link requests.

src/components/login/MagicLink.tsx
```diff
@@ -0,0 +1,154 @@
...
+                description: intl.formatMessage({
+                    id: 'login.email.description',
+                }),
+                minLength: 5,
+                type: 'string',
+            },
+        },
+        required: ['email'],
+        type: 'object',
+    };
```
travjenkins, line 43: We might want to eventually store away common "specs" that could end up being used elsewhere.

src/components/login/MagicLink.tsx
```diff
@@ -0,0 +1,154 @@
...
+        );
+    };
+
+    const handlers = {
+        submit: async (event: React.FormEvent<HTMLFormElement>) => {
+            event.preventDefault();
+            if (!isEmpty(formErrors) || isEmpty(formData.email)) {
+                setShowErrors(true);
+                return;
+            }
```
travjenkins, line 79: Starting to like the pattern when using async of checking for errors and then returning if there are some.

src/components/login/MagicLink.tsx
```diff
@@ -0,0 +1,154 @@
...
+            setSubmitError(null);
+            setShowErrors(false);
+            setLoading(true);
+
+            const { error } = await supabaseClient.auth
+                .signIn({
+                    email: formData.email,
+                })
+                .finally(() => {
+                    setLoading(false);
```
travjenkins, line 90: Need to make sure we set off loading no matter what.

src/components/login/OIDCs.tsx
```diff
@@ -1,27 +1,61 @@
...
     const redirectTo = useConstant(
         () => `${window.location.origin}` // `${window.location.origin}${routeDetails.registration.path}`
     );
 
+    const supabaseClient = useClient();
+    const intl = useIntl();
+
+    const { enqueueSnackbar } = useSnackbar();
+
+    const loginFailed = (key: Provider) => {
```
travjenkins, line 19: All OIDCs can use the same general pattern when messaging that they were unable to login.

src/components/login/OIDCs.tsx
```diff
@@ -1,27 +1,61 @@
...
             sx={{
                 display: 'flex',
                 justifyContent: 'center',
             }}
         >
-            <GoogleAuthButton
-                onError={() => onError('login.loginFailed.google')}
-                redirectPath={redirectTo}
-            />
+            <GoogleButton onClick={() => login('google')} />
```
travjenkins, line 58: The wrapper that was there no longer was serving any purpose.

package.json
```diff
@@ -42,6 +42,7 @@
         "logrocket-react": "^5.0.1",
         "lru-cache": "^7.10.1",
         "monaco-editor": "^0.33.0",
+        "notistack": "^2.0.5",
```
travjenkins, line 45: New library for managing displaying notifications.

src/context/SWR.tsx
```diff
@@ -18,27 +16,32 @@ export const DEFAULT_POLLING = 2500;
...
+                            id: 'login.jwtExpired',
+                        }),
+                        {
+                            anchorOrigin: {
+                                vertical: 'top',
+                                horizontal: 'center',
+                            },
+                            variant: 'error',
+                        }
                     );
```
travjenkins, line 45: Just making the thing that is forcing the log out responsible for the notification. Makes a lot of stuff simpler. Also, with notistack we can have references to what opened it and put in custom actions so this empowers some interesting possibilities.

src/context/User.tsx
```diff
@@ -1,9 +1,39 @@
...
+                        id: 'login.passwordReset',
+                    }),
+                    {
+                        anchorOrigin: {
+                            vertical: 'top',
+                            horizontal: 'center',
+                        },
+                        variant: 'info',
+                    }
+                );
```
travjenkins, line 30: Super weird edge case but might as well catch it. If a user got a recovery email link just adding a notification that they should not need to do that.

src/context/User.tsx
```diff
@@ -1,9 +1,39 @@
...
+            } else if (event === 'SIGNED_OUT') {
+                navigate(logoutRoutes.path);
```
travjenkins, line 32: If the user signed out in any way navigate to the root path.

src/context/User.tsx
```diff
@@ -1,9 +1,39 @@
...
+                    {
+                        anchorOrigin: {
+                            vertical: 'top',
+                            horizontal: 'center',
+                        },
+                        variant: 'info',
+                    }
+                );
+            } else if (event === 'SIGNED_OUT') {
+                navigate(logoutRoutes.path, { replace: true });
```
travjenkins, line 32: If someone is logged out somehow we want to force them to the login path and replace any path they might be at.

src/services/ajv.ts
```diff
@@ -1,22 +1,25 @@
 import { createAjv } from '@jsonforms/core';
 
+// TODO (typing) Need to get this typed as the AJV Options type
+export const defaultAjvSettings: any = {
```
travjenkins, line 4: Need the default settings else where so making available

src/services/supabase.ts
```diff
@@ -21,6 +21,10 @@ const supabaseSettings = {
     anonKey: process.env.REACT_APP_SUPABASE_ANON_KEY,
 };
 
+// Related to the AUth stuff
+export const supabaseAuthError = 'error_description';
+export const userNotFound = 'User not found';
```
travjenkins, line 26: I want to find a better way to know these... might look into using the library from Supabase directly? Not sure yet.

kiahna-tucker (approved):

Issue #201, opened 2022-06-08
Reset entity table page number for each filter query.


kiahna-tucker:
The first table page should be presented to the user whenever the query string used to filter an entity table is edited, enabling them to view the query matches from the start. Currently, the table page is maintained whenever a filter is applied to the table. For example, if a user is on the second page of the table and applies a filter, they are shown the second page of query matches. If the number of query matches would result in a table with less pages than the current page number (e.g., the current page number was _two_ in the previous example), the _unmatched filter_ content is rendered despite the existence of query matches.

## Reference Images

**Maintained Page Number | Before Filter Application**

<img width="948" alt="issue-201__maintained-entity-table-page-number__maintained__before-filter" src="https://user-images.githubusercontent.com/77648584/172641294-464b13aa-7231-4eb2-8048-f8db0eebae1c.PNG">

<br />

**Maintained Page Number | After Filter Application**

<img width="960" alt="issue-201__maintained-entity-table-page-number__maintained__after-filter" src="https://user-images.githubusercontent.com/77648584/172641422-38c2a147-a6bc-4c21-971e-9605743d4010.PNG">

<br />

**Query with Matches | Searched on Page One**

<img width="960" alt="issue-201__maintained-entity-table-page-number__unmatched-filter__page-one" src="https://user-images.githubusercontent.com/77648584/172642865-e8a2e35a-46a2-4bfe-bbb8-2c29c071a654.PNG">

<br />

**Query with Matches | Searched on Page Two**

<img width="960" alt="issue-201__maintained-entity-table-page-number__unmatched-filter__page-two" src="https://user-images.githubusercontent.com/77648584/172642933-84a3c58e-844b-4255-9205-c77a3eeb1822.PNG">

<br />

PR #211, opened 2022-06-08
Reset Entity Table Page Upon Each Filter.

kiahna-tucker:
## Changes

Reset the entity table page whenever changes to the filter query string are detected.

## Tests

The following approaches to testing were performed:

1. Switch to a deeper page in the entity table, apply a filter that should have enough query matches to reach the new page, confirm the first page of query matches is presented.

1. Switch to a deeper page in the entity table, apply a filter that should **not** have enough query matches to reach the new page, confirm the first page of query matches is presented.

## Issues

#201


travjenkins (approved):
lgtm

Issue #219, opened 2022-06-14
Hide loading indicator when entity table updates due to a task transformation action.


kiahna-tucker:
When a task is successfully transformed (i.e., enabled, disabled, or deleted) and the entity table data is refreshed, the loading indicator is displayed while new data is fetched. The rows of the entity table should be updated seemingly seamlessly.

Issue #195, opened 2022-06-07
Refresh table after action


travjenkins:
Issue:
When a customer selects multiple rows (or a single row) and take an action like enable, disable, or delete when it is done the table shows old data

Fix:
When the action is complete the table should auto refresh itself.

PR #221, opened 2022-06-14
Refresh Entity Table After Task Action

kiahna-tucker:
## Changes

1. Create and expose a composite store dictionary consisting of the various _draft spec editor_ and _selectable table_ store instances via the _Zustand_ context provider.

1. Increase the scope of the _Zustand_ context provider so that it encapsulates all of the authenticated application.

1. Alter the `useZustandStore` hook to consume the name of the target slice of the composite store.

1. Extend the _selectable table_ store to include a counter that will increment with each successfully _transformed_ (i.e., enabled, disabled, or deleted) task.

1. Fetch updated entity table data when a task is successfully transformed without displaying the loading indicator.

**NOTE:** Changes to the Zustand store were made to enable this bug fix. Additional refactoring is needed.

## Tests

Manually tested the application by running through every function imaginable. Too many approaches to list.

## Issues

#195 
#219


travjenkins (approved):
looks fine

PR #222, opened 2022-06-14
Fix update entity query for specs

travjenkins:
## Changes

Need to query by the live spec ID to make sure we get the right spec

## Tests

Manual

## Issues

Note on the project board

## Content

None

## Screenshots

None

### DIFF
--- a/src/components/tables/RowActions/Shared/UpdateEntity.tsx
+++ b/src/components/tables/RowActions/Shared/UpdateEntity.tsx
@@ -43,10 +43,7 @@ function UpdateEntity({
     const [pubID, setPubID] = useState<string | null>(null);
     const [logToken, setLogToken] = useState<string | null>(null);
 
-    const { liveSpecs } = useLiveSpecsExtWithSpec(
-        entity.last_pub_id,
-        entity.spec_type
-    );
+    const { liveSpecs } = useLiveSpecsExtWithSpec(entity.id, entity.spec_type);
 
     useEffect(() => {
         const failed = (response: any) => {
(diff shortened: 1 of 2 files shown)


src/components/tables/RowActions/Shared/UpdateEntity.tsx
```diff
@@ -43,10 +43,7 @@ function UpdateEntity({
     const [pubID, setPubID] = useState<string | null>(null);
     const [logToken, setLogToken] = useState<string | null>(null);
 
-    const { liveSpecs } = useLiveSpecsExtWithSpec(
-        entity.last_pub_id,
-        entity.spec_type
-    );
+    const { liveSpecs } = useLiveSpecsExtWithSpec(entity.id, entity.spec_type);
```
travjenkins, line 46: Querying by the `last_pub_id` only works when users are just using the UI and only publishing things individually. This breaks right away if someone uses `flowctl` to publish multiple items at once. This _can_ cause the UI to fetch the wrong spec when trying to update the `shards` settings.

We should query by the `id` so we know we are fetching the correct spec

src/hooks/useLiveSpecsExt.ts
```diff
@@ -58,9 +58,7 @@ function useLiveSpecsExt(
 
                       return query
                           .eq('spec_type', specType)
-                          .or(
-                              `id.in.(${draftArray}),last_pub_id.in.(${draftArray})`
-                          );
+                          .or(`id.in.(${draftArray})`);
```
travjenkins, line 61: We no longer need to query by the `last_pub_id` as we were previously. Removing it so we only every query by `id` to make sure this hook is used for specific specs.

kiahna-tucker (approved):

Issue #217, opened 2022-06-10
Adding Collections to a materialization, naming them and then removing some prevents advance


dyaffe:
**Problem**
If I add collections to a materialization, name them and then remove some, there's a form error which prevents you from advancing.  Particularly annoying when you've just named 30 collections.

PR #223, opened 2022-06-15
Improve Materialization Resource Config Retention and Removal

kiahna-tucker:
## Changes

1. Remove the correct resource config data from the _materialization create_ store when a collection is removed from the selector.

1. Evaluate whether a given resource config contains data before adding it to the materialization definition.

1. Limit when resource config data in the _materialization create_ store is defaulted and/or reset.

## Tests

Iterate over the possible collection removal/addition and catalog generation sequences. Sample size of five collections.

Switch connector types after updating resource configurations.

## Issues

#217


travjenkins (approved):
lgtm

PR #225, opened 2022-06-15
Make Log Rocket better

travjenkins:
## Changes

1. No longer just totally remove the entire body from requests/responses. Created a white list of allow keys to be logged in the request/response
2. Clean up logging since this does go to Log Rocket
3. Add custom event for entity test/creation.

## Tests

Manual testing

## Issues

None - this is just some "tech debt" kinda work

## Content

N/A

### DIFF
--- a/src/components/capture/Create.tsx
+++ b/src/components/capture/Create.tsx
@@ -16,6 +16,7 @@ import { useZustandStore } from 'hooks/useZustand';
 import { useEffect } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { useNavigate } from 'react-router-dom';
+import { CustomEvents } from 'services/logrocket';
 import { startSubscription, TABLES } from 'services/supabase';
 import { entityCreateStoreSelectors, FormStatus } from 'stores/Create';
 import { getPathWithParam } from 'utils/misc-utils';
@@ -176,6 +177,7 @@ function CaptureCreate() {
                                 closeLogs={handlers.closeLogs}
                                 callFailed={helpers.callFailed}
                                 disabled={!hasConnectors}
+                                logEvent={CustomEvents.CAPTURE_TEST}
                             />
                         }
                         SaveButton={
@@ -184,6 +186,7 @@ function CaptureCreate() {
                                 callFailed={helpers.callFailed}
                                 disabled={!draftId}
                                 materialize={handlers.materializeCollections}
+                                logEvent={CustomEvents.CAPTURE_CREATE}
                             />
                         }
                     />
--- a/src/components/shared/Entity/Actions/Save.tsx
+++ b/src/components/shared/Entity/Actions/Save.tsx
@@ -8,7 +8,7 @@ import { useRouteStore } from 'hooks/useRouteStore';
 import { useZustandStore } from 'hooks/useZustand';
 import LogRocket from 'logrocket';
 import { FormattedMessage, useIntl } from 'react-intl';
-import { CustomEvents, MISSING } from 'services/logrocket';
+import { CustomEvents } from 'services/logrocket';
 import { endSubscription, startSubscription, TABLES } from 'services/supabase';
 import { entityCreateStoreSelectors, FormStatus } from 'stores/Create';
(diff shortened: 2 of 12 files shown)


src/components/capture/Create.tsx
```diff
@@ -179,6 +180,7 @@ function CaptureCreate() {
                                 closeLogs={handlers.closeLogs}
                                 callFailed={helpers.callFailed}
                                 disabled={!hasConnectors}
+                                logEvent={CustomEvents.CAPTURE_TEST}
```
travjenkins, line 183: Pass along the type of event was want to log when subscriptions are done

src/components/shared/Entity/Actions/Save.tsx
```diff
@@ -16,15 +18,27 @@ import useNotificationStore, {
 interface Props {
     disabled: boolean;
     onFailure: Function;
-    draftEditorStoreName: DraftEditorStoreNames;
+    logEvent: CustomEvents;
     dryRun?: boolean;
+    draftEditorStoreName: DraftEditorStoreNames;
 }
 
+const trackEvent = (logEvent: Props['logEvent'], payload: any) => {
```
travjenkins, line 26: Small helper function to call to LogRocket. I think eventually we should make some kind of service/map/typing/thingy where we can store the allowed events (`logEvent` down below) and their corresponding payloads.

package.json
```diff
@@ -35,6 +35,7 @@
         "ansicolor": "^1.1.100",
         "data-plane-gateway": "https://gitpkg.now.sh/estuary/data-plane-gateway/client/dist?main",
         "date-fns": "^2.28.0",
+        "filter-obj": "^3.0.0",
```
travjenkins, line 38: Library for easy implantation for an "allowed list"

src/forms/renderers/MultiLineSecret.tsx
```diff
@@ -85,25 +85,6 @@ const MultiLineSecretRenderer = (props: any) => {
...
-                    '.cr1',
-                    '.crt',
-                    '.p7c',
-                    '.p7b',
-                    '.p7r',
-                    '.der',
-                    '.pfx',
-                    '.spc',
-                    '',
-                ]}
```
travjenkins, line 106: @mdibaiee is of the opinion these are not really needed. So removing for now. I mainly added them to just be super extra safe but they mainly just get in the way.

src/services/logrocket.ts
```diff
@@ -20,10 +22,133 @@ interface Settings {
...
 
-const MASKED = '**MASKED**';
+type ParsedBody = [{ [k: string]: any }] | { [k: string]: any } | undefined;
+
+export enum CustomEvents {
+    CAPTURE_TEST = 'Capture_Test',
+    CAPTURE_CREATE = 'Capture_Create',
+    MATERIALIZATION_CREATE = 'Materialization_Create',
+    MATERIALIZATION_TEST = 'Materialization_Test',
+}
```
travjenkins, line 32: Current list of custom events for log rocket

src/components/shared/Entity/Actions/Save.tsx
```diff
@@ -69,16 +83,19 @@ function EntityCreateSave({
...
     const waitForPublishToFinish = (logTokenVal: string) => {
         resetFormState(status);
-        console.log('wait for finish');
         const subscription = startSubscription(
             supabaseClient.from(
                 `${TABLES.PUBLICATIONS}:draft_id=eq.${draftId}`
             ),
             async (payload: any) => {
                 if (payload.logs_token === logTokenVal) {
+                    const formStatus = dryRun
```
kiahna-tucker, line 92: Is this used in more than one place? If not, why is the value being stored in a constant?

kiahna-tucker (approved):

src/components/shared/Entity/Actions/Save.tsx
travjenkins: It was for a minute and then it wasn't and was too lazy to put it back how it was.

PR #237, opened 2022-06-20
Add OTP login method

travjenkins:
## Changes

We need this to enable local testing through VMs. It is also potentially how a user may want to use so it does add value for those users.

## Tests

Manual

## Issues

N/A

## Content

Some new messages

## Screenshots

Default view
image

After clicking the `already have a...` cta
image

### DIFF
--- a/src/components/login/MagicLinkInputs.tsx
+++ b/src/components/login/MagicLinkInputs.tsx
@@ -1,0 +1,142 @@
+import { JsonSchema } from '@jsonforms/core';
+import { materialCells } from '@jsonforms/material-renderers';
+import { JsonForms } from '@jsonforms/react';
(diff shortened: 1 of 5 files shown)


src/components/login/MagicLinkInputs.tsx
```diff
@@ -0,0 +1,144 @@
...
+    showValidation,
+} from 'services/jsonforms';
+
+interface Props {
+    onSubmit: Function;
+    schema: JsonSchema;
+    uiSchema: any; //UISchemaElement
+}
+
+const MagicLinkInputs = ({ onSubmit, schema, uiSchema }: Props) => {
```
travjenkins, line 23: Used to show the form and error handling so that stuff isn't repeated.

Issue #236, opened 2022-06-20
Monaco editor having issues on Windows


travjenkins:
When you select a file in the catalog editor the on change event fires off only on Edge for Windows. This does NOT happen on Edge for Linux.

![image](https://user-images.githubusercontent.com/270078/174675281-d7eb4834-3a6d-4282-a903-17987624c053.png)

travjenkins:
Updated the title as this is an issue with Chrome as well.

PR #238, opened 2022-06-21
Fix editor for Windows

travjenkins:
## Changes

No longer need to stringify the value every render... only when there is a new spec.
Updated the library version

## Tests

Manual - clicked between files as fast as I could

## Issues

Fixes: https://github.com/estuary/ui/issues/236

## Content

N/A

## Screenshots

N/A

### DIFF
--- a/src/components/editor/MonacoEditor.tsx
+++ b/src/components/editor/MonacoEditor.tsx
@@ -5,16 +5,15 @@ import Saved from 'components/editor/Status/Saved';
 import Saving from 'components/editor/Status/Saving';
 import ServerDiff from 'components/editor/Status/ServerDiff';
 import { EditorStatus, EditorStoreState } from 'components/editor/Store';
-import { DraftSpecQuery } from 'hooks/useDraftSpecs';
 import {
     DraftEditorStoreNames,
     LiveSpecEditorStoreNames,
     UseZustandStore,
 } from 'context/Zustand';
+import { DraftSpecQuery } from 'hooks/useDraftSpecs';
 import { debounce } from 'lodash';
 import * as monacoEditor from 'monaco-editor/esm/vs/editor/editor.api';
-import { useCallback, useRef, useState } from 'react';
-import { useIntl } from 'react-intl';
+import { useCallback, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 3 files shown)


src/components/editor/MonacoEditor.tsx
```diff
@@ -119,10 +117,15 @@ function MonacoEditor({
...
-    const debouncedChange = useCallback(debounce(updateValue, 1000), [
-        currentCatalog,
+    const debouncedChange = useCallback(debounce(updateValue, 750), [
+        currentCatalog.catalog_name,
     ]);
 
+    const specAsString = useMemo(
+        () => stringifyJSON(currentCatalog.spec),
+        [currentCatalog.spec]
+    );
```
travjenkins, line 127: The stringify isn't a _huge_ calculation but it seems this was messing stuff up for Windows... potentially messed it up in any OS and just saw it on Windows because it is slower due to running in a VM for me.

src/components/editor/MonacoEditor.tsx
```diff
@@ -138,7 +141,7 @@ function MonacoEditor({
         },
     };
 
-    if (currentCatalog?.spec) {
+    if (currentCatalog.catalog_name && specAsString) {
```
travjenkins, line 144: We only want to render if we have both the name and the string version of the spec

src/components/editor/MonacoEditor.tsx
```diff
@@ -187,10 +190,9 @@ function MonacoEditor({
                         theme={
                             theme.palette.mode === 'light' ? 'vs' : 'vs-dark'
                         }
-                        defaultValue={intl.formatMessage({
-                            id: 'common.loading',
-                        })}
```
travjenkins, line 192: Setting the default value to loading was a bad idea and caused a weird UX. A user could `ctrl+z` back to the loading value.

src/components/editor/MonacoEditor.tsx
```diff
@@ -187,10 +190,9 @@ function MonacoEditor({
...
-                        value={stringifyJSON(currentCatalog.spec)}
+                        saveViewState={false}
```
travjenkins, line 193: We do not need the state saved since we have a store and server handling stuff for us.

Issue #261, opened 2022-07-21
Elasticsearch resource binding JSONForms does not render Field Overrides correctly


mdibaiee:
Hello!

Elasticsearch resource bindings have a field called Field Overrides with this spec:
```
"resourceSpecSchema": {
    "$schema": "http://json-schema.org/draft-04/schema#",
    "required": [
      "index",
      "delta_updates"
    ],
    "properties": {
      "index": {
        "type": "string",
        "description": "Name of the ElasticSearch index to store the materialization results."
      },
      "delta_updates": {
        "type": "boolean"
      },
      "field_overrides": {
        "items": {
          "required": [
            "pointer",
            "es_type"
          ],
          "properties": {
            "pointer": {
              "type": "string",
              "description": "A '/'-delimitated json pointer to the location of the overridden field."
            },
            "es_type": {
              "required": [
                "field_type"
              ],
              "properties": {
                "field_type": {
                  "type": "string",
                  "description": "The elastic search field data types. Supported types include: boolean, date, double, geo_point, geo_shape, keyword, long, null, text."
                },
                "date_spec": {
                  "required": [
                    "format"
                  ],
                  "properties": {
                    "format": {
                      "type": "string",
                      "description": "Format of the date. See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html."
                    }
                  },
                  "additionalProperties": false,
                  "type": "object",
                  "description": "Spec of the date field, effective if field_type is 'date'. See https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html"
                },
                "keyword_spec": {
                  "required": [
                    "ignore_above"
                  ],
                  "properties": {
                    "ignore_above": {
                      "type": "integer",
                      "description": "Strings longer than the ignore_above setting will not be indexed or stored. See https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html"
                    }
                  },
                  "additionalProperties": false,
                  "type": "object",
                  "description": "Spec of the keyword field, effective if field_type is 'keyword'. See https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html"
                },
                "text_spec": {
                  "required": [
                    "dual_keyword",
                    "keyword_ignore_above"
                  ],
                  "properties": {
                    "dual_keyword": {
                      "type": "boolean",
                      "description": "Whether or not to specify the field as text/keyword dual field."
                    },
                    "keyword_ignore_above": {
                      "type": "integer",
[...]

Issue #250, opened 2022-07-11
Secret upload button disappears when navigating tabs


travjenkins:
Problem:
When a multiline secret is displayed in the basic info section it disappears when navigating between tabs. This does not happen for the multiline secrets in Network Tunneling for some reason.

Details:
OS: Pop_Os
Browser: Chrome

Screenshots:

Initial Load
![image](https://user-images.githubusercontent.com/270078/178335772-877be587-97e2-4636-8fb6-5de20957e4df.png)

Coming back from Network Tunneling
![image](https://user-images.githubusercontent.com/270078/178335797-d5cd0900-c6a2-4eb5-8d9e-68fbb9e1c5b5.png)

Issue #248, opened 2022-07-08
Unable to clear the connector field of the details form.


kiahna-tucker:
When the last character of a selected connector is deleted from the details form, the application breaks and a white page is displayed in the browser.

### Recreation Steps

1. Enter the capture or materialization creation flow to display the details form.

1. Select an option from the connector menu.

1. Remove the selected connector by highlighting the input and deleting all characters at once or removing characters in sets. When the final character is removed from the connector field, the application breaks.

### Screenshots

**Connector Selected**

<img width="1108" alt="issue-248__clear-connector-field__selection" src="https://user-images.githubusercontent.com/77648584/178017987-00203642-0654-456d-ba06-40189ff5a3aa.PNG">

<br />

**Connector Selection Deleted**

<img width="1909" alt="issue-248__clear-connector-field__post-deletion" src="https://user-images.githubusercontent.com/77648584/178018126-c2f92864-eec2-4905-8485-e18a3a67a6c7.PNG">

<br />

### System Information

| | Version |
| --- | --- |
| Operating System | Windows 10.0.19044 |
| Browser | Chrome 103.0.5060.114 (Official Build) (64-bit) |

Issue #256, opened 2022-07-18
UI Hangs instead of Showing Error


ytang07:
UI Hangs when trying to add connector when there is error 
```
time="2022-07-18T20:16:15Z" level=warning msg="Encountered 403 Forbidden with reason \"PERMISSION_DENIED\"" logSource="connector stderr" operation=discover
[1:39](https://estuaryworkspace.slack.com/archives/D03NY1VEM1T/p1658176753287659)
time="2022-07-18T20:16:15Z" level=error msg="Giving up get(...) after 1 tries (googleapiclient.errors.HttpError: <HttpError 403 when requesting https://sheets.googleapis.com/v4/spreadsheets/1l4ZSYtclekdqmj5feG0eCXl5f8kQ9DrcRQptPL20g4g?includeGridData=false&alt=json returned \"Google Sheets API has not been used in project 368839698543 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/sheets.googleapis.com/overview?project=368839698543 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.\". Details: \"[{'@type': '[type.googleapis.com/google.rpc.Help](http://type.googleapis.com/google.rpc.Help)', 'links': [{'description': 'Google developers console API activation', 'url': '[https://console.developers.google.com/apis/api/sheets.googleapis.com/overview?project=368839698543'}]}](https://console.developers.google.com/apis/api/sheets.googleapis.com/overview?project=368839698543%27}]}), {'@type': '[type.googleapis.com/google.rpc.ErrorInfo](http://type.googleapis.com/google.rpc.ErrorInfo)', 'reason': 'SERVICE_DISABLED', 'domain': '[googleapis.com](http://googleapis.com/)', 'metadata': {'consumer': 'projects/368839698543', 'service': '[sheets.googleapis.com](http://sheets.googleapis.com/)'}}]\">)" logSource="connector stderr" operation=discover
```

from Dave: the way I had to get this data was joining from drafts to discover on draft id to log line on token ID

`select ll.log_line from discovers as d join internal.log_lines as ll on ll.token = d.logs_token where d.draft_id = '06:c5:c2:d5:63:6f:6c:00'`

dyaffe:
This happened for Yujian when trying to perform a capture from sheets with invalid credentials.  The error was there, but there was no feedback in the UI and it hung forever.

travjenkins:
[...]

PR #262, opened 2022-07-21
Fix multi line secrets, some logging, and Forms UX

travjenkins:
## Changes

1. Multiline secrets now work when switching form categories.
2. Added some LogRocket logging
3. No longer show group wrapper
4. Can clear out connector field and not crash app

## Tests

Manual tests

## Issues

Fixes: https://github.com/estuary/ui/issues/250 
Fixes: https://github.com/estuary/ui/issues/261
Fixes: https://github.com/estuary/ui/issues/248 
Progress: https://github.com/estuary/ui/issues/256

## Content

None

### No longer displaying group wrapper in category
Before
image

After
image

### Scrolling in the resource config
image

### Able to delete items from list in config
image


src/forms/renderers/MultiLineSecret.tsx
```diff
@@ -7,17 +7,19 @@ import { withJsonFormsControlProps } from '@jsonforms/react';
...
+import { Formats, Options } from 'types/jsonforms';
 
 export const multiLineSecretTester: RankedTester = rankWith(
-    999,
-    and(optionIs('format', MULTI_LINE_SECRET))
+    10,
+    and(
+        optionIs(Options.format, Formats.password),
+        optionIs(Options.multi, true)
+    )
```
travjenkins, line 17: Better way to test for this as using a custom format caused issues trying to get the text input to be `password`.

src/forms/renderers/MultiLineSecret.tsx
```diff
@@ -47,9 +49,6 @@ const MultiLineSecretRenderer = (props: any) => {
         [handleChange, path]
     );
 
-    // Make it a password again so the input is masked
-    uischema.options.format = 'password';
-
```
travjenkins, line 52: This is what caused the secret button to disappear. After a single render altering the uischema breaks it forever. This was a very flawed approach to this problem.

src/services/jsonforms.ts
```diff
@@ -153,7 +152,7 @@ export const generateCategoryUiSchema = (uiSchema: any) => {
         elements: [
             {
                 type: 'Category',
-                label: 'Basic Information',
+                label: 'Basic Config',
```
travjenkins, line 155: Minor tweak to some content

src/services/jsonforms.ts
```diff
@@ -399,23 +400,28 @@ const generateUISchema = (
...
+            newControl.options.dateTimeSaveFormat = Patterns.dateTime;
+        }
+    } else {
+        if (isSecretText(jsonSchema)) {
+            addOption(controlObject, Options.format, Formats.password);
+        }
+        if (isMultilineText(jsonSchema)) {
+            addOption(controlObject, Options.multi, true);
         }
     }
```
travjenkins, line 426: We need to set `format` AND `multi` now so they are both stand alone ifs.

src/types/jsonforms.ts
```diff
@@ -0,0 +1,18 @@
+// THIS IS NOT AN EXHAUSTIVE LIST
```
travjenkins, line 1: Threw this together mainly just for the multiline secret. These values are important for our app and JSONForms... so not storing them in the component that uses it.

src/components/editor/ListAndDetails.tsx
```diff
@@ -48,7 +48,7 @@ function ListAndDetails({ list, details, height }: Props) {
                     className="right-pane"
                     minSize={MIN_RESIZE_WIDTH}
                     style={{
-                        overflow: 'hidden',
+                        overflow: 'auto',
```
travjenkins, line 51: need to be able to scroll... no clue why this was `hidden` to start with

src/components/shared/Entity/DetailsForm.tsx
```diff
@@ -145,7 +145,7 @@ function DetailsForm({
                 elements: [
                     {
                         label: intl.formatMessage({
-                            id: 'connector.label',
+                            id: 'entityCreate.connector.label',
```
travjenkins, line 148: Switched label from a common one to one that adds "search" to it

src/context/Theme.tsx
```diff
@@ -258,7 +264,7 @@ const themeSettings = createTheme({
                         width: '100%',
                         zIndex: buttonHoverIndex,
                     },
-                    '& 	.MuiChip-deleteIcon': {
+                    '& .MuiChip-deleteIcon': {
```
travjenkins, line 267: There was a tab here so removed it.

src/context/Theme.tsx
```diff
@@ -340,6 +346,13 @@ const ThemeProvider = ({ children }: BaseComponentProps) => {
             components: {
                 ...themeSettings.components,
                 MuiAccordion: {
+                    defaultProps: {
+                        sx: {
+                            '& .MuiIconButton-root': {
+                                zIndex: accordianButton,
+                            },
+                        },
+                    },
```
travjenkins, line 355: This makes sure any button inside an accordion is usable.

src/components/capture/Create.tsx
```diff
@@ -23,6 +24,16 @@ import { getPathWithParam } from 'utils/misc-utils';
 
 const connectorType = 'capture';
 
+const trackEvent = (payload: any) => {
```
kiahna-tucker, line 27: Did you run into an issue narrowing the type of the `payload` argument?

src/components/shared/Entity/DetailsForm.tsx
kiahna-tucker: Curious if placeholders in (certain) form fields should be a short statement that begins with a verb. For example, an active GitHub comment text field without text uses _Leave a comment_ as its placeholder text. Wondering if a similar approach would be helpful here (e.g., _Select a connector_). Thinking out loud.

src/context/Theme.tsx
```diff
@@ -134,11 +134,17 @@ const darkMode: PaletteOptions = {
...
+// Need to make the sticky header be on top
+export const stickyHeaderIndex = zIndexIncrement * 20;
+export const headerLinkIndex = zIndexIncrement * 30;
+
+// Makes sure the hovering styling can be seen
+const buttonHoverIndex = zIndexIncrement * 4;
+// To make the delete in a multi select to work
 const chipDeleteIndex = buttonHoverIndex + zIndexIncrement;
+// JSONForms accordian is hardcoded to 20 so making this "1 higher"
+const accordianButton = zIndexIncrement * 5;
```
kiahna-tucker, line 147: There is a typo in this name of this constant: `accordianButton` should be `accordionButton`. This misspelling applies to the related comment as well.

kiahna-tucker (approved):
Approved with comments. Would be great if the typo could be addressed before merging.

src/components/capture/Create.tsx
travjenkins: We just don't have the type created yet for the response of a subscription start.

src/components/shared/Entity/DetailsForm.tsx
travjenkins: For this field we specifically want to let them know they can search/type in the connector name.

src/context/Theme.tsx
```diff
@@ -134,11 +134,17 @@ const darkMode: PaletteOptions = {
...
+
+// Need to make the sticky header be on top
+export const stickyHeaderIndex = zIndexIncrement * 20;
+export const headerLinkIndex = zIndexIncrement * 30;
+
+// Makes sure the hovering styling can be seen
+const buttonHoverIndex = zIndexIncrement * 4;
+// To make the delete in a multi select to work
 const chipDeleteIndex = buttonHoverIndex + zIndexIncrement;
+// JSONForms accordi0n is hardcoded to 20 so making this "1 higher"
```
kiahna-tucker, line 146: This made me chuckle. I think you accidentally hit the _zero_ key when going for the _o_ key.

travjenkins: damn it

travjenkins: fixed the typo fix typo

PR #267, opened 2022-07-27
removing connector tag filtering

travjenkins:
## Changes

Removing filter of connector tags based on the version. We are moving towards the UI always using the latest version.

## Tests

Manual tested

## Issues

None - discussed via meeting

## Content

None

## Screenshots

Tested with these settings in my local DB
image


kiahna-tucker (approved):

mdibaiee:
I deleted the `:dev` rows, merging this to avoid an inconsistent behaviour.

PR #280, opened 2022-08-10
content - fix mismatched terms

oliviamiannone:
## Changes

Noticed the messaging above the catalog editor after catalog generation is inconsistent with the current button names (there is no "regenerate catalog" button). This is just a quick fix to the content file to address that. 
image

## Tests

na

## Issues

na

## Content

content fix only, see above 

## Screenshots


travjenkins (approved):
lgtm

Issue #270, opened 2022-08-01
Creation flow UI elements overlap the sticky header.


kiahna-tucker:
When the create header is positioned over the Monaco editor, the white, rectangular divider separating the file selector and editor is positioned above the create header. Additionally, the _Connector Help_ button also has a higher z-index than the create header.

### Recreation Steps

1. Enter the capture or materialization creation flow.

1. Enter information into the necessary forms, then click the _Discover Endpoint_ button.

1. Scroll down until the create header is positioned over the endpoint config accordion summary.

1. Scroll down until the create header is positioned over the Monaco editor.

### Screenshots

**Materialization Create Flow | Endpoint Config Accordion Summary**

<img width="571" alt="issue-270__connector-help-button-overlaps-create-header" src="https://user-images.githubusercontent.com/77648584/182210984-af5426f9-5923-447e-84c1-f40a5e8e44ec.PNG">

<br />

**Materialization Create Flow | Monaco Editor**

<img width="572" alt="issue-270__monaco-editor-divider-overlaps-create-header" src="https://user-images.githubusercontent.com/77648584/182204590-17bd8444-fb70-49a2-b948-4ad4a50966cb.PNG">

### System Information

| | Version |
| --- | --- |
| Operating System | Windows 10.0.19044 |
| Browser | Chrome 103.0.5060.134 (Official Build) (64-bit) |

PR #283, opened 2022-08-11
Fix sticky header and pre-population for create

travjenkins:
## Changes

1. Fix index issue with the sticky header
2. Populate the preselected connector with all the details. This is to prevent the selection from displaying while the store and validator do not have the proper details.

## Tests

Manually tested

## Issues

Fixes: https://github.com/estuary/ui/issues/270
Fixes: https://github.com/estuary/ui/issues/284

## Content

None

## Screenshots

None

### DIFF
--- a/src/components/shared/Entity/DetailsForm.tsx
+++ b/src/components/shared/Entity/DetailsForm.tsx
@@ -28,6 +28,7 @@ import {
 import { useEndpointConfigStore_reset } from 'stores/EndpointConfig';
 import { EntityFormState } from 'stores/FormState';
 import { Grants } from 'types';
+import { hasLength } from 'utils/misc-utils';
 
 interface Props {
     connectorTags: ConnectorWithTagDetailQuery[];
@@ -35,6 +36,15 @@ interface Props {
     draftEditorStoreName: DraftEditorStoreNames;
     formStateStoreName: FormStateStoreNames;
 }
+
+const getConnectorImageDetails = (connector: ConnectorWithTagDetailQuery) => {
+    return {
+        connectorId: connector.id,
+        id: connector.connector_tags[0].id,
+        imagePath: `${connector.image_name}${connector.connector_tags[0].image_tag}`,
+        iconPath: connector.image,
+    };
+};
 
 function DetailsForm({
     connectorTags,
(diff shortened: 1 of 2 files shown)


src/components/shared/Entity/DetailsForm.tsx
```diff
@@ -128,12 +122,7 @@ function DetailsForm({
         if (connectorTags.length > 0) {
             connectorTags.forEach((connector) => {
                 response.push({
-                    const: {
-                        connectorId: connector.id,
-                        id: connector.connector_tags[0].id,
-                        imagePath: `${connector.image_name}${connector.connector_tags[0].image_tag}`,
-                        iconPath: connector.image,
-                    },
+                    const: getConnectorImageDetails(connector),
```
travjenkins, line 125: Use the new function to get the values

src/components/shared/Entity/DetailsForm.tsx
```diff
@@ -89,20 +99,19 @@ function DetailsForm({
...
+        if (connectorID && hasLength(connectorTags)) {
+            connectorTags.forEach((connector) => {
+                if (connector.connector_tags[0].id === connectorID) {
+                    setDetails({
+                        data: {
+                            entityName: '',
+                            connectorImage: getConnectorImageDetails(connector),
+                        },
+                    });
+                }
```
travjenkins, line 111: Now we'll wait until we have connectorTags and _then_ set the prefilled connector. This allows us to populate all the props we need.

src/components/shared/Entity/DetailsForm.tsx
```diff
@@ -36,6 +37,15 @@ interface Props {
...
 }
 
+const getConnectorImageDetails = (connector: ConnectorWithTagDetailQuery) => {
+    return {
+        connectorId: connector.id,
+        id: connector.connector_tags[0].id,
+        imagePath: `${connector.image_name}${connector.connector_tags[0].image_tag}`,
+        iconPath: connector.image,
+    };
+};
```
travjenkins, line 47: One day was might want to make some pattern or something for mapping but right now this seems fine.

src/context/Theme.tsx
```diff
@@ -136,7 +136,7 @@ const darkMode: PaletteOptions = {
 export const zIndexIncrement = 5;
 
 // Need to make the sticky header be on top
-export const stickyHeaderIndex = zIndexIncrement * 20;
+export const stickyHeaderIndex = zIndexIncrement * 100;
```
travjenkins, line 139: cranking this up to 100 even though it just needed to be above the line below. However, I feel like if we stick with increments of 5/10 we'll be good for a long time until we need to worry about anything being above 100.

kiahna-tucker (approved):

Issue #240, opened 2022-06-22
Table rows stay selected after leaving page


travjenkins:
Issue
If you select a row(s) in the Captures table and then materialize the row - when you come back to the captures table the same row(s) are selected.

Resolution
Should have the row selected cleared anytime a user navigates away from the page.

PR #285, opened 2022-08-12
Clear Selected Table Rows when Navigation Changes.

kiahna-tucker:
## Changes

Clear any selected entity table rows when navigating away from the current page.

## Tests

Selected rows from one entity page (i.e., captures or materializations), switched to the other entity page, then returned to the original entity page and observed that the selection cleared.

Evaluated that existing selection functionality is still intact.

## Issues

Complete: #240

### DIFF
--- a/src/components/tables/EntityTable.tsx
+++ b/src/components/tables/EntityTable.tsx
@@ -39,6 +39,7 @@ import {
     useState,
 } from 'react';
 import { FormattedMessage, useIntl } from 'react-intl';
+import { useEffectOnce } from 'react-use';
 import {
     SortDirection,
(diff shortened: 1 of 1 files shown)


src/components/tables/EntityTable.tsx
```diff
@@ -166,6 +167,12 @@ function EntityTable({
         }
     };
 
+    useEffectOnce(() => {
+        return () => {
+            return resetSelection();
```
kiahna-tucker, line 172: When the `EntityTable` component unmounts, reset the `setAll` component state value and remove the rows identified in the Select Table store.

Issue #245, opened 2022-07-01
Unknown connectors cause the UI to break


psFried:
It's possible to use the CLI to create captures and materializations using connectors that are completely unknown to the control-plane/UI. Doing so, seems to completely break the UI, though.

Steps:

- Use the CLI to publish a materialization using a connector that isn't known to the control plane.
- Go to the UI and open the materializations page
- :boom: The UI goes all blank, and there is an error like the one below in the console

<details>
<summary>Console error</summary>
```
Uncaught TypeError: connector is null
    connectorName ConnectorName.tsx:14
    useConstant use-constant.esm.js:8
    ConnectorName ConnectorName.tsx:14
    React 12
    unstable_runWithPriority scheduler.development.js:468
    React 5
    setState index.mjs:567
    finishRequestAndUpdateState index.mjs:733
    revalidate index.mjs:860
    step index.mjs:50
    verb index.mjs:31
    fulfilled index.mjs:21
    promise callback*step index.mjs:23
    __awaiter index.mjs:24
    __awaiter index.mjs:20
    revalidate index.mjs:700
    onRevalidate index.mjs:937
    broadcastState index.mjs:263
    internalMutate index.mjs:300
    step index.mjs:50
    verb index.mjs:31
    __awaiter index.mjs:24
    __awaiter index.mjs:20
    internalMutate index.mjs:280
    EntityTable EntityTable.tsx:178
    React 5
    unstable_runWithPriority scheduler.development.js:468
    React 4
    unstable_runWithPriority scheduler.development.js:468
    React 17
    tsx index.tsx:25
    factory react refresh:6
    Webpack 3
2 ConnectorName.tsx:14
The above error occurred in the <ConnectorName> component:

ConnectorName@http://localhost:3000/static/js/bundle.js:1484:7
div
withEmotionCache/<@http://localhost:3000/static/js/bundle.js:27880:66
Box@http://localhost:3000/static/js/bundle.js:86013:72
Tooltip@http://localhost:3000/static/js/bundle.js:75337:83
td
withEmotionCache/<@http://localhost:3000/static/js/bundle.js:27880:66
TableCell@http://localhost:3000/static/js/bundle.js:71714:83
Connector@http://localhost:3000/static/js/bundle.js:16275:7
tr
withEmotionCache/<@http://localhost:3000/static/js/bundle.js:27880:66
TableRow@http://localhost:3000/static/js/bundle.js:73053:82
[...]

PR #289, opened 2022-08-22
Make open graph title optional for capture connectors

psFried:
## Changes

Previously, the open graph title was used exclusively as the display
name of a connector on the captures table page, which relies on the
assumption that open graph data will always be present for every capture
displayed in the UI. This assumption is not always valid, since we have
the ability to use custom images via the CLI. The result is that the
captures page would show as totally blank when any capture used a
connector that didn't have an open graph title.

This commit updates the capture rows to fallback to using the connector
image if the open graph title is not present. It also updates the query
to retrieve the image directly from the capture spec instead of the open
graph. The result is that the captures table can now render when a
capture uses a connector that isn't present in the database.

## Tests

I only tested this manually. Please let me know if there's a better way.

## Issues

This partially addresses #245. If the approach seems reasonable, then I'll put up a follow-on PR to apply the same fix to the materializations table.

## Content

n/a

## Screenshots

Note the first capture, which uses the image name instead of the open graph title.

image

### DIFF
--- a/src/components/tables/Captures/index.tsx
+++ b/src/components/tables/Captures/index.tsx
@@ -15,7 +15,7 @@ const queryColumns = [
     'catalog_name',
     'connector_image_name',
     'connector_image_tag',
-    'connector_open_graph->en-US->>image',
+    'spec->endpoint->connector->image',
     'connector_open_graph->en-US->>title',
     'id',
     'last_pub_id',
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,9 @@ jobs:
             - name: Get Deps
               run: npm install
 
+            - name: Print Versions
+              run: npm run print-versions
+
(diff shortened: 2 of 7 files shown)


src/components/tables/Captures/index.tsx
```diff
@@ -15,7 +15,7 @@ const queryColumns = [
     'catalog_name',
     'connector_image_name',
     'connector_image_tag',
-    'connector_open_graph->en-US->>image',
```
travjenkins, line 18: This is not the image path in terms of what docker image to pull... this is the image that we display as the icon.

psFried:
> @psFried - do we still want to delete this PR?

Yeah, I'm going to close this for now. The plan is to remove the open-graph fetching entirely, in favor of explicit declaration of the required fields. So we may have a better path forward on this in the near future.

PR #298, opened 2022-08-23
feature: Implement "preview" for collection data

jshearer:
## Changes

This is the outcome of my hackathon work around presenting "sample"/preview data for collections. The data can be presented either as a table or a master/detail view presenting pretty-printed JSON data. Based on my conversation with @jgraettinger, I set the default to JSON view.

**Question:** Do we just always want JSON view? Or should we try to be smart and switch view e.g based on size of documents, or something like that?

## Screenshots

Screen Shot 2022-08-23 at 11 11 18
Screen Shot 2022-08-23 at 11 12 11


src/components/collection/DataPreview/TableView.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+
+function TableView({
+    journalData: { data, error },
+    spec,
+}: PreviewTableModeProps) {
+    const specEntries = useMemo(
+        // TODO (typing) we need to fix typing
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        () => Object.entries(spec?.spec?.schema?.properties ?? {}),
```
psFried, line 36: Many schemas won't have a top level properties that's useful for this. For example, you may use an allOf to compose multiple schemas, which is actually how we handle composing in the `_meta` schemas. It might be good to replace this schema iteration with a call to fetch projections, once that's supported. All good for now, though, since the JSON view is the default. Just sharing my thoughts on it.

src/components/collection/DataPreview/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const spec = useMemo(() => publicationSpecs[0], [publicationSpecs]);
+
+    // TODO (typing) we need to fix typing
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    const journals = useJournalsForCollection(spec?.catalog_name);
+    const { data: journalsData, isValidating: journalsLoading } = journals;
+
+    // TODO (typing) we need to fix typing
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    const journal = useMemo(() => journalsData?.journals?.[0], [journalsData]);
```
psFried, line 48: If the collection is partitioned, this will only select a single partition. This is absolutely a reasonable compromise for now. IMO this type of thing is worth a comment, though, since otherwise it just looks "wrong".

src/hooks/useJournalData.ts
```diff
@@ -0,0 +1,252 @@
...
+            if (done) return;
+            // Else yield the chunk
+            yield value;
+        }
+    } finally {
+        reader.releaseLock();
+    }
+}
+
+const INCREMENT = 1024 * 10;
```
psFried, line 120: nit: this name is really vague, and mixing of the power of 2 and 10 makes it seem like the specific value is really important. Maybe add a comment?

src/hooks/useJournalData.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    const [refreshCount, setRefresh] = useState(0);
+
+    const [data, setData] = useState<JournalRecord[]>([]);
+    const [error, setError] = useState<any>(null);
+    const [loading, setLoading] = useState(false);
+
+    useEffect(() => {
+        void (async () => {
+            if (journalName && journalClient) {
```
psFried, line 226: It seems a little weird to me that we handle the case where `journalName` is null all the way down here. Is there a reason to? My normal intuition (which is often wrong WRT front-end) is that `journalName` should be a required parameter, and the caller should handle the case where there are no journals for a given collection.

src/utils/transformStream-polyfill.js
```diff
@@ -0,0 +1,32 @@
+const _jstds_wm = new WeakMap(); /* info holder */
```
psFried, line 1: Is the reason for this obvious to front-end devs? It's definitely not clear to me

src/hooks/useJournalData.ts
```diff
@@ -0,0 +1,252 @@
...
+    TOKEN_NOT_FOUND = 'Unauthenticated',
+    TOKEN_INVALID = 'Authentication failed',
+    OPERATION_INVALID = 'Unauthorized',
+}
+
+const useJournalsForCollection = (collectionName: string) => {
+    const { session } = Auth.useUser();
+    const { grantDetails } = usePreFetchData();
+
+    const [gatewayConfig, setGatewayConfig] = useLocalStorage(
```
psFried, line 28: I'm not sure whether local storage is appropriate for this. The JWT that's used for accessing the data-plane may be different for each collection, so it seems potentially incorrect to use the same local storage key for all of them. Additionally, they may have relatively short expirations, and are fairly cheap to generate, so I don't see a reason to persist them indefinitely.

psFried (commented):
I left some comments here and there, but take them all with a grain of salt, since I'm kind of a fish out of water with React.

src/components/collection/DataPreview/TableView.tsx
jshearer: So yeah this was why I was asking about how to get the projections for a collection. "How hard would it be" to implement "GET projections"? Good next task for me?

src/components/collection/DataPreview/index.tsx
jshearer: I'd love to be able to read data from all partitions. That's part of the flowctl-in-Rust work you're doing, right @psFried?

src/hooks/useJournalData.ts
jshearer: Super not important actually, this is just a number of how many bytes to read back if the amount of data returned isn't enough to satisfy the desired number of rows

src/hooks/useJournalData.ts
jshearer: So this is a weird react hooks thing (which I'm curious about your opinion on @travjenkins). Because you can't conditionally call a hook, the question is what to do when (in this case) the journal info hasn't loaded yet. My solution was to allow passing null/undefined for `journalName`, and if it's not provided, then don't attempt to fetch the data, and just return an empty array.

src/utils/transformStream-polyfill.js
jshearer: Not sure how obvious it is, but the purpose is to polyfill the [TextDecoderStream](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoderStream) API in browsers that don't provide it natively. I ran into this issue because the JournalClient from `data-plane-gateway` uses `TextDecoderStream`, which wasn't available in Firefox I guess?

src/hooks/useJournalData.ts
jshearer: This was roughly lifted from: 

https://github.com/estuary/ui/blob/0f4e1645a6a11de756722c9f99c523cf0665d212/src/hooks/useShardsList.ts#L27-L30

Should that not use localstorage either? Or is the token generation different in the case of reading journal data vs shard info?

travjenkins: Oh this is news to me (I think) so I am going to setup a meeting with all of us to discuss.

src/hooks/useJournalData.ts
travjenkins: Yup... this is just kinda how it is done. Also, we tend to put all our effects near the bottom (as in right before the return) since a lot of times they need the variables and stuff declared above.

src/utils/transformStream-polyfill.js
travjenkins: I moved to a stand alone folder... but polyfills are fairly common in the front end world.

src/hooks/useJournalData.ts
travjenkins: comment was added.

src/components/collection/DataPreview/index.tsx
psFried: I'm doing essentially the same thing that you did here for `flowctl`, at least for the short term :grin: But yea, the intention is definitely to make that part of the Rust client.

src/hooks/useJournalData.ts
travjenkins: Had a quick meeting and decided that we will no longer store the token in LS and just keep it in memory. The UI will attempt to refresh the token at the proper interval. If someone reloads the page then we will fetch a fresh token.

PR #301, opened 2022-08-23
content: update verbiage around specs/catalogs

oliviamiannone:
## Changes

TLDR: this PR updates UI content that mentions the "catalog" to reflect recent redefinition of some concepts. 

Details: A recent [effort to simplify Flow concepts for entry-level users](https://github.com/estuary/flow/pull/620) led to a discussion around the ambiguity of the term "catalog" and what it really is. The group landed on the following updates to definitions:

* The **catalog** refers collectively to everything that exists on the Flow control plane. From a user's perspective, all the currently living Flow entities, regardless of own or what data flow they are part of. 

* **Flow specifications** are... the specifications themselves, which we previously have referred to as "Catalogs." They are stored in "Flow specification files." Shorthand of "specifications" is OK too. 

So we publish a specification _to_ the catalog, rather than publishing catalogs. (This also makes sense with the concept of _drafts_).

### DIFF
--- a/src/lang/en-US.ts
+++ b/src/lang/en-US.ts
@@ -366,7 +366,7 @@ const Collections: ResolvedIntlConfig['messages'] = {
 
 const entityCreateHeader = `Endpoint Config`;
 const EntityCreate: ResolvedIntlConfig['messages'] = {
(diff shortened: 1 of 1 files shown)


travjenkins (approved):
lgtm

PR #304, opened 2022-08-29
refactor: Reference new connector metadata fields

jshearer:
This PR changes the UI to reference the new connector metadata fields as described in https://github.com/estuary/animated-carnival/issues/58

There were a few rough edges, specifically around how to alias fields selecting from json documents. As it turns out, the answer is that these fields _need_ to have explicit typecasts, otherwise PostgREST will fail to parse the query.

Depends on https://github.com/estuary/animated-carnival/pull/63

@travjenkins curious about your thoughts on my changes to `ConnectorName.tsx`.  I was having some type issues, and it looked like that component may have been written to support passing in data in multiple different shapes, but looking at all of its usages I only found strings provided to `connector` (which I renamed `title`), so I got rid of the polymorphic stuff and it seems to work fine... ¯\_(ツ)_/¯?


src/components/ConnectorName.tsx
```diff
@@ -1,34 +1,21 @@
...
 }
 
-function ConnectorName({ connector, iconPath, iconSize }: Props) {
-    const connectorName = useConstant(() =>
-        typeof connector === 'string'
-            ? connector
-            : connector === null
-            ? ''
-            : connector.title
-    );
```
travjenkins, line 20: These changes are fine. They were here as the original way to allow the connector name to default to the image. If we only allow a string to be passed in that is fine.

travjenkins (approved):
LGTM

jshearer:
Getting close on https://github.com/estuary/animated-carnival/pull/63, then @travjenkins should I merge this?

PR #313, opened 2022-09-01
Log Rocket: adding logging and typing

travjenkins:
## Changes

1. We no longer mask fields that are not private. This is to try to increase the value of UX logging for LogRocket.
2. Added some types where `any` was being used.
3. Logging WebSocket stuff for LogRocket. This is not permanent but only until we fix the issue of subscriptions randomly hanging forever

## Tests

Manual tested

## Issues

None

## Content
 
None

## Screenshots

None

### DIFF
--- a/src/components/ConnectorToolbar.tsx
+++ b/src/components/ConnectorToolbar.tsx
@@ -182,6 +182,7 @@ function ConnectorToolbar({
         <Toolbar
             disableGutters
             sx={{ flexDirection: belowMd ? 'column' : 'row' }}
+            data-public
         >
             <Grid
                 container
--- a/src/services/logrocket.ts
+++ b/src/services/logrocket.ts
@@ -3,7 +3,7 @@ import filterObject from 'filter-obj';
 import { isEmpty } from 'lodash';
 import LogRocket from 'logrocket';
 import setupLogRocketReact from 'logrocket-react';
-import { getUserDetails } from 'services/supabase';
+import { DEFAULT_FILTER, getUserDetails } from 'services/supabase';
 import {
     getAppVersion,
     getLogRocketSettings,
@@ -193,6 +193,8 @@ export const identifyUser = (user: User) => {
     ) {
         const traits = {} as IUserTraits;
         const userDetails = getUserDetails(user);
+
+        traits.id = userDetails.id ?? DEFAULT_FILTER;
 
         if (logRocketSettings.idUser.includeName) {
             traits.name = userDetails.userName;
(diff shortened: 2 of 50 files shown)


src/components/ConnectorToolbar.tsx
```diff
@@ -182,6 +182,7 @@ function ConnectorToolbar({
         <Toolbar
             disableGutters
             sx={{ flexDirection: belowMd ? 'column' : 'row' }}
+            data-public
```
travjenkins, line 185: Allows fields to be shown in LogRocket play back

src/services/logrocket.ts
```diff
@@ -194,6 +194,8 @@ export const identifyUser = (user: User) => {
         const traits = {} as IUserTraits;
         const userDetails = getUserDetails(user);
 
+        traits.id = userDetails.id ?? DEFAULT_FILTER;
```
travjenkins, line 197: Adding the ID to the log rocket traits.

src/services/supabase.ts
```diff
@@ -65,7 +68,16 @@ export enum FUNCTIONS {
...
-    supabaseSettings.anonKey
+    supabaseSettings.anonKey,
+    {
+        // TODO (realtime) This is temporary until we figure out why some
+        //      subscriptions just hang forever.
+        realtime: {
+            logger: (kind: string, msg: string, data?: any) => {
+                LogRocket.log(kind, msg, data);
+            },
+        },
```
travjenkins, line 79: This will make sure there are logs in LogRocket's UI for us to review. Hoping to remove this somewhat soon.

src/services/supabase.ts
```diff
@@ -124,9 +136,12 @@ export const getUserDetails = (user: User | null) => {
             email = user.email;
             emailVerified = false;
         }
+
+        id = user.id;
```
travjenkins, line 140: Fetching the id out so we can use it in LogRocket

src/services/supabase.ts
```diff
@@ -195,25 +210,22 @@ export const endSubscription = (subscription: RealtimeSubscription) => {
...
             const response = payload.new
                 ? payload.new
-                : payload.record
-                ? payload.record
+                : // TODO (typing) during manual testing I have seen record be in the response
+                //      even though the type says it is not there. Needs more research
+                // eslint-disable-next-line @typescript-eslint/dot-notation
+                payload['record']
+                ? // eslint-disable-next-line @typescript-eslint/dot-notation
+                  payload['record']
```
travjenkins, line 228: Wanna look into this but it feels like the Supabase types aren't totally accurate.

Issue #316, opened 2022-09-06
Smooth the transition to the Edit page.


kiahna-tucker:
The edit workflow begins by presenting a set of forms and a specification editor that reflect the current configuration of the given task. Presently, asynchronous data is retrieved after the initial rendering of the _Edit_ page, resulting in a visually jarring secondary render.

PR #318, opened 2022-09-09
Create Selector Hooks

kiahna-tucker:
## Changes

1. Create and integrate selector hooks for all stores.

1. Update the entity type context provider so that it identifies when collections are the entity of focus. The use of editor selector hooks within the details panel led to this adjustment.

1. Migrate the entity type context provider to the _context_ directory.

1. Create an entity workflow context provider and type. Existing entity workflows are as follows: capture create, capture edit, materialization create, and materialization edit.

1. Enable self-hydration for the resource config store (i.e., _src/stores/ResourceConfig.ts_).

1. Create a context provider for the resource config store. This provider wraps the top-level, materialization create and edit components.

## Tests

Only manual testing of impacted components was performed.

## Issues

#316


src/stores/ResourceConfig.tsx
```diff
@@ -0,0 +1,655 @@
...
+                },
+                () => {
+                    set(
+                        produce((state: ResourceConfigState) => {
+                            const error: HydrationError = {
+                                message:
+                                    'A problem was encountered when hydrating the resource configuration store.',
+                                details:
+                                    'This error occurred in the top-level create function.',
+                            };
```
travjenkins, line 431: Doesn't look like these are displayed - so I think they should not look like they are formatted for a user as they do not really add much value.

src/hooks/updates/useEvaluateResourceConfigUpdates.ts
```diff
@@ -0,0 +1,60 @@
...
+import useGlobalSearchParams, {
+    GlobalSearchParams,
+} from 'hooks/searchParams/useGlobalSearchParams';
+import useDraftSpecs from 'hooks/useDraftSpecs';
+import { useLiveSpecsExtWithSpec } from 'hooks/useLiveSpecsExt';
+import { isEqual } from 'lodash';
+import { useMemo } from 'react';
+import { ResourceConfigDictionary } from 'stores/ResourceConfig';
+
+const evaluateResourceConfigEquality = (
```
travjenkins, line 11: This is fine for now - but eventually this file should be moved. I think it is pretty weird to have a subfolder under `hooks` called `updated`. That would read to me as being hooks for running updates.

I also think if we're going do a path (which I think we should) where sections have all these hooks, stores, etc. we should make the section/entity the parent folder.

src/components/materialization/Edit.tsx
```diff
@@ -56,57 +55,28 @@ function MaterializationEdit() {
...
+    const resetFormState = useFormStateStore_resetState();
 
-    const resetResourceConfigState = useZustandStore<
-        ResourceConfigState,
-        ResourceConfigState['resetState']
-    >(resourceConfigStoreName, (state) => state.resetState);
+    const exitWhenLogsClose = useFormStateStore_exitWhenLogsClose();
+
+    // Resource Config Store
+    // const resourceConfigChanged = useResourceConfig_stateChanged();
```
kiahna-tucker, line 76: I would like to address this in a separate PR because it is related to a larger code _quirk_ of the create and edit workflows. Given both forms involve a degree of prefilled data, the form data is always in a _changed_ state and thus the data loss modal is prompted on any exit.

src/components/shared/Entity/ValidationErrorSummary/materialization/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { hasLength } from 'utils/misc-utils';
+
+interface Props {
+    errorsExist: boolean;
+    ErrorComponent?: any | boolean;
+    hideIcon?: boolean;
+    headerMessageId?: string;
+}
+
+function ValidationErrorSummary({
```
travjenkins, line 25: We chatted a bit about this... but generally we should not duplicate to this level without first breaking shared stuff into a stand alone file first.... mainly since a good chunk of this is very easily shared.

travjenkins (approved):
don't need to wait on the call outs - this is fine to merge and we can handle stuff later.

src/api/hydration.ts
```diff
@@ -0,0 +1,60 @@
+import { ConnectorTag } from 'hooks/useConnectorTag';
```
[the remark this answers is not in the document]
kiahna-tucker: _Pending PR draft reference._

src/stores/ResourceConfig.tsx
kiahna-tucker: _Pending PR draft reference._

src/components/shared/Entity/ValidationErrorSummary/materialization/index.tsx
kiahna-tucker: _Pending PR draft reference._

Issue #320, opened 2022-09-12
Temporarily remove realtime calls


travjenkins:
There is an issue where using the RealTimeJS library is causing the UI to hang. Looking in LogRocket we can see the WebSocket activity but it never makes it to the subscription.

We'll remove this for awhile and replace with a basic polling solution.

PR #322, opened 2022-09-12
Switching back to polling

travjenkins:
## Changes

Stop using the real time library and use polling
Handle when logs have network issue and message to user

## Tests

manual

## Issues

fixes: https://github.com/estuary/ui/issues/320

## Content

none

## Screenshots

New message to user when Logs cannot be reached
image


src/components/Logs.tsx
```diff
@@ -87,25 +90,33 @@ function Logs({
...
-                    return generateLogLine(logData);
-                });
-                setOffset(offset + viewLogsResponse.length);
-                setLogs(logs.concat(logsReduced));
-                if (disableIntervalFetching) {
+            const fetchLogsResponse = await fetchLogs();
+
+            if (fetchLogsResponse.error) {
+                setNetworkFailure(true);
+                setEmptyResponses(MAX_EMPTY_CALLS + 1);
```
travjenkins, line 104: If there was an error it is probably due to a networking issue. So we should just stop trying to fetch logs.

src/components/Logs.tsx
```diff
@@ -130,7 +141,13 @@ function Logs({
...
                     }}
                 >
-                    <FormattedMessage id="logs.toManyEmpty" />
+                    {networkFailure ? (
+                        <Alert severity="warning">
+                            <FormattedMessage id="logs.networkFailure" />
+                        </Alert>
+                    ) : (
+                        <FormattedMessage id="logs.tooManyEmpty" />
+                    )}
```
travjenkins, line 150: Message to user

src/components/capture/Create.tsx
```diff
@@ -42,11 +46,11 @@ const formStateStoreName = FormStateStoreNames.CAPTURE_CREATE;
...
-        name: payload.capture_name,
-        id: payload.id,
-        draft_id: payload.draft_id,
-        logs_token: payload.logs_token,
-        status: payload.job_status.type,
+        name: payload.capture_name ?? DEFAULT_FILTER,
+        id: payload.id ?? DEFAULT_FILTER,
+        draft_id: payload.draft_id ?? DEFAULT_FILTER,
+        logs_token: payload.logs_token ?? DEFAULT_FILTER,
+        status: payload.job_status?.type ?? DEFAULT_FILTER,
```
travjenkins, line 53: Temporary. We not longer get an entire payload back with polling... and honestly this is less important as long as polling works.

src/components/capture/Create.tsx
```diff
@@ -120,24 +124,12 @@ function CaptureCreate() {
...
-            } else {
-                setFailureState();
-            }
+        callFailed: (formState: any) => {
+            setFormState({
+                status: FormStatus.FAILED,
+                exitWhenLogsClose: false,
+                ...formState,
+            });
         },
```
travjenkins, line 133: Not longer need to clean up a subscription.

src/components/capture/Create.tsx
```diff
@@ -182,10 +174,18 @@ function CaptureCreate() {
...
+                .from(TABLES.DISCOVERS)
+                .select(
+                    `
+                    draft_id,
+                    job_status
+                `
+                )
+                .match({
+                    draft_id: discoverDraftId,
+                }),
```
travjenkins, line 188: Call the new poller and pass it the query you want to run and make sure eventually gets a job status of "success"

src/components/capture/GenerateButton.tsx
```diff
@@ -120,26 +120,21 @@ function CaptureGenerateButton({
...
+                return callFailed({
+                    error: {
+                        title: 'captureCreate.generate.failedErrorTitle',
+                        error: discoverResponse.error,
                     },
-                    discoversSubscription
-                );
+                });
             }
+            subscription(draftsResponse.data[0].id);
```
travjenkins, line 137: Moving the subscription call to after the discover call is checked for errors.

Most of the "changes" here are indentation.

src/components/materialization/Create.tsx
```diff
@@ -121,25 +118,12 @@ function MaterializationCreate() {
...
-                    .catch(() => {});
-            } else {
-                setFailureState();
-            }
+        callFailed: (formState: any) => {
+            setFormState({
+                status: FormStatus.FAILED,
+                exitWhenLogsClose: false,
+                ...formState,
+            });
```
travjenkins, line 126: same as above. No need to clean up subs.

src/components/shared/Entity/Actions/Save.tsx
```diff
@@ -30,11 +30,11 @@ interface Props {
...
-        id: payload.id,
-        draft_id: payload.draft_id,
-        dry_run: payload.dry_run,
-        logs_token: payload.logs_token,
-        status: payload.job_status.type,
+        id: payload.id ?? DEFAULT_FILTER,
+        draft_id: payload.draft_id ?? DEFAULT_FILTER,
+        dry_run: payload.dry_run ?? DEFAULT_FILTER,
+        logs_token: payload.logs_token ?? DEFAULT_FILTER,
+        status: payload.job_status?.type ?? DEFAULT_FILTER,
```
travjenkins, line 37: same as above. temp fix.

src/components/shared/Entity/Actions/Save.tsx
```diff
@@ -96,95 +96,105 @@ function EntityCreateSave({
...
+                    `
+                    job_status,
+                    logs_token,
+                    id
+                `
+                )
+                .match({
+                    draft_id: draftIdVal,
+                    logs_token: logTokenVal,
+                }),
```
travjenkins, line 118: same as above... calling new poller

src/components/shared/Entity/Actions/Save.tsx
```diff
@@ -96,95 +96,105 @@ function EntityCreateSave({
...
             async (payload: any) => {
-                if (payload.logs_token === logTokenVal) {
```
travjenkins, line 106: We now include the logs token in the query itself so we don't need this check.

src/components/shared/Entity/Actions/Save.tsx
```diff
@@ -96,95 +96,105 @@ function EntityCreateSave({
...
-            dryRun ?? false,
-            entityDescription
-        );
-        const publicationsSubscription = waitForPublishToFinish(
-            response.data[0].logs_token
-        );
-        if (response.error) {
-            onFailure(
-                {
+        if (draftId) {
```
travjenkins, line 169: Random wonder if this was ever blank and caused issues. So adding a check for a little while. If we don't see any logging in LogRocket then we can remove this soon.

src/services/supabase.ts
```diff
@@ -278,6 +227,134 @@ export const jobSucceeded = (jobStatus?: JobStatus) => {
     }
 };
 
+// START: Poller
```
travjenkins, line 230: Eventually we'll make a "parent" poller as we need polling in other places.

src/services/supabase.ts
```diff
@@ -278,6 +227,134 @@ export const jobSucceeded = (jobStatus?: JobStatus) => {
...
+                    } else {
+                        interval =
+                            interval < INTERVAL_MAX
+                                ? interval + INTERVAL_INCREMENT
+                                : INTERVAL_MAX;
+                        pollerTimeout = window.setTimeout(
+                            makeApiCall,
+                            interval
+                        );
+                    }
```
travjenkins, line 287: Most of this is identical to the subscription handler.

However we not need a block to handle calculating the interval.

src/lang/en-US.ts
```diff
@@ -309,7 +309,8 @@ const EntityTable: ResolvedIntlConfig['messages'] = {
 
 const LogsDialog: ResolvedIntlConfig['messages'] = {
     'logs.default': `Waiting for logs...`,
-    'logs.toManyEmpty': `Logs for this build may have ended. Click ${CTAs['cta.restart']} to check for new logs.`,
+    'logs.tooManyEmpty': `Logs for this build may have ended. Click ${CTAs['cta.restart']} to check for new logs.`,
+    'logs.networkFailure': `We are having issues fetching logs. There may be network issues. Click ${CTAs['cta.restart']} to try again.`,
```
kiahna-tucker, line 313: >We are having issues fetching logs.

This may cause a greater amount of panic than intended. If the source of the problem falls within the user's locus of control (e.g., the stability of their network connection, clicking a button that makes a duplicate API call, refreshing the window, etc.), I would recommend writing this _error acknowledgement_ in the past tense. If the source of the problem is outside the user's locus of control, then writing the _error acknowledgement_ in the present tense communicates a proper level of severity.

>There may be network issues.

It is unclear whether the user should take action to investigate the source of the error. If a weak client connection is likely at fault, I recommend explicitly advising the user to evaluate their network connection before attempting to fetch the logs again.

```ts
`We encountered a problem retrieving logs. Please check your network connection and click ${CTAs['cta.restart']} to try again.`
```

src/lang/en-US.ts
```diff
@@ -430,6 +431,7 @@ const EntityCreate: ResolvedIntlConfig['messages'] = {
     'entityCreate.bindingsConfig.noRowsTitle': `No selection made`,
 
     'entityCreate.connector.label': `${CommonMessages['connector.label']} Search`,
+    'entityCreate.errors.missingDraftId': `Missing draft id`,
```
kiahna-tucker, line 434: I believe `id` should be capitalized.

kiahna-tucker (approved):

src/lang/en-US.ts
travjenkins: We cannot know the cause as not being able to reach a server could be their network connection or the server network connection.

I like that... I'll update

PR #346, opened 2022-09-29
Update issue templates

travjenkins:


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,10 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Please propose features in the Discussions area as an `idea`**
```
mdibaiee, line 10: Can we instead just redirect them to the discussions page?

mdibaiee (approved):

.github/ISSUE_TEMPLATE/feature_request.md
travjenkins: I was hoping we could do something like that but I do not know much about configuring Github in that way.

Issue #294, opened 2022-08-23
Pass FE Config Values to BE


dyaffe:
**Problem**
Some OAuth providers (ex. MSFT) require user-input information.  They for example have a "Tenant ID" which is usually set to "common" but sometimes is over ridden and passed into OAuth.

**Proposal**
The frontend should send config to the backend.  Ideally I could use a variable like `{{ config.tenant_id }}` to populate tenant ID in an OAuth block.

travjenkins:
The frontend is already sending back all the information that is provided by the provider. We do no filtering at all.

dyaffe:
To clarify, this isn't provided by the provider, but rather user-input.  It's something that is in the form that users fill out and needs to be passed into OAuth config.

travjenkins:
So what endpoint does this need to be added to? We have three endpoints.

1. To generate the URL to send the user to
2. Get the access tokens
3. Encrypt the configuration

dyaffe:
Actually adding @mdibaiee 

I might be wrong here -- it looks like [the connector](https://docs.microsoft.com/en-us/advertising/guides/authentication-oauth-get-tokens?view=bingads-13#request-accesstoken) is actually the thing using that value and it isn't necessary to pass to the BE.

mdibaiee:
Not sure if this is necessary in case of Microsoft, but in general we want to allow OAuth flows to use properties of the config other than OAuth-specific properties.

Both `auth-url` and `access-token` functions accept `config` as one of their parameters, which is expected to be the whole JSON config of the connector (all that user has input for this connector), see: https://github.com/estuary/animated-carnival/blob/main/supabase/functions/oauth/auth-url.ts#L20 and https://github.com/estuary/animated-carnival/blob/main/supabase/functions/oauth/access-token.ts#L10

travjenkins:
For `auth-url` we just send the `connectorID` because that was the initial plan... that connectors could generate the correct URL for us. We send nothing extra with that.

For `access-token` we just send back the `state` and the `code` that is being returned from the provider.

mdibaiee:
@travjenkins okay, but can we also send the whole config of the connector? this is necessary to support some advanced use cases of OAuth

travjenkins:
I'll add this... but there is no guarantee that the config is filled out when we pass it. That all depends on how the user fills stuff out. So we'll need some good error messages getting returned from these endpoints if stuff is missing.

dyaffe:
[...]

PR #355, opened 2022-10-17
Adding config to oauth calls

travjenkins:
## Changes

1. add endpoint config to oauth related calls so that connectors that need user provided details can use it

## Tests

manual

## Issues

fixes: https://github.com/estuary/ui/issues/294

## Content

none

## Screenshots

none


mdibaiee (approved):
Thanks a lot Travis! LGTM as is

PR #363, opened 2022-10-26
Enable binding selection in capture workflows.

kiahna-tucker:
## Changes

## Tests

## Issues

## Content

## Screenshots

### DIFF
--- a/src/components/capture/Create.tsx
+++ b/src/components/capture/Create.tsx
@@ -37,6 +37,10 @@ import {
     useFormStateStore_resetState,
     useFormStateStore_setFormState,
 } from 'stores/FormState';
+import {
+    ResourceConfigHydrator,
+    useResourceConfig_resetState,
+} from 'stores/ResourceConfig';
 import { ENTITY } from 'types';
 import { getPathWithParams } from 'utils/misc-utils';
 
@@ -81,6 +85,9 @@ function CaptureCreate() {
     const resetFormState = useFormStateStore_resetState();
     const exitWhenLogsClose = useFormStateStore_exitWhenLogsClose();
 
+    // Resource Config Store
+    const resetResourceConfigState = useResourceConfig_resetState();
+
     // Reset the catalog if the connector changes
     useEffect(() => {
         setDraftId(null);
@@ -89,6 +96,7 @@ function CaptureCreate() {
     const resetState = () => {
         resetDetailsForm();
         resetEndpointConfigState();
+        resetResourceConfigState();
         resetFormState();
     };
 
@@ -157,9 +165,11 @@ function CaptureCreate() {
                 }),
             (payload: any) => {
                 setDraftId(payload.draft_id);
+
                 setFormState({
                     status: FormStatus.GENERATED,
                 });
+
                 trackEvent(payload);
             },
             (payload: any) => {
@@ -180,44 +190,49 @@ function CaptureCreate() {
                     'https://docs.estuary.dev/guides/create-dataflow/#create-a-capture',
             }}
         >
-            <EntityCreate
-                title="browserTitle.captureCreate"
-                connectorType={entityType}
-                resetState={resetState}
-                errorSummary={
-                    <ValidationErrorSummary
-                        errorsExist={detailsFormErrorsExist}
-                    />
-                }
-                toolbar={
-                    <EntityToolbar
-                        GenerateButton={
-                            <CaptureGenerateButton
-                                disabled={!hasConnectors}
-                                callFailed={helpers.callFailed}
-                                subscription={discoversSubscription}
-                            />
-                        }
-                        TestButton={
-                            <EntityTestButton
-                                closeLogs={handlers.closeLogs}
-                                callFailed={helpers.callFailed}
-                                disabled={!hasConnectors}
-                                logEvent={CustomEvents.CAPTURE_TEST}
-                            />
-                        }
-                        SaveButton={
-                            <EntitySaveButton
-                                closeLogs={handlers.closeLogs}
-                                callFailed={helpers.callFailed}
-                                disabled={!draftId}
-                                materialize={handlers.materializeCollections}
-                                logEvent={CustomEvents.CAPTURE_CREATE}
-                            />
-                        }
-                    />
-                }
-            />
+            <ResourceConfigHydrator>
+                <EntityCreate
+                    title="browserTitle.captureCreate"
+                    connectorType={entityType}
+                    showCollections
+                    resetState={resetState}
+                    errorSummary={
+                        <ValidationErrorSummary
+                            errorsExist={detailsFormErrorsExist}
+                        />
+                    }
+                    toolbar={
+                        <EntityToolbar
+                            GenerateButton={
+                                <CaptureGenerateButton
+                                    disabled={!hasConnectors}
+                                    callFailed={helpers.callFailed}
+                                    subscription={discoversSubscription}
+                                />
+                            }
+                            TestButton={
+                                <EntityTestButton
+                                    closeLogs={handlers.closeLogs}
+                                    callFailed={helpers.callFailed}
+                                    disabled={!hasConnectors}
+                                    logEvent={CustomEvents.CAPTURE_TEST}
+                                />
+                            }
+                            SaveButton={
+                                <EntitySaveButton
+                                    closeLogs={handlers.closeLogs}
+                                    callFailed={helpers.callFailed}
+                                    disabled={!draftId}
+                                    materialize={
+                                        handlers.materializeCollections
+                                    }
+                                    logEvent={CustomEvents.CAPTURE_CREATE}
+                                />
+                            }
+                        />
+                    }
+                />
+            </ResourceConfigHydrator>
         </PageContainer>
     );
 }
--- a/src/components/capture/Create.tsx
+++ b/src/components/capture/Create.tsx
@@ -2,8 +2,11 @@ import { generateCaptureDraftSpec, updateDraftSpec } from 'api/draftSpecs';
 import { authenticatedRoutes } from 'app/Authenticated';
 import CaptureGenerateButton from 'components/capture/GenerateButton';
 import {
+    useEditorStore_editDraftId,
     useEditorStore_id,
     useEditorStore_pubId,
+    useEditorStore_resetState,
+    useEditorStore_setEditDraftId,
     useEditorStore_setId,
 } from 'components/editor/Store';
 import EntitySaveButton from 'components/shared/Entity/Actions/SaveButton';
@@ -15,7 +18,7 @@ import PageContainer from 'components/shared/PageContainer';
 import { GlobalSearchParams } from 'hooks/searchParams/useGlobalSearchParams';
 import { useClient } from 'hooks/supabase-swr';
 import useConnectorWithTagDetail from 'hooks/useConnectorWithTagDetail';
-import { DraftSpecQuery } from 'hooks/useDraftSpecs';
+import useDraftSpecs, { DraftSpecQuery } from 'hooks/useDraftSpecs';
 import LogRocket from 'logrocket';
 import { useEffect } from 'react';
 import { useNavigate } from 'react-router-dom';
@@ -82,7 +85,12 @@ function CaptureCreate() {
     const draftId = useEditorStore_id();
     const setDraftId = useEditorStore_setId();
 
+    const editDraftId = useEditorStore_editDraftId();
+    const setEditDraftId = useEditorStore_setEditDraftId();
+
     const pubId = useEditorStore_pubId();
+
+    const resetEditorStore = useEditorStore_resetState();
 
     // Endpoint Config Store
     const resetEndpointConfigState = useEndpointConfigStore_reset();
@@ -101,6 +109,9 @@ function CaptureCreate() {
 
     const resetResourceConfigState = useResourceConfig_resetState();
 
+    const { mutate: mutateDraftSpecs, ...draftSpecsMetadata } =
+        useDraftSpecs(editDraftId);
+
     // Reset the catalog if the connector changes
     useEffect(() => {
         setDraftId(null);
@@ -111,6 +122,7 @@ function CaptureCreate() {
         resetEndpointConfigState();
         resetResourceConfigState();
         resetFormState();
+        resetEditorStore();
     };
 
     const helpers = {
@@ -247,6 +259,9 @@ function CaptureCreate() {
                 setCurrentCollection(updatedBindings[0].target);
             }
         }
+
+        setDraftId(newDraftId);
+        setEditDraftId(newDraftId);
     };
 
     // TODO (optimization): Create a shared discovers table subscription.
@@ -256,6 +271,7 @@ function CaptureCreate() {
         resourceConfig: ResourceConfigDictionary
     ) => {
         setDraftId(null);
+
         jobStatusPoller(
             supabaseClient
                 .from(TABLES.DISCOVERS)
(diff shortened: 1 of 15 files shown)


src/components/shared/Entity/WrapperWithHeader/index.tsx
```diff
@@ -53,7 +53,7 @@ function WrapperWithHeader({
             expanded={expanded}
             onChange={handlers.change}
             sx={{
-                mt: 2,
+                my: 2,
```
kiahna-tucker, line 56: When a section of holistic create/edit form is unable to be rendered and an error alert takes its place, this margin ensures adequate space exists between the alert and the form section above it.

src/context/Theme.tsx
```diff
@@ -197,6 +205,21 @@ export const jsonFormsGroupHeaders = {
     dark: 'transparent',
 };
 
+export const reflexSplitterBackground = {
+    light: slate[25],
+    dark: slate[200],
+};
+
+export const alternativeReflexContainerBackground = {
```
kiahna-tucker, line 213: Currently, the _alernative_ styles are only applied to the `ListAndDetails` component.

src/components/editor/Bindings/index.tsx
```diff
@@ -60,35 +73,59 @@ function BindingsMultiEditor({ draftSpecs = [], readOnly = false }: Props) {
...
-    }, [setServerUpdateRequired, resourceConfigUpdated, workflow]);
+        setServerUpdateRequired(resourceConfigUpdated);
+    }, [setServerUpdateRequired, resourceConfigUpdated]);
+
+    const { liveSpecs } = useLiveSpecs('collection');
+
+    const fetchingSpecs =
+        entityType === ENTITY.MATERIALIZATION
+            ? liveSpecs.length === 0
+            : draftSpecs.length === 0;
```
kiahna-tucker, line 103: PENDING: This condition is inaccurate for captures. Whether the draft specs _and_ live specs data is undergoing validation is the determinizing factor. Additionally, draft spec and live spec data should be passed to the `CollectionPicker` component via props to prevent duplicate API calls.

src/components/collection/Picker.tsx
```diff
@@ -1,85 +1,176 @@
...
+        updateCollections: (
+            _event: React.SyntheticEvent,
+            value: CollectionData[]
+        ) => {
+            setResourceConfig(value.map(({ name }) => name));
+
+            value
+                .filter(({ name }) => discoveredCollections?.includes(name))
+                .forEach(({ name }) => {
+                    setRestrictedDiscoveredCollections(name);
```
kiahna-tucker, line 119: Remove a given, discovered collection from the list of restricted discovered collections when reselected.

src/components/shared/Entity/WrapperWithHeader/index.tsx
travjenkins: Is there a reason we don't add the space to the error itself?

src/components/collection/Picker.tsx
```diff
@@ -1,85 +1,189 @@
...
+
     const handlers = {
-        updateCollections: (event: React.SyntheticEvent, value: any) => {
-            setResourceConfig(value);
+        updateCollections: (
+            event: React.SyntheticEvent,
+            value: CollectionData[],
+            reason: AutocompleteChangeReason
+        ) => {
+            const removeOptionWithBackspace =
```
kiahna-tucker, line 120: This boolean detects whether the user is attempting to remove one of the selected autocomplete options via the backspace key. While this feels a little hacky, it is MUI's recommended solution to this problem.

The same can be said of the logic used to prevent tags from rendering in an autocomplete component with the `multiple` prop set. MUI advises one to supply a function that returns null to the `renderTags` prop to achieve this.

src/lang/en-US.ts
```diff
@@ -479,10 +479,15 @@ const CaptureCreate: ResolvedIntlConfig['messages'] = {
     'captureCreate.editor.default': `Before you can edit the capture specification, you must fill out the Connection Configuration section and click "${CTAs['cta.generateCatalog.capture']}." `,
     'captureCreate.finalReview.instructions': `The following Flow specification was generated from the details you provided. To make changes, you can enter new values in the form above and click "${CTAs['cta.generateCatalog.capture']}" again. You can also edit the specification file directly. Click "${CTAs['cta.saveEntity']}" to proceed.`,
 
+    'captureCreate.collections.heading': `Output Collections`,
+    'captureCreate.collectionSelector.heading': `Collection Selector`,
+    'captureCreate.collectionSelector.instructions': `The collections bound to your existing capture. To make changes, you can enter new values in the this section of the form or edit the YAML file shown in the ${EntityCreate['entityCreate.catalogEditor.heading']} section below.`,
```
kiahna-tucker, line 484: The use of the word _existing_ makes this messaging flawed for the capture create workflow. At the very least, _existing_ should be dropped. Would that change suffice?

For reference, the materialization analog of this message is: _Choose one or more collections to materialize._

src/components/editor/ListAndDetails.tsx
```diff
@@ -34,25 +52,43 @@ function ListAndDetails({ backgroundColor, list, details, height }: Props) {
                     size={INITIAL_SELECTOR_WIDTH}
                     minSize={MIN_RESIZE_WIDTH}
                 >
-                    <div className="pane-content" style={{ height: heightVal }}>
-                        {list}
+                    <div
+                        className="pane-content"
+                        style={{
+                            height: heightVal,
+                            border: displayBorder ? slateOutline[200] : '',
```
travjenkins, line 59: I don't think the default/fallback should be an empty string. It should be `undefined` as that is equivalent to not passing the `border` property

src/context/Theme.tsx
```diff
@@ -156,6 +159,11 @@ export const tableAlternateRowsSx: SxProps<Theme> = {
     },
 };
 
+// TODO (optimization): Replace instances of outlineSx with slateOutline.
+export const slateOutline = {
+    200: `1px solid ${slate[200]}`,
+};
+
```
travjenkins, line 166: This makes no sense to me. We are duplicating the `200` here for honestly very little code de-duplication of the `1px solid`. 

If we want to share these kind of properties this should be a function that consumes arguments to generate this so it works for any predefined color and shade.

src/components/editor/Bindings/Selector.tsx
```diff
@@ -1,21 +1,62 @@
...
+
+function DeleteButton({ collection, disabled }: DeleteButtonProps) {
+    const removeCollection = useResourceConfig_removeCollection();
+
+    const setRestrictedDiscoveredCollections =
+        useResourceConfig_setRestrictedDiscoveredCollections();
+
+    const handlers = {
+        removeCollection: (event: React.MouseEvent<HTMLElement>) => {
+            event.preventDefault();
```
travjenkins, line 42: So this is kind of an issue... the prevent default isn't stopping anything. The row click action still fires. This means that delete is working via a weird side effect.

If you add `event.stopPropagation` then it stops the rowClick and delete no longer works correctly. I do not think this is a blocking issue for the PR but this will need to get fixed later.

src/stores/ResourceConfig.tsx
```diff
@@ -194,16 +237,134 @@ const getInitialState = (
...
+                    state.restrictedDiscoveredCollections =
+                        restrictedDiscoveredCollections.filter(
+                            (collection) => collection !== value
+                        );
+                } else if (discoveredCollections?.includes(value)) {
+                    state.restrictedDiscoveredCollections = [
+                        value,
+                        ...restrictedDiscoveredCollections,
+                    ];
+                }
```
travjenkins, line 361: I think this might be a spot where it makes sense to do the setting of `state.restrictedDiscoveredCollections` on a single line and storing what we're gonna put there in a variable.

Unless you think eventually there will be more logic difference in the two `if` blocks.

src/stores/ResourceConfig.tsx
```diff
@@ -364,14 +523,21 @@ const getInitialState = (
             if (data && data.length > 0) {
                 const { setResourceConfig, preFillCollections } = get();
 
-                data[0].spec.bindings.forEach((binding: any) =>
-                    setResourceConfig(binding.source, {
+                const collectionNameProp =
+                    entityType === ENTITY.MATERIALIZATION ? 'source' : 'target';
```
travjenkins, line 527: We should potentially move this to a stand alone function... could wait until we need it used elsewhere.

But would be good to try to start keeping as much of the difference between entities in a single location if possible.

travjenkins (changes requested):

src/components/capture/Create.tsx
```diff
@@ -156,10 +258,17 @@ function CaptureCreate() {
                     draft_id: discoverDraftId,
                 }),
             (payload: any) => {
-                setDraftId(payload.draft_id);
+                void storeDiscoveredCollections(
+                    payload.draft_id,
+                    resourceConfig
+                );
```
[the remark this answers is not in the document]
kiahna-tucker: All errors are caught internally.

src/components/capture/Create.tsx
```diff
@@ -141,8 +180,71 @@ function CaptureCreate() {
...
+                draftSpecsResponse,
+                resourceConfig,
+                existingCollections,
+                restrictedDiscoveredCollections
+            );
+            if (updatedDraftSpecsResponse.error) {
+                return helpers.callFailed({
+                    error: {
+                        title: 'captureCreate.generate.failedErrorTitle',
+                        error: draftSpecsResponse.error,
```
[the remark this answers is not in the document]
kiahna-tucker: This is clearly a typo -- copy and paste error. Analogous capture edit logic is aptly written.

src/components/capture/Edit.tsx
```diff
@@ -208,39 +231,58 @@ function CaptureEdit() {
...
-                {
-                    data: draftSpecsResponse.data[0].spec.endpoint.connector
-                        .config,
-                },
-                'capture_edit'
-            );
+            if (
+                updatedDraftSpecsResponse.data &&
+                updatedDraftSpecsResponse.data.length > 0
+            ) {
```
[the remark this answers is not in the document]
kiahna-tucker: Agreed. I have been adapting this as I consolidated create and edit for each task.

src/components/collection/Picker.tsx
```diff
@@ -1,85 +1,189 @@
...
+
+            if (entityType === ENTITY.CAPTURE) {
+                collectionsOnServer = [
+                    ...draftSpecs
+                        .filter(
+                            ({ spec_type }) => spec_type === ENTITY.COLLECTION
+                        )
+                        .map(({ catalog_name }) => ({
+                            name: catalog_name,
+                            classification: 'Discovered Collections',
```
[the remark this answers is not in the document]
kiahna-tucker: I disagree. If you would like me to remove _Collections_ from these strings and create an interpolated string in the Autocomplete `groupBy` function, fine. _Translation_, to me, is confusing as no aspect of the collection name is being transformed in any way. It is, however, being classified or categorized.

src/components/editor/Bindings/Selector.tsx
```diff
@@ -24,9 +65,20 @@ const initialState = {
     },
 };
 
-function BindingSelector() {
+const typographyTruncation: TypographyProps = {
+    noWrap: true,
+    sx: {
+        minWidth: 0,
+    },
+};
```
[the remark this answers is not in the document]
kiahna-tucker: This is how words are being truncated in this instance, yes. If you would like me to move it to the theme, that is fine. Whether this will be useful elsewhere is impossible to determine at this point.

src/components/editor/Bindings/index.tsx
```diff
@@ -60,35 +73,59 @@ function BindingsMultiEditor({ draftSpecs = [], readOnly = false }: Props) {
...
 
     return (
         <>
-            <CollectionSelector readOnly={readOnly} />
+            <Typography variant="h5" sx={{ mb: 1 }}>
+                <FormattedMessage
+                    id={`${messagePrefix}.collectionSelector.heading`}
+                />
+            </Typography>
```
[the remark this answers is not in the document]
kiahna-tucker: It is not an `<h5>` element. It is a span.

src/stores/ResourceConfig.tsx
```diff
@@ -173,13 +211,18 @@ const getInitialState = (
         );
     },
 
-    preFillCollections: (value) => {
+    preFillCollections: (value, entityType) => {
```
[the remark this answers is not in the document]
kiahna-tucker: Yup.

src/stores/ResourceConfig.tsx
kiahna-tucker: It is more suitable to address this in the PR to combine the create and edit workflows. Not a necessary fix now.

src/context/Theme.tsx
kiahna-tucker: Will remove

src/components/editor/Bindings/Selector.tsx
```diff
@@ -69,44 +141,60 @@ function BindingSelector() {
...
-                },
-                '& .MuiDataGrid-columnSeparator': {
-                    display: 'none',
-                },
-            }}
-        />
+        <>
+            <CollectionPicker readOnly={readOnly} />
+
+            <Box sx={{ height: 280 }}>
```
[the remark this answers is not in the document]
kiahna-tucker: Prevents secondary default scroll bar.

src/components/editor/ListAndDetails.tsx
kiahna-tucker: I try to avoid setting value/attribute explicitly to `undefined`. If that is an acceptable practice in your opinion, it's your codebase. Will correct.

src/components/collection/CollectionSkeletons.tsx
```diff
@@ -0,0 +1,48 @@
...
+                <Skeleton variant="rectangular" />
+            </Box>
+
+            <>
+                <Skeleton variant="rectangular" sx={{ mb: 1 }} />
+
+                <Skeleton variant="rectangular" height={40} sx={{ mb: 1 }} />
+
+                <Skeleton variant="rectangular" />
+            </>
```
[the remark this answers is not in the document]
kiahna-tucker: To avoid a needless div.

src/utils/workflow-utils.ts
```diff
@@ -0,0 +1,92 @@
...
+export const modifyDiscoveredDraftSpec = async (
+    response: {
+        data: DraftSpecQuery[];
+        error?: undefined;
+    },
+    resourceConfig: ResourceConfigDictionary,
+    existingCollections: string[],
+    restrictedDiscoveredCollections: string[],
+    lastPubId?: string
+): Promise<CallSupabaseResponse<any>> => {
```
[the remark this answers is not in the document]
kiahna-tucker: If you would prefer existing collections to be recalculated, fine.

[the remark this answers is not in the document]
travjenkins: Naw... this was nonblocking but I didn't mark it was such. Mainly just want to include the thought. Sorry.

src/components/capture/Create.tsx
```diff
@@ -141,8 +180,71 @@ function CaptureCreate() {
...
+        if (draftSpecsResponse.error) {
+            return helpers.callFailed({
+                error: {
+                    title: 'captureCreate.generate.failedErrorTitle',
+                    error: draftSpecsResponse.error,
+                },
+            });
+        }
+
+        if (draftSpecsResponse.data && draftSpecsResponse.data.length > 0) {
```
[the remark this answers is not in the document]
kiahna-tucker: That scenario _should_ never be reached, but I would rather handle an edge case with a very simple solution than leaving the code open to error. Handling empty response data in a dedicated conditional block is the better approach across the board and is already in flight on a separate branch.

src/utils/workflow-utils.ts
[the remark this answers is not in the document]
kiahna-tucker: Updated.

src/stores/ResourceConfig.tsx
kiahna-tucker: If that is what you feel is best.

travjenkins (approved):
lgtm

PR #366, opened 2022-11-03
Bindings selection patches for capture workflows.

kiahna-tucker:
## Changes

1. Restrict the list of available collections in the capture create workflow to discovered collections.

1. Remove discovered collections that have been explicitly removed by the user from the `draft_specs` table in the success handler of the discover subscription.

1. Remove the empty endpoint config form data check for the capture edit workflow.

## Tests

Only manual testing was performed. Currently, I am unable to _install_ any connectors outside of the defaults in my local environment. I created a couple mock scenarios to evaluate the third _feature_ in the list in the section above. It went smoothly, but I would prefer for this connector to be tested explicitly if possible.


travjenkins (approved):
lgtm

PR #367, opened 2022-11-04
prod: Prevent duplication in the available collections list

kiahna-tucker:
## Changes

Updated the logic which calculates the options for the list of available collections so that the array of existing collections is explicitly pruned, removing any existing collection that has a discovered counterpart.

## Tests

Only manual testing was performed. I ran through all of the capture and materialization workflows, evaluating the list of available collections after taking actions on the form. For captures, taking special note to monitor the list of available collections after subsequent discoveries.


travjenkins (approved):
lgtm

Issue #372, opened 2022-11-09
Some postgrest calls fail due to URI size limit


travjenkins:
**Bug**
If a user deletes a significant amount of collections from a capture there can be a 414 server error. This causes the UI to enter into a failure state that they cannot really get out of on their own.

**Expected**
Users can remove as many collections from a capture as they want without issue.

**Screenshots**
![image](https://user-images.githubusercontent.com/270078/200704409-2cc7ea87-bbea-4284-899b-bed86cbe0543.png)

**Desktop (please complete the following information):**
N/A

**Additional context**
Slack Conversation:
We talked on the stand up line, and were thinking that the best course of action might be to put in a "temporary" fix just for the delete calls that chunks them into multiple requests if necessary. Joseph was going to start on that and see if he could get something in quickly, so that it can maybe be ready for the call with the customer tomorrow.

Short term : fix good enough for delete
Long term : ensure any postgrest call that's too big does not fail

PR #373, opened 2022-11-09
Chunk DELETE requests, and add `Add All`/`Remove All` buttons for collection selection

jshearer:
## Changes
Per our convo earlier today, this adds delete chunking, currently set to a chunk size of 10 and concurrency limit of 3 to avoid overloading supabase or running afowl of ratelimits.

It also adds `Add All`/`Delete All` buttons. @travjenkins or @kiahna-tucker, these took me a bunch of fiddling to get right, and I'm not entirely sure I did. They appear to work, but I might have missed some edge cases?

## Screenshots

See `DELETE` chunking:
Screen Shot 2022-11-08 at 21 00 03

`Add All`/`Remove All` buttons:
Screen Shot 2022-11-08 at 21 01 11

**Note** that `Add All` adds collections without their fields being autofilled. This is less than optimal, but is the same behavior that we have today when adding a collection from the dropdown, so I figured it was good enough for this PR.

Fixes https://github.com/estuary/ui/issues/372

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,7 @@
                 "madge": "^5.0.1",
                 "monaco-editor": "^0.33.0",
                 "notistack": "^2.0.5",
+                "p-limit": "^4.0.0",
                 "path-list-to-tree": "^1.1.1",
                 "react": "^17.0.2",
                 "react-dom": "^17.0.2",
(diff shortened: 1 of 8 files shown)


jgraettinger:
Thank you for putting this together! I'm not a code owner here, but the chunking logic LGTM and this seems reasonable to me given the context.

jgraettinger (approved):

PR #375, opened 2022-11-09
Update logic to remove all bindings.

kiahna-tucker:
## Changes

## Tests

## Issues

## Content

## Screenshots

### DIFF
--- a/src/api/draftSpecs.ts
+++ b/src/api/draftSpecs.ts
@@ -184,12 +184,13 @@ export const deleteDraftSpecsByCatalogName = async (
     const promises: Array<Promise<PostgrestResponse<any>>> = [];
     let index = 0;
 
-    const deletePromiseGenerator = (idx: number) =>
-        supabaseClient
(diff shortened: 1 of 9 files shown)


src/components/collection/Picker.tsx
```diff
@@ -114,30 +120,63 @@ function CollectionPicker({ readOnly = false }: Props) {
...
+    }, [collections, collectionOptions]);
+
+    useEffect(() => {
+        if (populateCollectionValues && collections) {
+            const values = collections.map(
+                (collection) =>
+                    collectionOptions.find(
+                        ({ name }) => name === collection
+                    ) ?? {
+                        name: collection,
```
kiahna-tucker, line 159: The condition that sets `populateCollectionValues` should guarantee that there is an autocomplete option for each collection in the `collections` array, therefore the right-hand side of the nullish coalescing operator should never be reached. This assertion has proven to be true in my rounds of testing.

travjenkins (approved):
lgtm

PR #376, opened 2022-11-10
Implementing new table store and moving more endpoints to API approach

travjenkins:
## Changes

1. Get tables using a store to get data
2. Wire up fetching stats (hidden now)
3. Move table state into a hook
4. Move data/queries to the API approach for table related endpoints

## Tests

Manually tested viewing the tables, sorting, searching, etc.

## Issues

N/A - tech debt

## Content

N/A - this adds content but we'll worry about approving that when the stats is made visible to users

## Screenshots

N/A - should be no visible differences to users


src/AppLayout.tsx
```diff
@@ -8,10 +8,10 @@ import Navigation from './components/navigation/Navigation';
 function AppLayout() {
     const [navigationConfig, setNavigationConfig] = useLocalStorage(
         LocalStorageKeys.NAVIGATION_SETTINGS,
-        { open: true }
+        { open: false }
```
travjenkins, line 11: Not related to table... most folks close by default anyway.

src/components/tables/AccessGrants/index.tsx
```diff
@@ -1,67 +1,60 @@
...
+        return getGrantsForEverything(pagination, searchQuery, [
+            {
+                col: columnToSort,
+                direction: sortDirection,
             },
-        },
-        [pagination, searchQuery, columnToSort, sortDirection]
-    );
+        ]);
+    }, [columnToSort, pagination, searchQuery, sortDirection]);
```
travjenkins, line 29: New approach for queries... just create a fetched based on the API stuff and memoize it.

src/components/tables/AccessGrants/index.tsx
```diff
@@ -1,67 +1,60 @@
...
-                setSortDirection={setSortDirection}
-                columnToSort={columnToSort}
-                setColumnToSort={setColumnToSort}
-                header="accessGrantsTable.title"
-                filterLabel="accessGrantsTable.filterLabel"
+            <TableHydrator
+                query={query}
                 selectableTableStoreName={SelectTableStoreNames.ACCESS_GRANTS}
-            />
+            >
```
travjenkins, line 36: The hydrator gets the query and is the one that actually puts all the stuff into the table store.

src/components/tables/Captures/Rows.tsx
```diff
@@ -21,15 +22,15 @@ import { QUERY_PARAM_CONNECTOR_TITLE } from 'services/supabase';
 import { SelectTableStoreNames } from 'stores/names';
 import { useShardDetail_setShards } from 'stores/ShardDetail/hooks';
 import { getPathWithParams } from 'utils/misc-utils';
-import { LiveSpecsExtQuery } from './types';
 
 interface RowsProps {
-    data: LiveSpecsExtQuery[];
+    data: CaptureQueryWithStats[];
```
travjenkins, line 27: Leaving the stats stuff in but it is not actually getting used.

src/components/tables/Captures/StatsHydrator.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useEffect } from 'react';
+import { SelectTableStoreNames } from 'stores/names';
+import { BaseComponentProps } from 'types';
+
+// Hydrator
+interface StatsHydratorProps extends BaseComponentProps {
+    selectableTableStoreName: SelectTableStoreNames;
+}
+
+const StatsHydrator = ({
```
travjenkins, line 15: This is all still WIP as I am still sketching out code to decide how we want to do this.

src/components/tables/EntityTable.tsx
```diff
@@ -95,22 +94,36 @@ function EntityTable({
...
+
+    const selectDataCount = useZustandStore<
+        SelectableTableStore,
+        SelectableTableStore['query']['count']
+    >(selectableTableStoreName, selectableTableStoreSelectors.query.count);
+
+    const hydrate = useZustandStore<
+        SelectableTableStore,
+        SelectableTableStore['hydrate']
+    >(selectableTableStoreName, selectableTableStoreSelectors.query.hydrate);
```
travjenkins, line 125: Converting what used to be handles with SWR into fetching out of a store

src/components/tables/EntityTable.tsx
```diff
@@ -152,9 +214,9 @@ function EntityTable({
 
     useEffect(() => {
         if (successfulTransformations > 0) {
-            mutateSelectData().catch(() => {});
+            hydrate();
```
travjenkins, line 217: We no longer have mutate to trigger a re-fetch so calling hydrate directly.

src/components/tables/Filters/Date.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+type FilterOptions =
+    | `today`
+    | `yesterday`
+    | `lastWeek`
+    | `thisWeek`
+    | `lastMonth`
+    | `thisMonth`;
+
+function DateFilter() {
```
travjenkins, line 15: All WIP and not being used right now as this is for stats.

src/components/tables/hooks.ts
```diff
@@ -0,0 +1,28 @@
+import { useState } from 'react';
+import { SortDirection } from 'types';
+import { getPagination } from './EntityTable';
+
+function useTableState(defaultSortCol: any, defaultSortDir?: SortDirection) {
```
travjenkins, line 5: Taken from the repeated stuff in table components. More should go in here eventually... maybe. Some of this kind of stuff might go into the store... but even if it does we'll probably get it all out via a hook.

src/forms/renderers/MultiLineSecret.tsx
```diff
@@ -79,36 +79,38 @@ const MultiLineSecretRenderer = (props: any) => {
...
-                }}
-                dialogProps={{
-                    open: isUploadOpen,
-                    sx: {
-                        '& .MuiDropzoneArea-root': {
-                            'minHeight': 150,
-                            'padding': 5,
-                            '& .MuiDropzonePreviewList-root': {
-                                justifyContent: 'center',
+            {isUploadOpen ? (
```
travjenkins, line 82: Super simple fix(?) for React complaining that components were doing stuff after not being rendered.

src/hooks/supabase-swr/hooks/useSelect.ts
```diff
@@ -12,4 +12,15 @@ const useSelect = <Data>(
     return useSWR(query, fetcher, swrConfig);
 };
 
+export const useSelectNew = <Data>(
```
travjenkins, line 15: One step towards using a better approach for integrating with SWR

src/hooks/supabase-swr/query.ts
```diff
@@ -6,7 +6,7 @@ export type QueryConfig<Data> = {
     count?: Count;
     head?: boolean;
 };
-export type Query<Data> = [string | (() => string), QueryConfig<Data>] | null;
+export type Query<Data> = [string, QueryConfig<Data>] | null;
```
travjenkins, line 9: We never had a function that returned the string so removed that type

src/stores/EndpointConfig.tsx
```diff
@@ -207,6 +201,7 @@ const getInitialState = (
     get: StoreApi<EndpointConfigState>['getState']
 ): EndpointConfigState => ({
     ...getInitialStateData(),
+    ...getStoreWithHydrationSettings('Endpoint Config', set),
```
travjenkins, line 204: Make standard stuff with hydration standard.

src/stores/Hydration.ts
```diff
@@ -0,0 +1,57 @@
...
+                    state.hydrationErrorsExist = value;
+                }),
+                false,
+                `${key} Hydration Errors Detected`
+            );
+        },
+    };
+};
+
+export interface AsyncOperationProps {
```
travjenkins, line 42: This chunk is a WIP for handling async ops in Zustand in a standard way.

src/api/combinedGrantsExt.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    if (adminOnly) {
+        queryBuilder = queryBuilder.eq('capability', 'admin');
+    }
+
+    return queryBuilder.eq('user_id', userId);
+};
+
+// Used to display all the grants for everything in the admin page
+const getGrantsForEverything = (
```
kiahna-tucker, line 23: The query helper naming convention will take a few iterations to see what _feels right_, so I am comfortable moving forward with the original function name. That said, the naming convention applied to the hydration-related query helpers is as follows:

**Differentiating queries that pull the same classification of data from a table (e.g., endpoint schema and/or resource schema from the `connector_tags` table):**

> `getDataClassification` -- an overarching query that pulls multiple forms of a data classification (e.g., both the resource and endpoint schemas from the `connector_tags` table).
> 
> `getDataClassification_SpecificIdentifier` -- a query that pulls a single form of a data classification (e.g., either the resource schema or the endpoint schema from the `connector_tags` table).

This was inspired by the store selector hook naming convention. It can be observed by the `getSchema_Endpoint` and `getSchema_Resource` helpers in the aforementioned file.

<br />

**Differentiating queries that pull data from the same column(s) of a table (e.g., the `spec` column of the `live_specs_ext` table):**

>`getDataClassificationByFocalColumn` -- a query that pulls data from a specific column(s) of a table where the _focal column_ is the distinguishing filter applied.

With those definitions in mind, `getGrants` is a suitable name alternative for `getGrantsForEverything` since the more nuanced queries identify the _use-case_ or distinguishing factor in the name. I would _recommend_ that adjustment. As far as the other query helper names go, I feel the `getDataClassification_SpecificIdentifier` definition is most applicable since you are targeting the `Grant` type explicitly (i.e., `getGrants_User` and `getGrants_AuthToken`). _Thinking out loud_.

src/api/directives.ts
```diff
@@ -74,4 +79,27 @@ const submitDirective = async (
     }
 };
 
-export { exchangeBearerToken, submitDirective };
+const getApplieDirectives = (type: keyof typeof DIRECTIVES, userId: string) => {
```
kiahna-tucker, line 82: >`getApplieDirectives`

There is a typo in this function name. The _d_ in _Applied_ is missing (i.e., `getAppliedDirectives`).

kiahna-tucker, line 82: >`(type: keyof keyof DIRECTIVES`

Since `type` is a keyword, I would suggesting renaming this input argument. Also, I have never seen `keyof` used in succession before. Interesting.

src/api/directives.ts
```diff
@@ -74,4 +79,27 @@ const submitDirective = async (
...
+            directives !inner(spec->>type)
+        `);
+
+    queryBuilder = queryBuilder.eq('directives.spec->>type', type);
+    queryBuilder = queryBuilder.eq('user_id', userId);
+
+    return DIRECTIVES[type]
+        .queryFilter(queryBuilder)
+        .order('updated_at', { ascending: false })
+        .limit(1);
```
kiahna-tucker, line 102: I have been wondering if queries that should only return a single result should be explicitly limited. How do you feel about imposing explicit limits where expected?

src/api/liveSpecsExt.ts
```diff
@@ -0,0 +1,127 @@
...
+export interface CaptureQueryWithStats extends CaptureQuery {
+    stats?: CatalogStats;
+}
+export interface MaterializationQuery extends LiveSpecsExtBaseQuery {
+    reads_from: string[];
+}
+export interface MaterializationQueryWithStats extends MaterializationQuery {
+    stats?: CatalogStats;
+}
+export type CollectionQuery = Pick<
```
kiahna-tucker, line 41: Query typing has been on my mind pretty heavily, particularly because of the constraints they can impose on segments of a store state. The `spec` and `setSpecs` slices of the editor store state are a good example of a _query type constraint_ place on a store. Here, it is important that there is a `draft_specs` query type that is an analog of a `live_specs` query. This is easy enough to achieve, but as we extend/prune query types and apply them to query helper functions, utility functions, and store assets, it may prove beneficial to have a _specific_ approach for query type generation. An idea periodically comes to mind: create a base type that identifies the structure of an entire table and utilize the `Pick` utility to create common subsets of that type. _Thinking out loud._

src/components/tables/Filters/Date.tsx
```diff
@@ -0,0 +1,102 @@
...
+                id="stat-filter-selector-menu"
+                anchorEl={anchorEl}
+                open={open}
+                onClose={handlers.closeMenu}
+                anchorOrigin={{
+                    vertical: 'bottom',
+                    horizontal: 'left',
+                }}
+            >
+                <MenuItem onClick={() => handlers.setFilter('today')}>
```
kiahna-tucker, line 79: Did you consider using employing `map` to generate this list of menu items?

src/hooks/supabase-swr/hooks/useSelectSingle.ts
```diff
@@ -12,4 +13,17 @@ const useSelectSingle = <Data>(
...
+export const useSelectSingleNew = <Data>(
+    fetcher: any,
+    swrConfig?: Omit<SWRConfiguration, 'fetcher'>
+): SWRResponse<SuccessResponseSingle<Data>, PostgrestError> => {
+    const key = useMemo(() => (fetcher ? fetcher.url.href : null), [fetcher]);
+    const fetchFunction = useMemo(
+        () => () => fetcher.throwOnError().single(),
+        [fetcher]
+    );
+    console.log('select single new', key);
```
kiahna-tucker, line 25: Are you intentionally leaving this `console.log` behind?

kiahna-tucker (approved):
Approved without manual testing since a _decent_ portion of this functionality was demoed. There is one typo that I would like to have fixed before this code is merged and an opportunity for code cleanup (i.e., lingering `console.log`). All other comments can be considered a discussion point.

src/api/directives.ts
travjenkins: Well that explains why I kept having issues finding this function when searching for it!

Fixed

travjenkins: I considered that - but they are stored as `type` so sticking with it.

src/api/directives.ts
travjenkins: I have been wondering about that. I think it is a good idea when we know the exact count since if you do this and there are multiples returned then an error is thrown. 

However, I kinda of don't like queries with limits ever and would rather just always have arrays and handle multiples from day one (in case we ever need to handle multiples).

In our system I am feeling we're probably safe limiting the queries we think should because they are really rare so kinda of nice to call out those out.

src/api/liveSpecsExt.ts
travjenkins: Yeah - I think once we get the Supabase CLI updated and have it type the DB this will be a lot easier. As we'll always be starting with "a base type" of the table. 

I agree with the thinking out loud - as I was coding like this for this PR it felt alright and easy to work with.

src/components/tables/Filters/Date.tsx
travjenkins: Yup - I'm gonna eventually make all this read from some kind of "filter settings" and generate the menu but only after I know how they are actually different.

src/hooks/supabase-swr/hooks/useSelectSingle.ts
travjenkins: I was... but I feel like this is working well enough to not need anymore. I'll remove.

src/api/combinedGrantsExt.ts
travjenkins: I actually only called is that because it was the last query I found while working through these and was "Oh crap... this should be called getGrants".

travjenkins: I am still not sold on using the selector hooks naming pattern with apis. I played with that name with these and did not like it as much. But agree - it'll take us awhile to figure out how we're gonna name and structure stuff with the different approach.

PR #377, opened 2022-11-14
admin/api: give service account JSON instead of token

mdibaiee:
## Changes

- Instead of token, generate a JSON service account to be used with https://github.com/estuary/flow/pull/792

## Tests

I actually was not able to test this yet, I had issues with running agent locally 😬 

## Issues

- https://github.com/estuary/animated-carnival/issues/43

## Content

_Were there any changes to [content](https://github.com/estuary/ui/tree/main/src/lang) that you need to get reviewed?_ No

## Screenshots


src/components/admin/Api.tsx
```diff
@@ -43,7 +43,13 @@ function AdminApi() {
...
                     minRows={4}
                     cols={50}
-                    value={session?.access_token}
+                    value={JSON.stringify({
+                        access_token: session?.access_token,
+                        refresh_token: session?.refresh_token,
+                        expires_at:
+                            (session?.expires_at ?? 0) +
+                            (session?.expires_in ?? 0),
+                    })}
```
travjenkins, line 52: I think we can just generally check for session as the start

```
value={session ? JSON.stringify({
    access_token: session.access_token,
    refresh_token: session.refresh_token,
    expires_at:
        (session.expires_at ?? 0) +
        (session.expires_in ?? 0),
}) : ""}
```

The session "can be null" but that usually just means they get logged out right away. If we are missing session then we can just not show anything in the text area.

src/lang/en-US.ts
```diff
@@ -336,8 +336,8 @@ const AdminPage: ResolvedIntlConfig['messages'] = {
     'admin.roles.message': `These are all the ${CommonMessages['terms.permissions']} that are currently provisioned. An administrator can update them in the {docLink}.`,
     'admin.roles.message.docLink': `authorization settings`,
     'admin.roles.message.docPath': `https://go.estuary.dev/provision`,
-    'admin.accessToken': `Access Token`,
-    'admin.accessToken.message': `Copy the access token below to authenticate the CLI client.`,
+    'admin.accessToken': `Service Account`,
+    'admin.accessToken.message': `Copy the command below to authenticate the CLI client.`,
```
travjenkins, line 340: Since we are now giving the user a command to run in the terminal I think we need to include more details or at the very least link to the docs page: https://docs.estuary.dev/reference/authentication/#authenticating-flow-using-the-cli

Also... we'll need to update the docs page when we merge this.

Issue #326, opened 2022-09-13
Recommended connectors are not listed first in card view


dyaffe:
**Problem**
We have a "recommended field" that we use to order connectors which we show in the UI and the new card-based implementation is ignoring them.

travjenkins:
Yes... we never implemented this. We need to discuss requirements and stuff. I'll change this over to an enhancement.

dyaffe:
That's not true, it has always been there.  You can verify that it has when you select a connector and then check the drop down in the UI.  You will see that the drop down is ordered with the recommended field so this should not be an enhancement.  It is taking away functionality that previously existed.

travjenkins:
Oh okay - we'll fix it then.

dyaffe:
One comment -- I'm actually seeing that the drop down is now flipped.  The recommended ones are now shown last.

dyaffe:
Reopening since we still don't display the cards in order of recommended first.

PR #378, opened 2022-11-14
Ordering connector cards and adding label

travjenkins:
## Changes

1. Have connector tiles order with recommended first
2. Add the "recommended" label to cards based on the one on estuary.dev
3. Refactor default table filter to support multiple filters

## Tests

Manually tested

## Issues

Fixes: https://github.com/estuary/ui/issues/326

## Content

Added the common word "recommended" taken from estuary.dev

## Screenshots

image

### DIFF
--- a/src/components/ConnectorTiles.tsx
+++ b/src/components/ConnectorTiles.tsx
@@ -24,7 +24,12 @@ import {
 } from 'hooks/useConnectorWithTagDetail';
 import { useEffect, useMemo, useRef, useState } from 'react';
 import { FormattedMessage, useIntl } from 'react-intl';
-import { CONNECTOR_NAME, defaultTableFilter, TABLES } from 'services/supabase';
+import {
+    CONNECTOR_NAME,
+    CONNECTOR_RECOMMENDED,
+    defaultTableFilter,
+    TABLES,
+} from 'services/supabase';
 import {
     BaseComponentProps,
     EntityWithCreateWorkflow,
@@ -43,6 +48,7 @@ import ConnectorCardTitle from './connectors/card/Title';
 interface ConnectorTilesProps {
     protocolPreset?: EntityWithCreateWorkflow;
     replaceOnNavigate?: boolean;
+    hideSort?: boolean;
 }
 
 type TileProps = BaseComponentProps;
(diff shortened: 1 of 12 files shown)


src/components/ConnectorTiles.tsx
```diff
@@ -114,8 +119,16 @@ function ConnectorTiles({
...
+                    [
+                        {
+                            col: CONNECTOR_RECOMMENDED,
+                            direction: 'desc',
+                        },
+                        {
+                            col: columnToSort,
+                            direction: sortDirection,
+                        },
+                    ],
```
travjenkins, line 131: Now we can pass an array of sorting options. So we always hardcode the sorting of recommended first.

src/components/connectors/card/index.tsx
```diff
@@ -53,22 +63,56 @@ function ConnectorCard({ cta, docsUrl, logo, title, details }: Props) {
...
+                            <Box
+                                sx={
+                                    recommended
+                                        ? {
+                                              ...connectorImageBackgroundSx,
+                                              mb: 0,
+                                              borderBottomLeftRadius: 0,
+                                              borderBottomRightRadius: 0,
+                                          }
+                                        : connectorImageBackgroundSx
```
travjenkins, line 80: If it is recommended we don't want the border radius on the bottom so the label can have the rounded corner.

src/components/connectors/card/index.tsx
```diff
@@ -53,22 +63,56 @@ function ConnectorCard({ cta, docsUrl, logo, title, details }: Props) {
...
+                                            connectorImageBackgroundRadius,
+                                        borderBottomRightRadius:
+                                            connectorImageBackgroundRadius,
+                                        opacity: 0.75,
+                                        textAlign: 'center',
+                                    }}
+                                >
+                                    <FormattedMessage id="common.recommended" />
+                                </Box>
+                            ) : null}
```
travjenkins, line 114: Basic box that leaned on the approach of estuary.dev to follow. The purple color did not work well since the CTA is already using that. `teal[500]` with some opacity works well for both dark and light so went with that.

src/components/tables/AccessGrants/index.tsx
```diff
@@ -28,8 +28,12 @@ function AccessGrantsTable() {
...
                     ['user_full_name', 'subject_role', 'object_role'],
                     searchQuery,
-                    columnToSort,
-                    sortDirection,
+                    [
+                        {
+                            col: columnToSort,
+                            direction: sortDirection,
+                        },
+                    ],
```
travjenkins, line 36: We now always pass an array of sorting options so this change happened in a lot of files.

kiahna-tucker (approved):

Issue #382, opened 2022-11-16
Make Sign Up vs Registration more apparent 


travjenkins:
Right now it is confusing for a new user (of even existing ones that have forgotten) if they are signing up or registering for Flow.

We should more the UI login page to something like this:
![image](https://user-images.githubusercontent.com/270078/202260269-0ce5e88e-0239-4707-8ef0-d0ee34d3a321.png)

Reference:
https://ux.stackexchange.com/questions/11840/best-way-to-combine-sign-in-and-sign-up-in-the-same-form

PR #384, opened 2022-11-17
Make registration apparent on login page

travjenkins:
## Changes

1. Add new path that shows registration first
2. Show login/registration tabs
3. Change language for registration

## Tests

Manual

## Issues

https://github.com/estuary/ui/issues/382

## Content

Some minor message changes for stuff where `login` becomes `registration`

## Screenshots
image

Magic link is only used to login on locals
image

### DIFF
--- a/src/app/Unauthenticated.tsx
+++ b/src/app/Unauthenticated.tsx
@@ -11,6 +11,10 @@ const Unauthenticated = () => {
                 path={unauthenticatedRoutes.magicLink.path}
                 element={<Auth />}
             />
+            <Route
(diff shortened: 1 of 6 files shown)


kiahna-tucker (approved):
Approved without manual testing. I do believe the presence of the term _register_ in variable names (and intl IDs) reads awkwardly to me. At the least, I would replace those instances with the term _registration_ (e.g., `showRegister` in _src/pages/Login.tsx_ would become `showRegistration`). Nevertheless, this is a non-blocking suggestion.

PR #391, opened 2022-11-29
Phil/flowctl auth 2

psFried:
## Changes

Adds support for a new way for `flowctl` to authenticate.  First, the `/cli-auth/login` page was added. Once a user logs in there, they are redirected to `/cli-auth/success`. Both of those pages use separate Context providers so that they will use an entirely separate go-true session and local storage entry than the rest of the web app. Effectively, `flowctl` has its own separate session, and only it will attempt to refresh the credentials for it.

As part of this, I had to do a bit of refactoring to allow re-using our existing `Login` component with different redirect urls. I also fixed an issue where redirects weren't working if the user manually entered the OTP rather than clicking the link from the email. In that case, we need to handle the redirect ourselves, since the `/auth/v1/validateToken` endpoint will not respond with redirects.

There's definitely still a bunch of room for improvement here. Some things that I'd like to eventually see:

- The login page for the CLI being differentiated from the general login page
- Better handling for if a user ends up on `/cli-auth/success` without there being a session in localStorage. Currently, we just generate a bogus token value, which is pretty rough.
- Passing the credential to the CLI via an XHR request.
- Some sort of UX around logging out and invalidating the CLI session / refresh token

Note that this leaves the existing Admin - API tab untouched for the time being, so that current CLI users are not impacted. Once the new `flowctl` is released, we can remove that tab or replace it with a note about using `flowctl auth login` instead.

## Tests

I've tested this manually by navigating to the `/cli-auth/login` url and going through the flow end to end with `flowctl`.

## Content

There's a little bit of copy that was added.

## Screenshots

After a successful login:

image

### DIFF
--- a/src/app/Unauthenticated.tsx
+++ b/src/app/Unauthenticated.tsx
@@ -13,9 +13,19 @@ const Unauthenticated = () => {
             />
             <Route
                 path={unauthenticatedRoutes.register.path}
-                element={<Login showRegistration={true} />}
+                element={
+                    <Login
+                        showRegistration={true}
+                        redirectTo={`${window.location.origin}/auth`}
+                    />
+                }
             />
-            <Route path="*" element={<Login />} />
+            <Route
+                path="*"
+                element={
+                    <Login redirectTo={`${window.location.origin}/auth`} />
+                }
+            />
         </Routes>
     );
 };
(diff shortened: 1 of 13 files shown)


travjenkins (approved):

psFried:
Closing, because this simply isn't going to work. See estuary/flow#821 for the explanation.

PR #393, opened 2022-12-01
jsonforms: keep Z pattern in forms

mdibaiee:
## Changes

- Keep the Z pattern in forms as the literal `Z` stands for UTC and that's what connectors expect

## Tests

Tested locally

## Issues

## Content

## Screenshots

Local test:
image


travjenkins (approved):
lgtm - I just removed the processing since it isn't needed.

Issue #381, opened 2022-11-16
Integrate Google Analytics and Hubspot with Registration Page


dyaffe:
**Background**
PLG requires that we integrate analytics and CRM to capture leads and automate responses.  Our chosen systems for that, so far, are Hubspot and Google Analytics and we should integrate those.

**Hubspot**
Hubspot has a [form api](https://legacydocs.hubspot.com/docs/methods/forms/submit_form), which appears to be the best thing to integrate with our internal forms.  Here is the embed code for it which I believe has all the information we'd need to use it (ex. portal ID and form ID):

```
<script charset="utf-8" type="text/javascript" src="//js.hsforms.net/forms/embed/v2.js"></script>
<script>
  hbspt.forms.create({
    region: "na1",
    portalId: "8635875",
    formId: "1e475d85-2300-42b3-babf-9c179af2ac9f"
  });
</script>
```

**Google Analytics**
GA is integrated through Google tag manager and this gets slightly more complicated.  If we integrate Google Tag manager into the page, everything can be managed through that.  Here are instructions on how to do so:

Paste this code as high in the <head> of the page as possible:
```
<!-- Google Tag Manager -->
<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
})(window,document,'script','dataLayer','GTM-WK8SB2L');</script>
<!-- End Google Tag Manager -->
```

Paste this code immediately after the <body> tag:
```
<!-- Google Tag Manager (noscript) -->
<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-WK8SB2L"
height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
<!-- End Google Tag Manager (noscript) -->
```

PR #395, opened 2022-12-01
Add consent management trial and Tag manager

travjenkins:
## Changes

1. Add in Google tag manager and settings
2. Trial out Osano for compliance. This is starting in LISTENING mode first so should have no impact on users.

## Tests

N/A

## Issues

Fixes https://github.com/estuary/ui/issues/381

## Content

There is some content but it is coming from Osano and not customizable without paying... also might be best to not mess with it to reduce chance of issues with compliance.

## Screenshots

Initial consent banner
image

Banner changes based on country
image
image
image

If Osano is running in discover/listening mode then we do not show any settings to configure
image

When running we show the tab
image

image

image


kiahna-tucker (approved):
Approved without manual review.

Issue #311, opened 2022-08-31
Default Materialization Table Names


dyaffe:
**Problem** 
I was just on a call with a client and got some confirmation that our current requirement for typing out every single binding name within a materialization is _extremely_ painful.  

Every time that I've attempted to create a materialization or a client has, the user has desired it was defaulted for them.

**Proposal**
We should default names when creating a materialization to everything after the last `/`.  That seems to be the desired state for all of them.

travjenkins:
This is now complete.

PR #397, opened 2022-12-02
resource-config: set collection-name on annotated field

mdibaiee:
## Changes

- Automatically fill in field marked with `x-collection-name` with last piece of collection name

## Tests

Tested locally

## Issues

#311 

## Content

N/A

## Screenshots

image

### DIFF
--- a/src/components/collection/ResourceConfigForm.tsx
+++ b/src/components/collection/ResourceConfigForm.tsx
@@ -1,6 +1,7 @@
 import { materialCells } from '@jsonforms/material-renderers';
 import { JsonForms } from '@jsonforms/react';
 import { StyledEngineProvider } from '@mui/material';
+import { cloneDeep } from 'lodash';
 import { useEffect, useRef } from 'react';
 import { setDefaultsValidator } from 'services/ajv';
 import { custom_generateDefaultUISchema } from 'services/jsonforms';
@@ -21,6 +22,8 @@ type Props = {
     readOnly?: boolean;
 };
 
+const COLLECTION_NAME_ANNOTATION = 'x-collection-name';
+
 function ResourceConfigForm({ collectionName, readOnly = false }: Props) {
     const name = useRef(collectionName);
 
@@ -28,7 +31,7 @@ function ResourceConfigForm({ collectionName, readOnly = false }: Props) {
     const resourceConfig = useResourceConfig_resourceConfig();
     const setConfig = useResourceConfig_setResourceConfig();
 
-    const formData = resourceConfig[collectionName].data;
+    let formData = resourceConfig[collectionName].data;
 
     const resourceSchema = useResourceConfig_resourceSchema();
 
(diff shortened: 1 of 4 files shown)


travjenkins (changes requested):
This creates an infinite loop. Do not merge this. I'll take a look at it.

travjenkins (approved):
Switched the approach to setting the `default` property in the schema. This allows `ajv` to set the value once it makes it into `jsonForms`.

Please test and make sure this is still working as expected for you.

Also... I am assuming we're taking the approach that only a single field should ever have this annotation.

src/components/collection/ResourceConfigForm.tsx
```diff
@@ -50,10 +50,40 @@ function ResourceConfigForm({ collectionName, readOnly = false }: Props) {
...
+                    ([_, value]) =>
+                        value && Annotations.defaultResourceConfigName in value
+                ) ?? [];
+
+            // Try to fetch the key
+            return collectionNameField[0];
+        }
+
+        return null;
+    }, [resourceSchema.properties]);
```
travjenkins, line 68: This chunk isn't _too_ heavy but since we don't need to recalculate it often decided to move this into a stand alone memo.

src/components/collection/ResourceConfigForm.tsx
```diff
@@ -50,10 +50,40 @@ function ResourceConfigForm({ collectionName, readOnly = false }: Props) {
...
+        if (collectionNameFieldKey) {
+            // Add a default property set to the stripped collection name
+            const response = cloneDeep(resourceSchema);
+            response.properties[collectionNameFieldKey].default =
+                stripPathing(collectionName);
+            return response;
+        }
+
+        return resourceSchema.properties;
+    }, [collectionName, collectionNameFieldKey, resourceSchema]);
```
travjenkins, line 81: Make a memo of the schema which will either be the normal schema or one with the injected default value.

src/index.tsx
```diff
@@ -24,7 +24,10 @@ import AppProviders from './context';
 
 initGoogleTagManager();
 initLogRocket();
+
+// Setup immer
 enableMapSet();
+setAutoFreeze(false);
```
travjenkins, line 30: This allows AJV to more easily manipulate the data being passed to it. Related to this change but one I have been eyeing for a while so made it here.

src/services/ajv.ts
```diff
@@ -30,6 +32,7 @@ export const addKeywords = (ajv: any) => {
     ajv.addKeyword('advanced'); // Should be collapsed by default (over ridden if section contains required fields)
     ajv.addKeyword('order'); // Used to order the fields in the UI
     ajv.addKeyword('x-oauth2-provider'); // Used to display OAuth
+    ajv.addKeyword('x-collection-name'); // Used to default name in resource configs
```
travjenkins, line 35: We keep all custom annotations in here so we can keep them all documented together.

kiahna-tucker (approved):

PR #401, opened 2022-12-06
prod: Remove entity name changed listener from edit workflows

kiahna-tucker:
## Changes

Remove entity name changed listener from edit workflows. The details form is inaccurately detecting a change in the entity name field upon entry into the edit workflow.


travjenkins (approved):
lgtm

Issue #404, opened 2022-12-07
Able to submit an invalid form during create


travjenkins:
**Bug**
If you submit the form while nothing is entered, add a name, and then submit again you can submit the form.

**Expected**
The form should not submit and display errors.

**Screenshots**
Step 1 - click the next button to see the details and endpoint config errors
![image](https://user-images.githubusercontent.com/270078/206086783-158045cc-4555-4204-aab4-1203438b8feb.png)

Step 2 - fill in the entity name
![image](https://user-images.githubusercontent.com/270078/206086820-d26327c0-f7fe-40ed-a2e8-f25bd1244a33.png)

Step 3 - click the next button and the form submits
![image](https://user-images.githubusercontent.com/270078/206086884-4d753a66-a698-47a5-b1eb-8e7b92a54ee1.png)

PR #406, opened 2022-12-07
Prevent submission of an invalid form

kiahna-tucker:
## Changes

1. Revise the endpoint config error flag condition for the _Next_ CTA in the capture edit workflow.

1. Listen for resource config errors when the _Next_ CTA is clicked in a capture workflow and prevent discovery if one exists.

1. Consolidate the two form validation error components.

## Tests

Only manual testing of the capture workflows and the materialization create workflows was performed. I am unable to test the materialization edit workflow locally.

## Issues

#404 

## Screenshots


src/components/capture/GenerateButton.tsx
```diff
@@ -78,21 +80,26 @@ function CaptureGenerateButton({ disabled, callFailed, subscription }: Props) {
...
     const serverUpdateRequired = useEndpointConfig_serverUpdateRequired();
 
     // Resource Config Store
     const resourceConfig = useResourceConfig_resourceConfig();
+    const resourceConfigErrorsExist =
+        useResourceConfig_resourceConfigErrorsExist();
 
     const endpointConfigErrorFlag = editWorkflow
-        ? endpointConfigChanged() && endpointConfigErrorsExist
+        ? endpointConfigErrorsExist
```
kiahna-tucker, line 91: Looking at the logic, I cannot see a scenario where the encrypted endpoint config in the store could differ from that on the server (which is the evaluation performed by the `endpointConfigChanged` function) that should affect the generate button. `endpointConfigChanged` appropriately influences whether the data loss modal appears if you attempt to navigate away from the page, but I do not think it should factor in here. Thoughts?

travjenkins: Sounds right to me. I was trying to test this same chunk of code in the OAuth branch and could never get this portion to work correctly.

travjenkins (approved):
LGTM

PR #411, opened 2022-12-09
fix: Redirect user to origial URL they were going to use on login

### LINKED ISSUE #399: URLs get lost if user is not logged in before navigation
**Bug**

Here's the steps:

- logout of the ui
- navigate to (https://dashboard.estuary.dev/admin/api)
- login when presented with the page

**Expected**

The expected behavior is that you'd be redirected to the original URL (https://dashboard.estuary.dev/admin/api) after completing the sign-in process.

What's actually happening is that you always just land on the front page.

**Additional context**

I had this fixed in #391, which couldn't be merged for `$reasons`, so I'm creating this issue to ensure that we circle back around and fix this.

The impact here is that `flowctl auth login` opens your browser to `/admin/api`, and if you're not already logged in there, then you'll never be taken to the right page.

jshearer:
## Changes

We want the user to end up on the original URL they were trying to navigate to, even if they have to log in to get there. 

Because [`<Unauthenticated>`](https://github.com/estuary/ui/blob/main/src/app/Unauthenticated.tsx) captures all routes, so long as we don't actually navigate anywhere else before authenticating, we'll automatically be placed at the originally intended page. This happens currently when you use the OTP flow and copy-paste the code from your email -- as soon as you successfully authenticate, we swap to the `<Authenticated>` router and render the page you were trying to get to all along.

In the more likely scenario that you need to go through an indirect flow, currently either by clicking the login link in your email, or using an OIDC provider, we would lose the originally intended route and instead just redirect you to the main page. This PR fixes that by including a `redirect_to` query parameter that's used by the `/auth` route to determine where to redirect you after successfully authenticating you.

Fixes #399

## Tests

I was able to test the OTP/email flows locally, but wasn't able to test google/github since AFAICT those aren't set up to work locally.


src/app/routes.ts
```diff
@@ -81,3 +81,5 @@ export const authenticatedRoutes = {
         path: '*',
     },
 };
+
+export const REDIRECT_TO_PARAM_NAME = 'redirect_to';
```
travjenkins, line 85: Weird request - but can you toss this to a top. It feels like it would be easy to miss known search params listing at the bottom of a routes file.

travjenkins (changes requested):
One small call out

src/app/routes.ts
jshearer: done!

src/components/login/MagicLink.tsx
```diff
@@ -68,6 +69,10 @@ const MagicLink = () => {
 
     custom_generateDefaultUISchema;
 
+    const redirectTo = `${redirectToBase}?${REDIRECT_TO_PARAM_NAME}=${encodeURIComponent(
+        window.location.pathname
```
travjenkins, line 73: Given we have some paths that use search params we should probably include search here.

Example: We put the selected connector id in the URL during create.

travjenkins (changes requested):
Tested out... looks good except for missing search params.

src/components/login/MagicLink.tsx
jshearer: Yuup, good point. Done!

travjenkins (approved):
lgtm. Thanks so much for helping us out!

PR #414, opened 2022-12-12
fix: Large documents causing excessive network traffic

### LINKED ISSUE #374: Fail data preview quicker for large datasets
Phil: I'm noticing an issue with collection preview and large documents. I just opened the details for -------------- collection, which has a bunch of documents in the 2-12MB size range. The UI made 397 requests to read a total of 1.5GB from data plane gateway, and then after about 3 minutes it displays an error saying that it exceeded the maximum number of bytes before reaching the desired number of documents. It feels like the UX could be a lot better in this (admittedly niche) case

Phil:  Could we fetch at most 16MB of data and then whether it's 1 document or many, just display whatever we got?
Phil: 16MMB would guarantee that we get at least one
Phil: but we could also limit it to far less data, and just say that we don't support preview for enormous documents. From my perspective, just arriving at that answer more quickly would still be a big improvement in UX.

jshearer:
## Changes

* Bump the range grow rate to start at 1mb
* Grow request  exponentially rather than linearly up to the limit
* Bump limit to 16mb, which is the max document size, ensuring we'll always get at least 1 doc if it exists

Also fixed:
* Display data preview even if there are warnings so long as there is any data to display
* Wrap long lines in the JSON preview editor, preventing excessively wide table rows

## Issues

Fixes #374

## Screenshots

Screen Shot 2022-12-12 at 14 41 57
Screen Shot 2022-12-12 at 14.43.19

### DIFF
--- a/src/hooks/useJournalData.ts
+++ b/src/hooks/useJournalData.ts
@@ -219,6 +219,7 @@ function isJournalRecord(val: any): val is JournalRecord {
 const useJournalData = (
     journalName?: string,
     desiredCount: number = 50,
+    // 16mb, which is the max document size, ensuring we'll always get at least 1 doc if it exists
     maxBytes: number = 16 * 10 ** 6
 ) => {
     const [gatewayConfig] = useLocalStorage(
--- a/src/hooks/useJournalData.ts
+++ b/src/hooks/useJournalData.ts
@@ -119,7 +119,7 @@ async function* streamAsyncIterator<T>(stream: ReadableStream<T>) {
 
 // We increment the read window by this many bytes every time we get back
 // fewer than the desired number of rows.
-const INCREMENT = 1024 * 10;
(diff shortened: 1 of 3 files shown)


src/hooks/useJournalData.ts
```diff
@@ -216,7 +219,7 @@ function isJournalRecord(val: any): val is JournalRecord {
 const useJournalData = (
     journalName?: string,
     desiredCount: number = 50,
-    maxBytes: number = 5 * 10 ** 6
+    maxBytes: number = 16 * 10 ** 6
```
travjenkins, line 222: Can you add your PR comment of `16mb, which is the max document size, ensuring we'll always get at least 1 doc if it exists` here as a code comment so we don't lose why this is 16 in the codebase.

travjenkins (approved):
lgtm - did not manually test.

Issue #358, opened 2022-10-23
Binding selector is missing collections


dyaffe:
**Bug**
I expect to see the following collection available to be added to a materialization when editing it, but it isn't: `trial/mothership/estuary_intercom/companies`

I have permission to all of trial/

travjenkins:
Probably due to the default `limit` on queries is 1000. Most customers (probably none... for a while) won't have 1000 collections for awhile.

travjenkins:
Marking with wontfix for right now so I don't keep opening this when looking for tasks :)

travjenkins:
Going to put in a hacky band aid for right now where we will sort the collections by recent update so newer ones are in the list.

PR #415, opened 2022-12-12
Query newest collections first

travjenkins:
## Changes

While we work on rebuilding the collection selector in general... we can sort by `updated_at` do the newest collections are first.

## Tests

Manually tested

## Issues

Progress https://github.com/estuary/ui/issues/358 

## Content

N/A

## Screenshots

N/A

### DIFF
--- a/src/hooks/useLiveSpecs.ts
+++ b/src/hooks/useLiveSpecs.ts
@@ -6,6 +6,8 @@ import { useQuery, useSelect } from './supabase-swr/';
 export interface LiveSpecsQuery {
     catalog_name: string;
     spec_type: string;
+    // Filtering only
+    updated_at: undefined;
 }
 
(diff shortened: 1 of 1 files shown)


src/hooks/useLiveSpecs.ts
```diff
@@ -31,19 +36,17 @@ function useLiveSpecs(specType: string) {
...
         schema: {
             properties: Record<string, any>;
             required?: string[];
         };
         key: string[];
     };
-    spec_type: string;
 }
-const specQuery = ['id', 'catalog_name', 'spec', 'spec_type'];
+const specQuery = queryColumns.concat(['id', 'spec']);
```
travjenkins, line 49: Changed this so that the fields that are shared in the type are shared via the query array.

Issue #389, opened 2022-11-29
Facebook section `Custom Insights` is missing fields


travjenkins:
**Bug**
There are three arrays in the Custom Insight section of the Facebook connector that are not displaying correctly.

**Expected**
Fields should render

**Screenshots**
![image](https://user-images.githubusercontent.com/270078/204571433-7f6422cd-be20-49e8-a812-c44dff1d9591.png)

**Desktop (please complete the following information):**
n/a

**Additional context**
I believe this is due to how these arrays are nested within another array. If you move the `enum` out of the `items` parent the fields render.

jshearer:
So, it took me a little while to dig through the various layers of abstraction here, but the first error here is talking about the field `custom_insights.0.action_breakdowns`, whose schema is defined as:

```json
"action_breakdowns": {
  "default": [],
  "description": "A list of chosen action_breakdowns for action_breakdowns",
  "items": {
    "description": "Generic enumeration.\n\nDerive from this class to define new enumerations.",
    "enum": [
      "action_canvas_component_name",
      "action_carousel_card_id",
      "action_carousel_card_name",
      "action_destination",
      "action_device",
      "action_reaction",
      "action_target_id",
      "action_type",
      "action_video_sound",
      "action_video_type"
    ],
    "title": "ValidActionBreakdowns"
  },
  "title": "Action Breakdowns",
  "type": "array"
}
```

Critically, the inner enum is missing a `"type": "string"` property! Defining the type for this, and the other two enums (`custom_insights.0.breakdowns.0`, and `custom_insights.0.fields.0`) shows the following form:

<img width="1482" alt="Screen Shot 2022-12-12 at 20 26 34" src="https://user-images.githubusercontent.com/4368270/207203459-8a0572ef-2e09-4889-ad34-0c64880b6b1c.png">
<img width="1490" alt="Screen Shot 2022-12-12 at 20 26 41" src="https://user-images.githubusercontent.com/4368270/207203465-32c6cae7-9877-4840-a09c-2006ce59b587.png">

There are still a bunch of missing renderers, but after seeing the above, I strongly suspect that the problem is just that **the schema is incomplete**. I don't really want to just guess what the schema should be, so does someone from the connectors team want to work with me to fix this schema? Maybe @mdibaiee?

jshearer:
A little hack to jsonforms shows this useful debugging output:

[...]

PR #417, opened 2022-12-13
fix: Handle invalid schema properties

jshearer:
## Changes

Rather than rendering broken-looking forms full of error messages about missing renderers, let's filter out invalid fields, and log an error when we see them 

## Issues

Related to #389


jshearer:
I did think of that actually. I bet with just a little more work I could change this to show a helpful error on the page. The problem with including `null`s was that it was showing _incorrect_ errors on the page (it was including all the errors from the root) so my thought was console error is better than incorrect error, but correct error would be even better.

jshearer:
I'll do that. Out of curiosity, why was it removed?

jshearer:
How's this look? Screen Shot 2022-12-15 at 14 25 32

travjenkins (approved):
lgtm

PR #418, opened 2022-12-13
fix: Allow submitting empty connector configs

### LINKED ISSUE #392: Empty config does not work for capture connector
**Bug**
We have removed configuration of Hello World source, because it doesn't really need any configuration, it's just an infinite list of Hello Worlds. But this doesn't seem to work with the UI as clicking "Discover" doesn't do anything.

**Expected**
I can click "Discover" and then publish this capture.

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Firefox

**Additional context**

### LINKED ISSUE #416: Cannot Test/Save entity that uses all connector defaults for endpoint config
**Bug**
1. Create a capture
2. Select Hello World
3. Click Next

**The next button still shows**

**Expected**
The Next button is displayed

**Screenshots**
image

**Desktop (please complete the following information):**
N/A

**Additional context**
This only happens when you do not edit the Hello World endpoint config.

jshearer:
## Changes

There's still something weird about this. #392 claims that the connector was updated, but I'm actually still getting an error about missing config once applying this change to allow the UI to submit an empty config object

Screen Shot 2022-12-12 at 22 53 58

## Issues

Fixes #392
Fixes #416 (I think?)

### DIFF
--- a/src/stores/EndpointConfig.tsx
+++ b/src/stores/EndpointConfig.tsx
@@ -215,6 +215,14 @@ const getInitialState = (
         set(
             produce((state) => {
                 state.endpointSchema = val;
+
+                const { endpointConfig, endpointCustomErrors } = get();
+
+                populateEndpointConfigErrors(
+                    endpointConfig,
+                    endpointCustomErrors,
+                    state
+                );
             }),
             false,
             'Endpoint Schema Set'
@@ -279,7 +287,7 @@ const getInitialState = (
                     : endpointConfig;
 
                 populateEndpointConfigErrors(
-                    endpointConfig,
+                    state.endpointConfig,
                     endpointCustomErrors,
                     state
                 );
--- a/src/components/capture/useDiscoverCapture.ts
+++ b/src/components/capture/useDiscoverCapture.ts
@@ -132,9 +132,14 @@ function useDiscoverCapture(
         useResourceConfig_resourceConfigErrorsExist();
     const resetCollections = useResourceConfig_resetConfigAndCollections();
 
+    // Errors if:
+    // * If we're in the edit workflow, defer to state
+    // * If we're in create workflow
+    //   * If we have an empty config, then there's no error
+    //   * If we don't have an empty config, defer to state
     const endpointConfigErrorFlag = editWorkflow
         ? endpointConfigErrorsExist
-        : endpointConfigErrorsExist || isEmpty(endpointConfigData);
+        : !isEmpty(endpointConfigData) && endpointConfigErrorsExist;
 
     const storeDiscoveredCollections = useCallback(
         async (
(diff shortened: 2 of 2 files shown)


jshearer:
I also see that the connector is still on `:v1` tag. Is the updated connector pushed to `:v1`, or do we need to change that to get this to work?

travjenkins (approved):
Looks good. 

I tested Hello World permutations : 
1. Greetings defaulted
2. Greetings with no default
3. Greetings removed (empty config)

All worked fine (in terms of the UI)

travjenkins (changes requested):
This only works if all of our schemas set default values. If a schema contains a require prop without a default - the load of the form will mean the endpointConfigData is empty but there are errors and you could submit an invalid form.

This schema will not be handled correctly:
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "Test Source Spec",
  "type": "object",
  "required": [
    "greetings"
  ],
  "properties": {
    "greetings": {
      "title": "Number of Greetings",
      "description": "Number of greeting documents to produce when running in non-tailing mode",
      "type": "integer"
    }
  }
}
```

Which renders this form
image


And the hook gets these props
image

jshearer:
Alright @travjenkins, I think this is better. Rather than trying to pile logic onto guessing whether the form was invalid or not, I just added a call to `populateEndpointConfigErrors` in EndpointConfig's `setEndpointSchema` so whenever we load the schema, we run the validation. My understanding is this should do the right thing: 
* If there are no fields, then there's nothing to validate, and errors will be false
* If there are fields but they all have defaults, then the error will also be false
* If there are fields and some/all don't have defaults, then the error will be true since we're missing data for those fields

There also appears to be a bug in the backend which prevents us from parsing sops-encrypted empty documents, which I have a fix for here:
https://github.com/estuary/flow/pull/854

src/stores/EndpointConfig.tsx
```diff
@@ -215,6 +215,14 @@ const getInitialState = (
...
             produce((state) => {
                 state.endpointSchema = val;
+
+                const { endpointConfig, endpointCustomErrors } = get();
+
+                populateEndpointConfigErrors(
+                    endpointConfig,
+                    endpointCustomErrors,
+                    state
+                );
```
travjenkins, line 225: This should be good and is a funny side effect of how things run.

When the endpoint schema is being set there should be no `endpointConfig` possible so we could honestly just pass an empty default here. However, I think we're fine just using whatever was last in the store. When switching connectors passing an old config here isn't an issue because right after this call we'll set the `endpointConfig` and then call the `populateEndpointConfigErrors` function.

travjenkins (approved):
I think this looks good.

I tested out editing my Google sheets captures and those were good. I think this is a good solution as this draws create/edit even closer in how they run.

Issue #370, opened 2022-11-08
Add stats to entity tables


travjenkins:
Right now there is no way to tell how much data is getting consumed in the system. We need to display some details in the entity tables and maybe even a performance dashboard kind of thing.

Each table should get 1-3 columns of data that show how much data is flowing through the system.

Captures idea:
![image](https://user-images.githubusercontent.com/270078/200667847-d45179f3-dccf-4a21-b980-fc379127992a.png)

Materialization idea:
![image](https://user-images.githubusercontent.com/270078/200667962-221c02f3-156c-463b-aca9-928b4763124c.png)

dyaffe:
I think we can safely get rid of "Reduced output" btw.

PR #422, opened 2022-12-14
Add stats

travjenkins:
## Changes

1. Add in stats for Captures, Materializations, and Collections
2. Add multiple ranges for stats
3. Hide stats if no tenant available for user

## Tests

Manually tested with fake data

## Issues

Progresses https://github.com/estuary/ui/issues/370

## Content

New content for the range selector and columns

## Screenshots

Docs will display up to 0-999,999 as a number
image

1,000,000 and above will be humane readable
image

Unformatted details are shown on hover
image
image

Default of the settings to "Today" and shown inside the column header
image

Options for changing range of stats
image

Loading state
image

Loading state if it takes longer than 800ms
image


package.json
```diff
@@ -72,6 +72,7 @@
         "react-social-login-buttons": "^3.6.1",
         "react-stay-scrolled": "^8.0.0",
         "react-use": "^17.4.0",
+        "readable-numbers": "^1.0.7",
```
travjenkins, line 75: Used to format long numbers into a short human readable format to save space in the table

src/components/tables/Captures/Rows.tsx
```diff
@@ -49,14 +54,18 @@ export const tableColumns = [
...
+        renderHeader: (index, selectableTableStoreName) => {
+            return (
+                <StatsHeader
+                    key={`captures-statsHeader-${index}`}
+                    header="entityTable.stats.written"
+                    selectableTableStoreName={selectableTableStoreName}
+                />
+            );
+        },
+    },
```
travjenkins, line 68: Entity table now allows you to inject your own header component.

src/components/tables/Captures/Rows.tsx
```diff
@@ -142,7 +163,7 @@ function Row({ isSelected, setRow, row, showEntityStatus }: RowProps) {
             <DetailsPanel
                 detailsExpanded={detailsExpanded}
                 lastPubId={row.last_pub_id}
-                colSpan={tableColumns.length}
+                colSpan={tableColumns.length + 1}
```
travjenkins, line 166: Now that one of the column headers has a `colSpan` of 2 we need to add 1 here.

src/components/tables/Captures/index.tsx
```diff
@@ -34,31 +35,39 @@ function CapturesTable() {
...
-                    setColumnToSort={setColumnToSort}
-                    header="captureTable.header"
-                    filterLabel="capturesTable.filterLabel"
-                    enableSelection
-                    rowSelectorProps={{
-                        selectableTableStoreName: SelectTableStoreNames.CAPTURE,
-                        showMaterialize: true,
-                    }}
-                    showEntityStatus={true}
+                <StatsHydrator
```
travjenkins, line 38: Adding back in the stats hydrator

src/components/tables/Collections/Rows.tsx
```diff
@@ -41,11 +57,33 @@ export const tableColumns = [
...
+    const calculatedDocs = useMemo(() => {
+        if (stats) {
+            return (
+                (stats[row.catalog_name]?.docs_written_by_me ?? 0) +
+                (stats[row.catalog_name]?.docs_written_to_me ?? 0)
+            );
+        } else {
+            return 0;
+        }
+    }, [row.catalog_name, stats]);
```
travjenkins, line 85: Collection needs to have these in a memo because it is just easier to write the calculations like this instead of inside nested ternaries down below.

src/components/tables/EntityTable.tsx
```diff
@@ -43,14 +43,19 @@ import {
...
 import { RowSelectorProps } from './RowActions/types';
 
+export interface ColumnProps {
+    field: string | null;
+    headerIntlKey?: string | null;
+    renderHeader?: (
+        index: number,
+        storeName: SelectTableStoreNames
+    ) => ReactNode;
+}
```
travjenkins, line 55: Pulled out this type so we could import in the `Rows` files when setting up a column array.

src/components/tables/Store.ts
```diff
@@ -9,11 +9,27 @@ import {
     getStoreWithHydrationSettings,
     StoreWithHydration,
 } from 'stores/Hydration';
-import { Schema } from 'types';
 import { devtoolsOptions } from 'utils/store-utils';
 import create, { StoreApi } from 'zustand';
 import { devtools, NamedSet } from 'zustand/middleware';
 
+export interface StatsSchema {
```
travjenkins, line 16: This is here just for the tables to render. Eventually this will get moved into the same type that is used with the query.

src/components/tables/Store.ts
```diff
@@ -115,50 +140,81 @@ export const getInitialState = (
...
+                            'Table Store Stats Hydration Success'
+                        );
+                    } else {
+                        set(
+                            produce((state) => {
+                                state.stats = [];
+                            }),
+                            false,
+                            'Table Store Stats Hydration Success : Empty'
+                        );
```
travjenkins, line 216: Making sure an empty response is handled and put into the `stats` array so we can check for this.

src/components/tables/cells/stats/Bytes.tsx
```diff
@@ -1,23 +1,62 @@
...
+                <Tooltip
+                    title={`${defaultedVal} ${intl.formatMessage({
+                        id: read
+                            ? 'entityTable.stats.bytes_read'
+                            : 'entityTable.stats.bytes_written',
+                    })}`}
+                >
+                    <Typography
+                        sx={{
+                            transitionDelay: statsLoading ? '800ms' : '0ms',
```
travjenkins, line 47: Used to prevent the entire table from flashing quickly when a user with fast network changes the range.

williamhbaker (commented):
This looks great as far as I can tell. The queries look good, and on a skim of the code it seems like it should do what it is supposed to. Didn't review most of the code in detail though since I've forgotten pretty much all of the little I ever knew about react 😃

src/components/tables/cells/stats/Docs.tsx
```diff
@@ -1,20 +1,57 @@
...
+                            transitionDelay: statsLoading ? '800ms' : '0ms',
+                            color: (theme) =>
+                                statsLoading
+                                    ? semiTransparentBackgroundIntensified[
+                                          theme.palette.mode
+                                      ]
+                                    : null,
+                        }}
+                    >
+                        {number} docs
```
kiahna-tucker, line 51: Any reason not to make this an `intl` string with a value?

src/components/tables/Store.ts
```diff
@@ -115,51 +140,83 @@ export const getInitialState = (
...
-                        if (currentStat) {
-                            Object.entries(currentStat).forEach(
-                                ([key, value]) => {
-                                    if (typeof value === 'number') {
-                                        currentStat[key] =
-                                            currentStat[key] || 0;
-                                        currentStat[key] += datum[key];
+                } else if (data) {
+                    if (data.length > 0) {
+                        const statsData = {};
```
kiahna-tucker, line 180: Can we apply a `StatsSchema` type annotation here?

kiahna-tucker (approved):
Approved with comments.

src/components/tables/Store.ts
```diff
@@ -115,51 +140,83 @@ export const getInitialState = (
...
+                        data.forEach((datum) => {
```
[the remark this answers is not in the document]
travjenkins: It's a for loop so folks should be able to figure it out

src/components/tables/cells/stats/Docs.tsx
travjenkins: I just forgot to - I'll fix it.

src/components/tables/Store.ts
travjenkins: Added

Issue #386, opened 2022-11-21
Introduce read-only collection schema view to workflows


kiahna-tucker:
The implementation of this [proposal](https://github.com/estuary/ui/discussions/364).

PR #423, opened 2022-12-15
revision: read-only collection schema viewer to output collection form section

kiahna-tucker:
## Changes

1. Create a read-only collection schema view in the _Output Collections_ section of the create and edit forms. The right-hand pane of this form section now contains a _Config_ tab, which houses the resource configuration form for the selected collection, and a _Schema_ tab, which houses the read-only collection schema view.

1. Display only the task specification in the _Specification Editor_ section of the create and edit forms. Consequently, the _file selector_ shown on the left-hand side of the editor was removed.

1. Create a component that facilitates the integrated display of content ending with a button that opens a popper (i.e., _src/components/content/MessageWithPopper.tsx_).

1. Provide a means to view, copy, and paste the CLI commands required to edit a given collection schema in the _Schema_ tab of the _Output Collections_ section of the create and edit forms.

1. Extend support for collection schema editing via the CLI in the capture create and edit workflow. All drafted collections remain associated with the draft in light of discovery. Drafted collection schemas are propagated across iterations of drafts (i.e., the schema for a drafted collection cannot be overwritten by discovery).

## Tests

## Issues

#386 

## Content

## Screenshots


src/components/capture/useDiscoverCapture.ts
```diff
@@ -347,35 +339,90 @@ function useDiscoverCapture(
...
+                    setFormState({
+                        status: FormStatus.GENERATED,
+                    });
+                } else {
+                    // TODO (optimization): Throw generic error.
+                }
             }
         },
         [
             createDiscoversSubscription,
```
travjenkins, line 419: Not a big deal... but here we could make memo functions of the multiple paths and then maybe the list of deps could be reduced here.

src/components/editor/Bindings/Editor.tsx
```diff
@@ -10,16 +24,129 @@ interface Props {
 }
 
 function BindingsEditor({ loading, skeleton, readOnly = false }: Props) {
+    const theme = useTheme();
+    const jsonTheme =
+        theme.palette.mode === 'dark' ? 'bright' : 'bright:inverted';
```
travjenkins, line 29: Not a big deal... we probably want to end up moving this into the theme and make it a function that returns the proper setting like how we handle other dark/light mode stuff.

src/components/editor/Bindings/Editor.tsx
```diff
@@ -10,16 +24,129 @@ interface Props {
...
+                        <ResourceConfig
+                            collectionName={currentCollection}
+                            readOnly={readOnly}
+                        />
+                    ) : collectionData || collectionData === null ? (
+                        <Stack
+                            spacing={2}
+                            sx={{
+                                '& .react-json-view': {
+                                    backgroundColor: 'transparent !important',
```
travjenkins, line 71: What is driving the important tag here?

src/components/editor/Bindings/Editor.tsx
```diff
@@ -10,16 +24,129 @@ interface Props {
...
+                                                onClick={handlers.updateSchema}
+                                            >
+                                                <Refresh />
+                                            </IconButton>
+                                        ) : (
+                                            <CircularProgress
+                                                size="1.5rem"
+                                                sx={{ ml: 1 }}
+                                            />
+                                        )} */}
```
travjenkins, line 110: Can we add a TODO to document why this is commented out.

src/components/editor/Bindings/SchemaEdit/Commands/DiscoveredSchema.tsx
```diff
@@ -0,0 +1,55 @@
...
+                <FormattedMessage id="workflows.collectionSelector.schemaEdit.message1" />
+            </Typography>
+
+            <SingleLineCode formattedMessage={pullDraftCommand} />
+
+            <Typography sx={{ mt: 3, mb: 1 }}>
+                <FormattedMessage id="workflows.collectionSelector.schemaEdit.message2" />
+            </Typography>
+
+            <SingleLineCode formattedMessage={pushLocalEditsCommand} />
```
travjenkins, line 48: We discussed in person but putting here for historical purposes

This should be passed in as a prop and make this component reusable.

src/components/editor/Bindings/SchemaEdit/Commands/ExistingSchema.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const intl = useIntl();
+
+    // Draft Editor Store
+    const draftId = useEditorStore_id();
+    const persistedDraftId = useEditorStore_persistedDraftId();
+
+    // Resource Config Store
+    const currentCollection = useResourceConfig_currentCollection();
+
+    const pullDraftCommands = {
```
travjenkins, line 22: Any specific reason for an object vs an array here?

src/components/editor/Bindings/SchemaInference/Dialog.tsx
```diff
@@ -0,0 +1,415 @@
...
+            spec: draftSpecResponse.data[0].spec,
+            belongsToDraft: true,
+        });
+
+        setLoading(false);
+        setOpen(false);
+    } else {
+        setSchemaUpdateErrored(true);
+        setLoading(false);
+    }
```
travjenkins, line 121: Given this is a big handler I feel like the `setLoading` call should be outside of the branches to make sure it is always there (in case we add more branches)

src/components/editor/Bindings/SchemaInference/Dialog.tsx
```diff
@@ -0,0 +1,415 @@
...
+                                </Box>
+                            ) : null}
+                        </Stack>
+                    </Box>
+
+                    {inferredSchema ? (
+                        <DiffEditor
+                            height={`${height}px`}
+                            original={stringifyJSON(collectionData.spec)}
+                            modified={stringifyJSON(inferredSchema)}
```
travjenkins, line 330: Could these calls end up being pretty heavy? I am wondering if we should throw these into memos? I'll leave it up to you.

src/components/editor/Bindings/Selector.tsx
```diff
@@ -248,7 +248,7 @@ function BindingSelector({
                 </Stack>
             </Box>
 
-            <Box sx={{ height: 280 }}>
+            <Box sx={{ height: 480 }}>
```
travjenkins, line 251: Just thinking... we should probably end up with a system/pattern for heights the same way we do for zindex.

src/components/editor/Bindings/Store/create.ts
```diff
@@ -0,0 +1,136 @@
...
+    set: NamedSet<BindingsEditorState>,
+    get: StoreApi<BindingsEditorState>['getState']
+): BindingsEditorState => ({
+    ...getInitialStateData(),
+
+    initializeCollectionData: (currentCollection, persistedDraftId) => {
+        const { setCollectionData } = get();
+
+        if (currentCollection) {
+            evaluateCollectionData(persistedDraftId, currentCollection).then(
```
travjenkins, line 59: There is a lot of async code in this PR and got me thinking.... we need to make a call on how we handle async code and returns. I kinda feel like we might want to always return promises/async stuff from any function call so that response is there if needed.

I think we'll want to make a call on this fairly soon.

src/services/schema-inference.ts
```diff
@@ -0,0 +1,26 @@
...
+const getInferredSchema = (
+    gatewayConfig: GatewayAuthTokenResponse,
+    collection: string
+): Promise<InferSchemaResponse> => {
+    const { gateway_url, token } = gatewayConfig;
+
+    const headers: HeadersInit = {};
+
+    headers.Authorization = `Bearer ${token}`;
+    headers['Content-Type'] = 'application/json';
```
travjenkins, line 19: No change needed for this pr

This approach for API calls should really be handle in a single function given it is getting duplicated.

travjenkins (approved):
LGTM - none of the comments are worth holding this up for.

src/components/editor/Bindings/SchemaInference/Dialog.tsx
```diff
@@ -0,0 +1,442 @@
...
+                            last_pub_id
+                        );
+
+                        processDraftSpecResponse(
+                            draftSpecResponse,
+                            schemaUpdateErrored,
+                            setSchemaUpdateErrored,
+                            setCollectionData,
+                            setOpen
+                        );
```
travjenkins, line 219: This is okay but got me thinking. Is this a function where it would be better to declare variables outside the logical blocks so that the `processDraftSpecResponse` call is not duplicated.

src/components/editor/Bindings/SchemaInference/Dialog.tsx
```diff
@@ -0,0 +1,442 @@
...
+                        short
+                        title={
+                            <Typography>
+                                <FormattedMessage id="workflows.collectionSelector.schemaInference.alert.lowDocumentCount.header" />
+                            </Typography>
+                        }
+                    >
+                        <FormattedMessage id="workflows.collectionSelector.schemaInference.alert.lowDocumentCount.message" />
+                    </AlertBox>
+                ) : null}
```
travjenkins, line 291: This is fine... but I think this is a spot where we might want to just have a single error/alert. Also... could be a good spot for a stand alone component.

travjenkins (approved):

src/components/editor/Bindings/SchemaEdit/Commands/ExistingSchema.tsx
kiahna-tucker: The _hope_ was to reinforce that the order of the messages in this constant matter (e.g., the first command should always correspond to the first index, etc.). The `SingleLineCode` component seems like one that may be used in a map in the future. This was an experiment to get a sense of the most scalable way to bundle messages while communicating the importance of their order.

Arrays have a greater likelihood of being extended haphazardly; they are often prepended or appended when needed to be extended manually as opposed to programmatically. If order does matter, it is often expected that a utility function will enforce the desired order -- even in cases of manual extension. Consequently, a numerically-keyed object felt like the safest option.

src/components/editor/Bindings/Editor.tsx
kiahna-tucker: This comment was intended to be temporary as I searched for an alternative location for the loading indicator. The header of the monaco editor proved to be an apt placement (in the interim).

src/components/editor/Bindings/SchemaInference/Dialog.tsx
```diff
@@ -0,0 +1,415 @@
...
+                        );
+                    }
+                }
+            } else {
+                setSchemaUpdateErrored(true);
+            }
+        },
+    };
+
+    return currentCollection ? (
```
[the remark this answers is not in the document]
kiahna-tucker: Agreed. This is a planned, fast-follow that I am in progress of taking care of at the time of this comment.

PR #428, opened 2022-12-20
feature: Display derivation status (if any) on collections

### LINKED ISSUE #420: Display shard status for derivations (in collections)

jshearer:
## Changes
Update the "Collections" view to show shard status of any derivations present on each collection.

Only shows status if there are derivations. Otherwise we'd be showing a lot of "no shard status found" dots, which felt wrong.

I do wish this were more clear that this is _derivation_ status, and especially which derivation the status is actually representing, but according to @travjenkins that's a known limitation currently. Maybe we need mocks?

Fixes #420

## Screenshots
Screen Shot 2022-12-20 at 13 22 47


jshearer:
Based on convo with @dyaffe and note from @kiahna-tucker, I changed non-capture-collections to present a green dot that says "Capture", and I also enabled the shard failures table on collections like it is on captures and materializations.

Screen Shot 2022-12-21 at 11 32 00

jshearer:
This should be good to be reviewed! :)

travjenkins (changes requested):
Need default color when service is down

travjenkins (approved):
lgtm - tested everything before and was fine. I pushed changes to make the status clear when things don't work.

Issue #434, opened 2023-01-03
flow#860 new discover naming change breaks UI discovery flow


jshearer:
This change seems to break the UI. I suspect it's due to the new behavior with respect to how captures are named?:

<img width="1331" alt="image" src="https://user-images.githubusercontent.com/209664/209447791-060b6d11-2649-44ed-b1e6-92867feba4e5.png">

Clicking next then gives:

<img width="1342" alt="image" src="https://user-images.githubusercontent.com/209664/209447808-c5f12d3e-d023-446d-beb7-472d85efd3d0.png">

I've confirmed that the `draft_specs` table is populated with the correct specs:

```
postgres=# select catalog_name from draft_specs where  draft_id = '08:5e:ff:91:f8:02:f8:00';
      catalog_name       
-------------------------
 johnnyCo/path/foobar
 johnnyCo/path/other
 johnnyCo/path/source-db
(3 rows)
```

Note these names are different as compared to what we currently produce, which is likely (?) the problem? Specifically given a discovered capture_name of `johnnyCo/path/source-db`, my capture is named exactly that now (instead of `johnnyCo/path/source-db/source-postgres`, and my collection is `johnnyCo/path/other` instead of `johnnyCo/path/source-db/other`, etc.

_Originally posted by @jgraettinger in https://github.com/estuary/flow/issues/860#issuecomment-1364570433_

PR #436, opened 2023-01-04
fix: generalize logic relating to discovered collection names

jshearer:
Discovered collections no longer must live underneath the capture that discovered them, so logic that made that assumption needs to be updated.

I'm fairly sure this isn't the only place that needs to be updated, it's just the specific place that was causing the trouble I was able to find.

## Changes

My intent with this change was to make the logic work both for the current setup where `aliceCo/my-capture` results in collections of `aliceCo/my-capture/my-collection`, as well as the future setup where the collections will be `aliceCo/my-collection`

## Issues

#434
Related to https://github.com/estuary/flow/pull/860

## Screenshots

Screen Shot 2023-01-04 at 12 15 11
_Nested collections_

Screen Shot 2023-01-04 at 12 24 43
_Un-nested collections_

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9039,7 +9039,7 @@
         "node_modules/data-plane-gateway": {
             "version": "0.0.0",
             "resolved": "https://gitpkg.now.sh/estuary/data-plane-gateway/client/dist?main",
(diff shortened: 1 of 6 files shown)


psFried:
@jshearer I noticed in the first screenshot that the spec names don't appear to be valid. The capture name of `aliceCo/a` is a prefix of the collection names (`aliceCo/a/store`), which is not allowed. So I just wanted to ask about how it got that way, and make sure we're not somehow generating names like that automatically. Looking at #860, I'd expect that the names of the generated collections would be `aliceCo/store` (without the `/a`).

travjenkins (approved):
lgtm

PR #440, opened 2023-01-06
Dep upgrades

travjenkins:
Just upgrading a bunch of deps to latest versions that work.

Changes caused by updates:

1. LogRocket filtering calls new function
2. JSON Forms typing that came up with newer typescript 
3. Resource Config - typescript threw an error with old approach


kiahna-tucker (approved):
Approved. Once this branch is synchronized with `main`, it is set to be merged.

Issue #448, opened 2023-01-16
Remove scary warning screen from "Refresh" button on Captures


dyaffe:
**Bug**
We're now doing deep merges when someone clicks refresh in captures, so we should remove the big scary warning sign.

**Screenshots**

![Screenshot 2023-01-16 at 12 12 41 PM](https://user-images.githubusercontent.com/8095717/212734005-a4d709d3-3927-4d5b-b1e7-293631dfbb85.png)

PR #450, opened 2023-01-17
Remove capture rediscovery confirmation modal

kiahna-tucker:
## Changes

Remove the confirmation modal that previously appeared when the _Refresh_ CTA in the _Output Collections_ section of capture workflow was clicked.

## Tests

Only manual testing was performed, confirming that the confirmation modal does not appear when the aforementioned _Refresh_ CTA is clicked.

## Issues

#448


travjenkins (approved):
lgtm - approving with a quick review and no manual testing

PR #469, opened 2023-01-31
Promote the use of existing tasks in create workflow

kiahna-tucker:
## Changes

## Tests

## Issues

## Content

## Screenshots


oliviamiannone (approved):
The messaging looks good, but I do have some suggested tweaks in order to...

* connect the dots for users who may not yet know that you can add binding by editing the capture/materialization (Using "Adding to")
* Make verbiage a touch less formal. 

"Adding to an existing capture/materialization can help save time and cut cost. Choose a capture/materialization from the list to edit it. If you’d still prefer to start from scratch, select the <> option below."

src/api/liveSpecsExt.ts
```diff
@@ -47,13 +58,17 @@ export interface CollectionQueryWithStats extends CollectionQuery {
...
-    .concat(['connector_image_name', 'writes_to'])
+const captureColumns = commonColumns.concat(['writes_to']).join(',');
+const captureColumnsWithSpec = commonColumns
+    .concat(['writes_to', 'spec'])
     .join(',');
 
 const materializationsColumns = commonColumns.concat(['reads_from']).join(',');
+const materializationsColumnsWithSpec = commonColumns
+    .concat(['reads_from', 'spec'])
+    .join(',');
```
travjenkins, line 69: Same as above.

src/components/capture/Create/index.tsx
```diff
@@ -142,67 +143,69 @@ function CaptureCreate() {
...
-                        <RediscoverButton
-                            entityType={entityType}
-                            disabled={!hasConnectors}
-                            callFailed={helpers.callFailed}
-                            postGenerateMutate={mutateDraftSpecs}
-                        />
-                    }
-                />
-            </ResourceConfigHydrator>
+            <EndpointConfigHydrator>
```
travjenkins, line 151: This compare is rough... I'm assuming it was just the nesting that changed and everything else is kept the same.

src/components/shared/CustomWidthTooltip.tsx
```diff
@@ -0,0 +1,14 @@
+import { styled, Tooltip, tooltipClasses, TooltipProps } from '@mui/material';
+
+const CustomWidthTooltip = styled(({ className, ...props }: TooltipProps) => (
```
travjenkins, line 3: Not for this PR

This is interesting... I never really thought about moving the `styled` stuff into stand alone components as they're almost always just needed in one place.

We should take some time and move these to their own folders so it is easy to see which MUI components we are deeply customizing.

src/components/shared/Entity/Create/Config.tsx
```diff
@@ -0,0 +1,114 @@
...
+    title: string;
+    entityType: EntityWithCreateWorkflow;
+}
+
+function EntityCreateConfig({ title, entityType }: Props) {
+    useBrowserTitle(title);
+
+    const [connectorId] = useGlobalSearchParams([
+        GlobalSearchParams.CONNECTOR_ID,
+    ]);
```
travjenkins, line 28: This does not have to be an array - the function will take a single param and return a single item.

src/components/shared/Entity/Create/Config.tsx
```diff
@@ -0,0 +1,114 @@
...
+        boolean | null
+    >(null);
+
+    useEffect(() => {
+        if (typeof connectorId === 'string') {
+            setShowConnectorTiles(false);
+        } else {
+            setShowConnectorTiles(true);
+        }
+    }, [connectorId]);
```
travjenkins, line 47: I am confused on why this was done this way as it seems way over engineered. Can you put in a comment on why this was done.

src/components/shared/Entity/Create/index.tsx
```diff
@@ -161,78 +141,67 @@ function EntityCreate({
...
-                            </AlertBox>
-                        ) : connectorTags.length > 0 ? (
-                            <ErrorBoundryWrapper>
-                                <DetailsForm
-                                    connectorTags={connectorTags}
-                                    accessGrants={combinedGrants}
-                                    entityType={entityType}
-                                />
-                            </ErrorBoundryWrapper>
+            {connectorTagsError ? (
```
travjenkins, line 150: Assuming this chunk is like the ones above where changing the indentation makes the compare rough. Everything looks identical but did not deeply review each line.

src/components/shared/Entity/ExistingEntityCards/Cards/Existing.tsx
```diff
@@ -0,0 +1,175 @@
...
+    logEvent: CustomEvents,
+    data: CaptureQueryWithSpec | MaterializationQueryWithSpec
+) => {
+    LogRocket.track(logEvent, {
+        id: data.id,
+        connector_tag_id: data.connector_id,
+    });
+};
+
+function ExistingEntityCard({ queryData }: Props) {
```
travjenkins, line 49: This is okay... but this probably should not have been architected like this. I think it would have been better to make a more common horizontal/long/whatever card and then built off that for ExistingEntity. But not big deal.

src/components/shared/Entity/ExistingEntityCards/Cards/Existing.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    const [detailsExpanded, setDetailsExpanded] = useState<boolean>(false);
+
+    const handlers = {
+        editTask: () => {
+            if (!isEmpty(queryData)) {
+                const logEvent =
+                    queryData.spec_type === 'capture'
+                        ? CustomEvents.CAPTURE_CREATE_CONFIG_EDIT
+                        : CustomEvents.MATERIALIZATION_CREATE_CONFIG_EDIT;
```
travjenkins, line 63: I feel this logic should live in the `trackEvent` function so it is self contained.

src/components/shared/Entity/ExistingEntityCards/Cards/Existing.tsx
```diff
@@ -0,0 +1,175 @@
...
+                        [GlobalSearchParams.LIVE_SPEC_ID]: queryData.id,
+                        [GlobalSearchParams.LAST_PUB_ID]: queryData.last_pub_id,
+                    })
+                );
+            }
+        },
+        toggleDetailsPanel: () => setDetailsExpanded(!detailsExpanded),
+    };
+
+    return queryData ? (
```
travjenkins, line 85: This should be handled with a negative check above and return `null`

src/components/shared/Entity/ExistingEntityCards/Cards/New.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    // Existing Entity Store
+    const connectorName = useExistingEntity_connectorName();
+
+    const createNewTask = () => {
+        if (entityType === 'capture' || entityType === 'materialization') {
+            const logEvent =
+                entityType === 'capture'
+                    ? CustomEvents.CAPTURE_CREATE_CONFIG_CREATE
+                    : CustomEvents.MATERIALIZATION_CREATE_CONFIG_CREATE;
```
travjenkins, line 40: Same as above

src/components/shared/Entity/ExistingEntityCards/Store/Hydrator.tsx
```diff
@@ -0,0 +1,53 @@
...
+    }, [
+        hydrateState,
+        setHydrated,
+        setHydrationErrorsExist,
+        connectorId,
+        entityType,
+        hydrated,
+    ]);
+
+    return <div>{children}</div>;
```
travjenkins, line 50: We don't need the `div` here. We can use the empty tags and throw an eslint comment

src/components/shared/Entity/ExistingEntityCards/Store/create.ts
```diff
@@ -0,0 +1,100 @@
...
+                const { setCreateNewTask, setHydrationErrorsExist } = get();
+
+                const { data, error } = await getLiveSpecsByConnectorId(
+                    entityType,
+                    connectorId
+                );
+
+                if (error) {
+                    setHydrationErrorsExist(true);
+                    setCreateNewTask(true);
```
travjenkins, line 68: We should `return` here

src/components/shared/Entity/ExistingEntityCards/Toolbar.tsx
```diff
@@ -0,0 +1,169 @@
...
+                isFiltering.current = hasQuery;
+
+                setSearchQuery(hasQuery ? filterQuery : null);
+            },
+            750
+        ),
+    };
+
+    return (
+        <Toolbar
```
travjenkins, line 105: This repeated stuff taken from `ConnectorToolbar` should have been pulled out and shared between these two.

src/components/shared/Entity/ExistingEntityCards/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+    }, [connectorId, entityType, searchQuery, sortDirection]);
+
+    const { data: useSelectResponse, isValidating } = useDistributedSelectNew<
+        CaptureQueryWithSpec | MaterializationQueryWithSpec
+    >(liveSpecQuery);
+
+    const selectData = useMemo(
+        () => (useSelectResponse ? useSelectResponse.data : []),
+        [useSelectResponse]
+    );
```
travjenkins, line 96: What is driving memoizing `liveSpecQuery` and `selectData` ?

src/stores/EndpointConfig/Hydrator.tsx
```diff
@@ -0,0 +1,44 @@
...
+                    setHydrated(true);
+                    setHydrationErrorsExist(true);
+
+                    LogRocket.log('Failed to hydrate endpoint config', error);
+                }
+            );
+        }
+    });
+
+    return <div>{children}</div>;
```
travjenkins, line 43: Same as above.

src/stores/names.ts
```diff
@@ -17,6 +17,10 @@ export enum EndpointConfigStoreNames {
     GENERAL = 'general-endpoint-config',
 }
 
+export enum ExistingEntityStoreNames {
+    GENERAL = 'general_existing_entity',
```
travjenkins, line 21: this needs to use hyphens in the name.

src/components/shared/Entity/Create/Config.tsx
kiahna-tucker: You authored this block of code. It originally lived in the shared `EntityCreate` component and was copied and pasted over when the component was spliced. Agreed, that it can be simplified.

src/components/capture/Create/index.tsx
kiahna-tucker: Yes, that is it. The addition of the endpoint config store hydrator altered the nesting of all components. No changes were made to them.

src/components/shared/Entity/Create/Config.tsx
kiahna-tucker: Yes, I know that. This code was shuffled quite a bit over the lifetime of this branch and an additional search parameter was extracted in legacy code. Will cleanup.

src/components/shared/Entity/ExistingEntityCards/Cards/Existing.tsx
kiahna-tucker: Originally, the existing entity cards were not innately buttons and had an assortment of functions built into them (e.g., specification and endpoint config preview, options menu that surfaced the edit option and an advanced specification comparer, a bespoke _Edit_ CTA when the options menu was vetoed, etc.).

Given the existing entity cards share so much in common with the new entity card, I agree that a base should be defined.

src/components/shared/Entity/ExistingEntityCards/Store/Hydrator.tsx
kiahna-tucker: Will do. Every developer has a unique threshold for explicit, linting overrides that are deemed _acceptable_, so I tend to err on the side of caution when working in collaborative codebases.

src/components/shared/Entity/ExistingEntityCards/Store/create.ts
kiahna-tucker: I feel combining the this if statement with the if/else statement below would make more sense than throwing in a return statement.

```ts
if (error) {
    ...
} else if (data && data.length > 0) {
    ...
} else {
    ...
}
```

travjenkins (approved):
lgtm

src/components/connectors/ConnectorTiles.tsx
```diff
@@ -150,8 +150,9 @@ function ConnectorTiles({
     const primaryCtaClick = (row: ConnectorWithTagDetailQuery) => {
         navigateToCreate(
             row.connector_tags[0].protocol,
-            row.connector_tags[0].id,
-            replaceOnNavigate
+            row.connector_tags[0].connector_id,
+            replaceOnNavigate,
+            row.connector_tags[0].protocol === 'capture'
```
kiahna-tucker, line 155: The `useEntityCreateNavigate` hook's fourth input parameter dictates whether the create form path is selected as the base for the next URL. At this point in time, the existing entity check must be hidden for captures. Clicking the _Capture_ CTA on a connector tile/card should always advance the user directly to the create form.

src/components/shared/Entity/DetailsForm/Form.tsx
```diff
@@ -194,7 +196,8 @@ function DetailsFormForm({
 
                 navigateToCreate(
                     entityType,
-                    details.data.connectorImage.id,
+                    details.data.connectorImage.connectorId,
+                    true,
                     true
```
kiahna-tucker, line 201: The `useEntityCreateNavigate` hook's fourth input parameter dictates whether the create form path is selected as the base for the next URL. Since the callback function generated by this hook is called every time a change is made to the details form, it is integral that this parameter is set to `true`. Otherwise, the router will attempt to navigate to the existing entity check page if the original connector is reselected from the list. It should also be noted that this results in jarring, partial renderings of the create form if a different connector is selected from the list.

src/services/supabase.ts
```diff
@@ -132,6 +133,47 @@ export const defaultTableFilter = <Data>(
     return queryBuilder;
 };
 
+export const distributedTableFilter = <Data>(
```
kiahna-tucker, line 136: This is a direct copy of `defaultTableFilter` above, with the exception of the distributed type annotation applied to the `query` input parameter. In a fast follow, we should look at consolidating the two functions.

Issue #468, opened 2023-01-29
Make it Easier to Select Like Collections during Materialization


dyaffe:
**Problem**
Currently when you are selecting multiple collections to be materialized, what you typed in clears after selecting a collection.  This is annoying because usually you want several collections with the same prefix. 

It would be nice if you could select multiple collections after typing in a prefix without it clearing and then clear it manually.

PR #471, opened 2023-02-01
(UX Tweaks 7) Do not clear collection search input when option selected

travjenkins:
## Changes

1. After a user has typed something into the collection finder and selects an options the value they typed in will stay.
2. Put a max amount of chips to display in the table

## Tests

Manually tested

## Issues

Fixes https://github.com/estuary/ui/issues/468

## Content

None

## Screenshots

Typing in search
image

Selecting a result (input stays populated)
image

Selected all options (input stays populated)
image

Hiding long lists of elements
image


src/components/collection/Picker.tsx
```diff
@@ -59,6 +62,7 @@ function CollectionPicker({ readOnly = false }: Props) {
         CollectionData[]
     >([]);
     const [missingInput, setMissingInput] = useState(false);
+    const [inputValue, setInputValue] = useState('');
```
travjenkins, line 65: Need state so we can manage the input value

src/components/collection/Picker.tsx
```diff
@@ -226,10 +230,19 @@ function CollectionPicker({ readOnly = false }: Props) {
                 getOptionLabel={(option) => option.name}
                 isOptionEqualToValue={(option, value) => isEqual(option, value)}
                 value={collectionValues}
+                inputValue={inputValue}
```
travjenkins, line 233: Populate the input value

src/components/collection/Picker.tsx
```diff
@@ -226,10 +230,19 @@ function CollectionPicker({ readOnly = false }: Props) {
...
                 fullWidth
                 onChange={handlers.updateCollections}
+                onInputChange={(_event, newInputValue, reason) => {
+                    const inputBeingReset =
+                        detectAutoCompleteInputReset(reason);
+
+                    if (!inputBeingReset) {
+                        setInputValue(newInputValue);
+                    }
+                }}
```
travjenkins, line 245: We will normally update the input value except when it is being reset.

kiahna-tucker (approved):
Approved without manual testing. I am not the biggest fan of the appearance of the _show more_ option in the chip list; it does not isn't obviously clickable since it is styled in a similar fashion as the non-interactive chips. Not a blocking point of contention.

Issue #467, opened 2023-01-27
Cancel Button after Discover


dyaffe:
**Problem**
We don't currently have a cancel button after someone issues a discover.  The issue is that if you get something wrong and it hangs forever, you lose everything you just input. 

**Proposal**
Even though we don't support a cancellation of operations on the back end, we should add a cancel button which allows the user to try again.  This would be a _much better_ UX than the current version.

travjenkins:
@psFried 
Do we know why discovers are hanging forever? This feels like an enhancement that could end up biting us in the butt. Do we have an idea of what it would take to enable cancellation on the backend?

dyaffe:
@travjenkins we do -- this is what happens when a user attempts to capture with the wrong credentials.  @psFried and I discussed this and decided that a frontend cancel was the right thing for now.

I believe we do have a timeout and the backend stopped listening to the request but somehow the frontend didn't know.

travjenkins:
@dyaffe  Do we have the session for this happening (or discover id or anything really)? The "frontend didn't know" is not sitting well with me. The UI pings the server constantly to fetch the most recent status so we need to know why the UI was not aware the discovery ended.

dyaffe:
It's super easy to reproduce.  Just use SSH tunneling and get the password wrong.  I personally had it happen on Friday when I was setting this up and we saw a user do it on Thursday during office hours @travjenkins

travjenkins:
This is actually a bug with running discover a second time. Looking into it.

dyaffe:
One quick comment here -- I never received a successful discovery response here so I don't think this was the cause, though it does seem like something to address @travjenkins

PR #472, opened 2023-02-02
Allow discover to run multiple times if it fails

travjenkins:
## Changes

1. We used to always disable discover after the first time the user used the next button. This meant that if a user submitted a discover that failed (like wrong credentials) they could never run another discover with the proper credentials. Now we only turn off discover if the status of the form is `generated`

## Tests

Manually tested:

1. Create with discover working the first time
2. Create with discover failing the first time and working the second time
3. Edit with discover working the first time

## Issues

Fixes: https://github.com/estuary/ui/issues/467

## Content

N/A

## Screenshots

N/A


kiahna-tucker (approved):

PR #473, opened 2023-02-03
Add a GTM register event

travjenkins:
## Changes

Added in a call to log when someone registers a new tenant

## Tests

Manually tested with the Tag Assistant Companion 

## Issues / Content / Screenshots
N/A

### DIFF
--- a/src/directives/BetaOnboard.tsx
+++ b/src/directives/BetaOnboard.tsx
@@ -6,6 +6,7 @@ import AlertBox from 'components/shared/AlertBox';
 import ExternalLink from 'components/shared/ExternalLink';
 import { useState } from 'react';
 import { FormattedMessage, useIntl } from 'react-intl';
+import { fireGtmEvent } from 'services/gtm';
 import { jobStatusPoller } from 'services/supabase';
 import { hasLength, PREFIX_NAME_PATTERN } from 'utils/misc-utils';
 import { jobStatusQuery, trackEvent } from './shared';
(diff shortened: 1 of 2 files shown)


src/directives/BetaOnboard.tsx
```diff
@@ -60,6 +61,9 @@ const BetaOnboard = ({ directive, mutate }: DirectiveProps) => {
                 jobStatusPoller(
                     jobStatusQuery(data),
                     async () => {
+                        fireGtmEvent('Register', {
+                            tenant: requestedTenant,
+                        });
```
travjenkins, line 66: Fire the GTM event once the tenant is registered and we know it was accepted.

src/services/gtm.ts
```diff
@@ -1,8 +1,11 @@
 import TagManager from 'react-gtm-module';
+import { Schema } from 'types';
 import { getGoogleTageManagerSettings, isProduction } from 'utils/env-utils';
 
 const googleTagManagerSettings = getGoogleTageManagerSettings();
 
+type EVENTS = 'Register';
```
travjenkins, line 7: Keeping a list of allowed events

src/services/gtm.ts
```diff
@@ -12,3 +15,18 @@ export const initGoogleTagManager = () => {
         TagManager.initialize({ gtmId: googleTagManagerSettings.id });
     }
 };
+
+export const fireGtmEvent = (event: EVENTS, data: Schema | undefined = {}) => {
+    if (
+        isProduction &&
+        googleTagManagerSettings.enabled &&
+        googleTagManagerSettings.id
+    ) {
```
travjenkins, line 24: Duplicating check for ease of approach. Needs to later be shared with the function above

src/services/gtm.ts
```diff
@@ -12,3 +15,18 @@ export const initGoogleTagManager = () => {
...
+        TagManager.dataLayer({
+            dataLayer: {
+                ...data,
+                event,
+            },
+        });
```
travjenkins, line 30: Put the spread at the top so the event property could not be overwritten.

kiahna-tucker (approved):

PR #483, opened 2023-02-16
Display create new task option first on existing entity check page

kiahna-tucker:
## Changes

On the existing entity check page, the create new task option will be the first card in the list to prevent it from being buried by moderate to long lists of existing entities.

## Tests

Confirmed that the create new task option is the first card displayed on the existing entity check page.

## Screenshots

reverse_existing_entity_check_card_order


travjenkins (approved):
lgtm

Issue #481, opened 2023-02-14
Support for Connector-Based Schema Inference


dyaffe:
**Background**
The backend now supports annotations for connector-based schema inference.  We should plumb those through in the UI.

CC @travjenkins since we just spoke about creating this issue

kiahna-tucker:
The schema inference annotation to look for: `x-infer-schema`.

travjenkins:
@mdibaiee 

I see this is done: https://github.com/estuary/connectors/issues/499

However - I am not seeing this annotation in the schemas out on Production. I looked in Kafka, Kinesis, and GCS. I could easily be missing something... have these changes been deployed out yet?

mdibaiee:
@travjenkins they are available, they are part of the schema of the collection, not the connector. See the example below for Firestore:
<img width="1246" alt="image" src="https://user-images.githubusercontent.com/2807772/219716417-76c06576-6439-4901-a542-13a8af445890.png">

travjenkins:
@mdibaiee ahhh that makes much more sense. Thanks man!

PR #487, opened 2023-02-17
UX Tweaks: bindings selector

kiahna-tucker:
## Changes

1. Default the resource config of collections added to a materialization via a materialize CTA according to the collection name annotation.

1. Display the schema inference CTA in the _Output Collections_ form section when a collection schema contains the `x-infer-schema` annotation.

## Tests

The following tests were conducted:

* Create a capture using a _schema-inference enabled_ connector and confirm that the schema inference CTA remains hidden in the capture create workflow.

* Edit a capture that **only** has _schema-inference enabled_, bound collections and confirm that the schema inference CTA appears.

* Edit a capture that **only** has _schema-inference disabled_, bound collections and confirm that the schema inference CTA remains hidden.

* Edit a capture that has a mix of bound collections and confirm that the schema inference CTA appears **only** when a _schema-inference enabled_, bound collection is selected.

**NOTE:** The final three tests were repeated for the materialization create and edit workflows.

## Issues

#481

### Default Resource Config Field of Prefilled Collections

**Materialization Edit | Original Task Bindings**

pr_screenshot-487-bindings_selector_ux_tweaks-prefill_collection-default

<br />

**Materialization Edit | Prefilled Resource Config**

pr_screenshot-487-bindings_selector_ux_tweaks-prefill_collection-redux_dev_tools

### Schema Inference CTA

**Capture Edit | Schema Inference Enabled**

pr_screenshot-487-bindings_selector_ux_tweaks-infer-schema-enabled

<br />

**Capture Edit | Schema Inference Disabled**

pr_screenshot-487-bindings_selector_ux_tweaks-infer-schema-disabled

<br />


src/components/editor/Bindings/Store/create.ts
```diff
@@ -98,12 +101,33 @@ const getInitialState = (
...
             evaluateCollectionData(persistedDraftId, currentCollection).then(
-                (response) => setCollectionData(response),
-                () => setCollectionData(undefined)
+                (response) => {
+                    setCollectionData(response);
+
+                    if (response) {
+                        const writeSchemaKey = response.spec.hasOwnProperty(
+                            'writeSchema'
+                        )
```
travjenkins, line 114: Thinking out loud
We are doing this check in two places. We should make a `schemaHlper` or something soon.

src/components/editor/Bindings/Store/create.ts
```diff
@@ -98,12 +101,33 @@ const getInitialState = (
...
+                            ? 'writeSchema'
+                            : 'schema';
+
+                        const inferenceAnnotationExists = !response.spec[
+                            writeSchemaKey
+                        ].hasOwnProperty(Annotations.inferSchema);
+
+                        setSchemaInferenceDisabled(inferenceAnnotationExists);
```
travjenkins, line 122: Why are we checking if the property does not exist and then store it in a variable stating it does exist? Also - what is driving not checking the actual value of the prop and only the exists of the key?

src/services/ajv.ts
```diff
@@ -49,22 +52,58 @@ export const setDefaultsValidator = (function () {
...
 function setJSONFormDefaults(jsonSchema: any, formData: any) {
     const hydrateAndValidate = setDefaultsValidator.compile(jsonSchema);
+
     hydrateAndValidate(formData);
+
     return hydrateAndValidate;
 }
 
-export function createJSONFormDefaults(jsonSchema: any): {
+function defaultResourceSchema(resourceSchema: any, collection: string) {
```
travjenkins, line 61: If this is being handled in a store then I think we're good with removing the similar logic in the resource config form.

src/components/editor/Bindings/Store/create.ts
kiahna-tucker: Q1: Typo. If you look at the commit history, the property used to be named `schemaInferenceEnabled`.

Q2: That can be done. In practice, the annotation is applied in a similar fashion to an optional argument that can only have the value of `true`. Will record the value explicitly _just in case_ a change in annotation application occurs.

src/services/ajv.ts
kiahna-tucker: No, that logic cannot be removed as it stands today. Additional logic would need to be added to the store to handle defaulting bindings added to the list post-hydration. Could be another fast follow but I would advocate for it be considered out of scope for this PR. This PR adds a needed UX improvement.

src/services/ajv.ts
```diff
@@ -70,10 +71,18 @@ function defaultResourceSchema(resourceSchema: any, collection: string) {
 
     if (collectionNameFieldKey) {
         // Add a default property set to the stripped collection name
-        const modifiedSchema = resourceSchema;
-
-        modifiedSchema.properties[collectionNameFieldKey].default =
-            stripPathing(collection);
+        let modifiedSchema = {};
```
travjenkins, line 74: What is driving the use of `let` here?

src/services/ajv.ts
travjenkins: Ahhh- yup... that's why it is still here :)

Yeah do not do it in this PR for sure - can be a nice little win later on.

src/services/ajv.ts
kiahna-tucker: Working through the kinks of the bug you noted, I reassigned this variable. Agreed it can become a constant now.

src/components/editor/Bindings/Store/create.ts
kiahna-tucker: Should be resolved.

travjenkins (approved):
lgtm

PR #490, opened 2023-02-21
Turn off Magic Link and make mobile login look better

travjenkins:
## Changes

Disable magic link
On the login page - set the smallest window width smaller to allow better mobile login

## Tests

Manually tested

## Issues
N/A

## Content
N/A

## Screenshots
N/A


.env.development.local
```diff
@@ -6,6 +6,7 @@ REACT_APP_SUPABASE_URL=http://localhost:5431
 REACT_APP_SUPABASE_ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24ifQ.625_WdcF3KHqz5amU0x2X5WWHP-OEs_4qj0ssLNHzTs
 REACT_APP_GATEWAY_AUTH_TOKEN_URL=http://localhost:5431/rest/v1/rpc/gateway_auth_token
 
+REACT_APP_SHOW_EMAIL_LOGIN=true
```
travjenkins, line 9: We need to keep this on for local dev. Makes setting up locals easier and is also the best option to login when testing locally through a image.

src/context/Theme.tsx
```diff
@@ -368,9 +368,12 @@ const themeSettings = createTheme({
         },
         MuiCssBaseline: {
             styleOverrides: {
-                body: {
+                'body': {
                     minWidth: sm,
                 },
+                'body.loginPage': {
+                    minWidth: xs,
+                },
```
travjenkins, line 376: Eventually we can remove this extra code but until we do a full "mobile audit" I think the smallest impact to the application will be only doing this on the login page.

Issue #485, opened 2023-02-16
Retain and indicate selected collections in binding options list


dyaffe:
**Problem**
When selecting collections in the materialization creation UI, it isn't currently possible to see your collections that have been selected because the drop down covers up the selected collections.

I'm not sure what MUI supports, but it would be nice to have the following features in that selector:

1. If you search for something, your search stays after selecting so you can select more with the same prefix (this works now).
2. You can see all of the things you've selected so far

This was feedback from a user.

travjenkins:
This work? 
![image](https://user-images.githubusercontent.com/270078/219558001-a6c8dbef-819a-436b-bbdc-eaaff49ba09b.png)

dyaffe:
Yeah, I think that's great.

Issue #475, opened 2023-02-06
Make Using Date Pickers More Obvious


dyaffe:
**Background**
Our date pickers only open if you select them with the calendar icon which is small and to the right of a very large form.  It would be nice if they opened when you selected anywhere in the box.  

We should expect manual entry to be something that no one really wants to do, so doing that should be harder & not the standard route.

Issue #474, opened 2023-02-04
Timezone discrepancy when fetching stats


williamhbaker:
**Bug**
Stats queries for the daily grain (and possibly others) appear to not be using the correct timezone conversions.

Example:
After 4 PM pacific time on Feb 3, which is 12 AM GMT Feb 4, the UI continues to query daily grains for Feb 3. The stats table starts getting updates for the daily grain in accordance with GMT though, so its the Feb 4th grain that starts getting updated.

**Expected**
See stats update in the UI after 4 PM pacific. But there's a dead zone where it only queries the stale Feb 3 stats.

**Screenshots**
Include a screenshot of what you saw when the bug happened.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**
I assume that at 12 AM pacific it would start querying for "the next day" but I haven't confirmed that directly.

williamhbaker:
This is still happening: Was looking at stats at 4 PM pacific and noticed that they were showing in the UI as the stats from the previous day. The stats are updating correctly in the supabase DB but the UI is still querying for the previous day.

travjenkins:
Oh I see what the issue is. We are fetching `daily` by using the helper `startofToday` and not with the current time for the user. Should be a simple fix.

travjenkins:
Testing with my local time changed to the end of the eastern day and looks to line up now
![image](https://user-images.githubusercontent.com/270078/225088836-ef51a088-e098-4be2-8e02-8e7b4ef3e4d2.png)

Reference
![image](https://user-images.githubusercontent.com/270078/225089097-9edbe219-e723-49b8-a64f-2b1dd9569b0d.png)

travjenkins:
The rest of the times

![image](https://user-images.githubusercontent.com/270078/225089859-60b308b2-48c7-4218-8141-7c0de799c513.png)

PR #464, opened 2023-01-25
Adding page for details of entity

travjenkins:
## Changes

1. Updated React Router to latest to start using their new data router
1. Added a details page for all entity types
     - Made entity name in tables the link to the details page
     - Changed table action menu to just be the edit button again
     - Opened the shard error accordion by default
5. Added a search query management context for table filters
     - Did not switch over all the other uses of search query yet
6. Updated stats so that they fetch the time in GMT
7. Open date pickers automatically when the input is entered or the button is active
     - This does impact accessibility as the focus does not switch to the picker. Noted in comments.
8. Mark the selected collections in the bindings selector.
9. Started messing with React Charts
     - Added in some code to support a future "history" page. The component is removed to reduce size of bundle

## Tests

Manually

## Issues

Fixes: https://github.com/estuary/ui/issues/438
Fixes: https://github.com/estuary/ui/issues/474
Fixes: https://github.com/estuary/ui/issues/475
Fixes: https://github.com/estuary/ui/issues/485

## Content

There is a little bit of new content

### Captures
image

### Collections
image

### Materializations
image

Faked multiple shard errors
image

### Bindings selector
We'll now show which items are selected
image

PR #492, opened 2023-02-21
UX Tweaks: Binding selecting, stats, and date time picker

travjenkins:
## Changes

These changes were pulled from the old branch that handle details page

485: Adding a renderer that shows a check box next to items selected
475: Open the date/time/date-time picker open calendar by default
474: Changed how we fetch the stats to use the proper GMT timezone
Tweak: Update the icon color in alert box as Icon Noir's much thinner lines made them hard to see when colored 

## Tests

Manually tested

## Issues

Fixes https://github.com/estuary/ui/issues/485
Fixes https://github.com/estuary/ui/issues/475
Fixes https://github.com/estuary/ui/issues/474

## Content

N/A

## Screenshots

See https://github.com/estuary/ui/pull/464 for details

image

### DIFF
--- a/src/components/collection/Picker.tsx
+++ b/src/components/collection/Picker.tsx
@@ -3,12 +3,15 @@ import {
     AutocompleteChangeReason,
     Box,
     TextField,
+    Typography,
 } from '@mui/material';
 import { useEditorStore_persistedDraftId } from 'components/editor/Store/hooks';
 import { useEntityType } from 'context/EntityContext';
+import { truncateTextSx } from 'context/Theme';
 import { useEntityWorkflow } from 'context/Workflow';
 import useDraftSpecs from 'hooks/useDraftSpecs';
 import useLiveSpecs from 'hooks/useLiveSpecs';
+import { Check } from 'iconoir-react';
 import { isEqual } from 'lodash';
 import { useEffect, useMemo, useState } from 'react';
 import { useIntl } from 'react-intl';
@@ -232,7 +235,6 @@ function CollectionPicker({ readOnly = false }: Props) {
                 value={collectionValues}
                 inputValue={inputValue}
                 size="small"
-                filterSelectedOptions
                 fullWidth
                 onChange={handlers.updateCollections}
                 onInputChange={(_event, newInputValue, reason) => {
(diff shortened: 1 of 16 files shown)


kiahna-tucker:
Would you mind introducing some space between an alert and a form in this PR?

pr_review-492-form_error_alert

src/components/collection/Picker.tsx
```diff
@@ -257,6 +259,34 @@ function CollectionPicker({ readOnly = false }: Props) {
...
+                renderOption={(props, option, { selected }) => {
+                    return (
+                        <li {...props}>
+                            <Box
+                                sx={{
+                                    ml: -2,
+                                    mr: 0.5,
+                                }}
+                            >
+                                <Check
```
kiahna-tucker, line 271: What color is this icon? My browser dev tools are being a pain, but it _looks_ like the it is white in dark mode and black in light mode as opposed to being the primary text color set in the theme.

src/components/collection/Picker.tsx
```diff
@@ -257,6 +259,34 @@ function CollectionPicker({ readOnly = false }: Props) {
                         onBlur={handlers.validateSelection}
                     />
                 )}
+                renderOption={(props, option, { selected }) => {
+                    return (
+                        <li {...props}>
```
kiahna-tucker, line 264: Two visual oddities about this list:

* The hover style of a list item lingers on the last item that was hovered.

* The checkmark appears to be uncomfortably close to the left edge of the popper in **light mode only**.

Unsure if the former occurs in production (at the time of commenting), but calling both out for awareness.


### Reference

**Dark Mode**

pr_review-492-available_bindings-dark

<br />

**Light Mode**

pr_review-492-available_bindings-light

src/components/editor/Bindings/Selector.tsx
```diff
@@ -176,13 +176,15 @@ function BindingSelector({
...
                         <>
-                            <WarningCircle
-                                style={{
-                                    marginRight: 4,
-                                    fontSize: 12,
-                                    color: theme.palette.error.main,
-                                }}
-                            />
+                            <Box>
+                                <WarningCircle
```
kiahna-tucker, line 180: Curious if a defined min width on the icon is an alternative to prevent shrinkage here.

kiahna-tucker (approved):
Approved with comments.

src/components/collection/Picker.tsx
travjenkins: The hover style does not linger for me. What browser are you testing on?

src/components/collection/Picker.tsx
travjenkins: They do match. Forced autocomplete open and they are set to the same color

src/components/collection/Picker.tsx
travjenkins: Solved the icons shifting with a hack/force of the padding left.

Not sure about the hover styling staying there. If we notice it again and can dig into the cause we can fix it later on.

src/components/editor/Bindings/Selector.tsx
travjenkins: Potentially... however I think we'll need to decide if we want to use `style` for icons or include a wrapper that allows us to style with `sx`

Issue #438, opened 2023-01-06
Add a larger details view for entities 


travjenkins:
Need to add support for all the entities having their own "details view" instead of just expanding a table row.

PR #500, opened 2023-02-28
Open connector help docs in a side panel

travjenkins:
## Changes

Add a side panel to open docs

## Tests

Manually tested

## Issues

Fixes: https://github.com/estuary/ui/issues/438

## Content

Have some new messages mainly for displaying a warning when 3rd party docs cannot be opened inline.

## Screenshots

In line docs open
image

When closed there is a button to reopen
image

If small screen we do nothing new and just show the old button
image

If the docs are 3rd party we close the panel and offer open link
image


.env
```diff
@@ -40,4 +40,8 @@ REACT_APP_LOGROCKET_ID_USER_INCLUDE_EMAIL=true
 
 # Analytics settings
 REACT_APP_GOOGLE_TAG_MANAGER_ENABLED=true
-REACT_APP_GOOGLE_TAG_MANAGER_ID=GTM-WK8SB2L
\ No newline at end of file
+REACT_APP_GOOGLE_TAG_MANAGER_ID=GTM-WK8SB2L
+
+# Used to post message to docs for Dark/Light mode toggle
+REACT_APP_DOCS_ORIGIN=https://docs.estuary.dev
+REACT_APP_DOCS_IFRAME_STRING_INCLUDE=estuary.dev
```
travjenkins, line 47: New props for handling inline docs

src/app/Layout.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { Outlet } from 'react-router';
+import { useLocalStorage } from 'react-use';
+import {
+    useSidePanelDocsStore_animateOpening,
+    useSidePanelDocsStore_setAnimateOpening,
+    useSidePanelDocsStore_show,
+} from 'stores/SidePanelDocs/hooks';
+import { LocalStorageKeys } from 'utils/localStorage-utils';
+
+function AppLayout() {
```
travjenkins, line 16: Moved layout over to a new file

src/app/Layout.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    const toggleNavigationDrawer = () => {
+        setNavigationConfig({ open: !navigationOpen });
+    };
+
+    // Splitter for the side panel docs
+    const [leftPaneFlex, setLeftPaneFlex] = useState<any>(0.0);
+    const [rightPaneFlex, setRightPaneFlex] = useState<any>(0.0);
+    const showDocs = useSidePanelDocsStore_show();
+    const displaySidePanel = showDocs && !belowMd;
```
travjenkins, line 38: Storing state related to docs. Thinking this _could_ eventually end up in a stand alone hook but not sure yet.

src/components/content/MessageWithLink.tsx
```diff
@@ -15,9 +16,12 @@ function MessageWithLink({ messageID }: Props) {
...
                     <ExternalLink
-                        link={intl.formatMessage({
-                            id: `${messageID}.docPath`,
-                        })}
+                        link={
+                            link ??
+                            intl.formatMessage({
+                                id: `${messageID}.docPath`,
+                            })
+                        }
```
travjenkins, line 24: This kinda sucks but might be okay. We used to have a steady pattern on how all inline links were handled and now we are breaking it... but it feels better breaking it than it would to duplicate all the code in this component.

src/components/shared/Entity/EndpointConfig/Header.tsx
```diff
@@ -29,12 +28,12 @@ function EndpointConfigHeader({ docsPath }: Props) {
                 <FormattedMessage id="entityCreate.endpointConfig.heading" />
             </Typography>
 
-            {docsPath && docsPath.length > 0 ? (
+            {belowMd && docsUrl && docsUrl.length > 0 ? (
```
travjenkins, line 31: We now can fetch the URL from the store now and then we just show for belowMd because then we just skip showing the docs inline.

src/components/shared/Entity/EndpointConfig/index.tsx
```diff
@@ -97,6 +103,16 @@ function EndpointConfig({
...
 
+    useEffect(() => {
+        if (connectorTag) {
+            setDocsURL(connectorTag.documentation_url);
+        }
+
+        return () => {
+            resetState();
+        };
+    }, [connectorTag, resetState, setDocsURL]);
```
travjenkins, line 114: This might not be the best spot in the long run... but it works for now. This is previously the component that passed the `documentation_url`  to the header so it seemed to make sense.

kiahna-tucker (changes requested):
The docs panel overlaps the CLI edit command(s) popper, rendering it virtually unusable when the panel is open. And when the docs panel is in light mode but the application is in dark mode, the docs are illegible.

Additionally, there is a single suggestion in a review comment and I have a few questions:

* Do we feel that it is clear that the docs panel can be resized? I only ask because the reflex splitter used to partition the _list and details_ component required a styling adjustment for users to recognize that the section widths could be adjusted.

* Should the docs panel **always** open when the selected connector changes, even if the user had explicitly closed the panel while in the current workflow? That is the present behavior which feels a little _forceful_, for lack of a better term.

* Should the close button in the docs panel header be larger given the size of the header text (or vice versa)? I am on fence as to whether the close button should have a greater presence in the header than it does currently.

### References

**Hidden CLI Popper**

pr_review-500-docs_panel-cli_commands_popper_hidden

<br />

**Docs Panel in Light Mode | Application in Dark Mode**

pr_review-500-docs_panel-dark_app_with_light_panel

src/components/shared/Entity/EndpointConfig/index.tsx
```diff
@@ -35,6 +39,8 @@ function EndpointConfig({
     hideBorder,
 }: Props) {
     const intl = useIntl();
+    const setDocsURL = useSidePanelDocsStore_setUrl();
+    const resetState = useSidePanelDocsStore_resetState();
```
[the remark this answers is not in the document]
travjenkins: Yeah.. I was considering that before.

kiahna-tucker (approved):
> The styling doesn't matter as you are not running local docs

Approved, given the following interpretation of the quoted comment is apt: _The styling doesn't matter if running local docs_.

PR #509, opened 2023-03-07
feature (GitPod): Implement `+ New Collection` UI

jshearer:
## Changes

This adds a New Collection button to the collections table, which opens up a modal that takes you to a GitPod instance.

Some things:
* It turns out that we need to specify the derivation's name. As it is, I just added the name input before the "Proceed to GitPod" button since there weren't mocks for it. Thoughts on that?
  * On this topic, it feels kinda smelly to re-implement this name input logic when it already exists on the entity creation pages. I tried to reuse that logic but it was too deeply tied into the logic for entity creation and I gave up.
* We don't have the actual derivation templates set up yet. Currently we send the following data to GitPod, which we'll use to fill out a template for the new derivation:
  * `FLOW_DRAFT_ID`
  * `FLOW_TEMPLATE_TYPE`: `sql`, or `typescript`
  * `FLOW_TEMPLATE_MODE`: `single`, or `multi`
     This will determine whether we use the "filter/map" template or the "join" template
  * `FLOW_COLLECTION_NAME`: The name of the new collection to be created

I'm sure that I'm missing important things like internationalization, but I wanted to get this PR up sooner rather than later in order to get eyeballs on it. Questions I have for the UI team:

* Do we have a standard way to style components? I've used at least 3 different ways in this PR (`styled()`, `sx={{}}`, and `style={{}}`), and it feels smelly
* Do you think this component is big enough to warrant decoupling state into a store? It feels like it might be, but I didn't really have enough experience with Zustand to comfortably do that while also trying to get this working and looking good
* I have a bunch of experience with React, but less with our specific codebase. What would you say I should do differently with work like this?

## Issues

The GitPod URLs generated are currently set to use the `jshearer/getting_started` branch in the gitpod repo. Once that branch is merged (https://github.com/estuary/flow-gitpod-base/pull/1), we'll swap over to using the main branch instead.

## Screenshots

Screen Shot 2023-03-07 at 21 07 54

Screen Shot 2023-03-07 at 21 06 15

Screen Shot 2023-03-07 at 21 17 33


dyaffe:
The design for collection name looks good to me!

kiahna-tucker (approved):
Approved in advance.

Issue #510, opened 2023-03-07
Don't Blow Away Shard Setting on Materialization Edit


dyaffe:
**Bug**
If a user manually adds shard settings to a materialization via the CLI, then edits that materialization via the UI, we delete them.

**Expected**
We shouldn't blow away those shard settings on deletion

**Proposal**
We could either manually blow those settings away or add maxTxnDuration and minTxnDuration as advanced settings in the UI.  The latter would be preferable if possible and the names in the UI would ideally be:

`Max Sync Frequency` and `Min Sync Frequency`

dyaffe:
Those names mean kind of nothing to me -- definitely less than the ones I proposed @jgraettinger and I'd imagine they would mean the same too users too

jgraettinger:
> we will still sync data to other systems continuously and only close transactions (if that's even something the connector considers to be meaningful) at those intervals. 🤦 

This is inaccurate. Inside of a materialization transaction, we'll stream loads as needed but we _do_ hold back documents to store until the end of the materialization transaction, so calling it a sync frequency actually is pretty reasonable.

PR #512, opened 2023-03-13
Preserve task specification properties that cannot be edited via UI form

kiahna-tucker:
## Changes

Create a single, task specification generation utility function that preserves the values of specification properties that cannot be edited via a UI form.

## Tests

1. Create a capture without manipulating form data after discovery.

1. Create a capture and manipulate data in the details form at least once after initial discovery.

1. Create a capture and manipulate data in the endpoint config form at least once after discovery.

1. Create a capture and manipulate data in the resource config form at least once after discovery.

1. Create a materialization without manipulating form data after discovery.

1. Create a materialization and manipulate data in the details form at least once after initial processing.

1. Create a materialization and manipulate data in the endpoint config form at least once after initial processing.

1. Create a materialization  and manipulate data in the resource config form at least once after initial processing.

## Issues

#510


src/utils/workflow-utils.ts
```diff
@@ -1,8 +1,58 @@
...
+            const resourceConfig = resourceConfigs[collectionName].data;
+
+            const existingBindingIndex = draftSpec.bindings.findIndex(
+                (binding: any) => binding[collectionNameProp] === collectionName
+            );
+
+            if (existingBindingIndex > -1) {
+                draftSpec.bindings[existingBindingIndex].resource = {
+                    ...resourceConfig,
+                };
```
travjenkins, line 42: Why are we using the spread operator here / down below?

travjenkins: Ahhh... I think I made that change and can not longer remember what was driving it :(

src/utils/workflow-utils.ts
travjenkins, line 1: Moving these functions to workflow utils was a good call.

travjenkins (approved):
looks good

PR #525, opened 2023-03-20
prod: prune unbound discovered collections at end of workflows only

kiahna-tucker:
## Changes

Prevent a discovered binding removed via the binding selector in the UI from being deleted from the draft. This prevents the collection from being bound to the collection a subsequent time in a given workflow. Unbound, discovered collections should only be pruned at the end of a workflow, prompted by a click of the _Save and Publish_ CTA.

## Tests


travjenkins (approved):
looks good

PR #527, opened 2023-03-20
prod: adjust end of workflow collection pruning approach

kiahna-tucker:
## Changes

Adjust the collection pruning approach used by the _Save and Publish_ CTA as follows: obtain a list of all collections that belong to the draft, filter out the collections that are bound to the given task, and explicitly delete the collections that are determined to **not** be bound to the task.

## Tests


travjenkins (approved):
looks good.

I added a handler for when there is nothing to delete

Issue #540, opened 2023-04-04
Skip Materialization Selection Section of None Found


dyaffe:
**Problem**
If a user has now possible materializations that are pre-existing for a particular capture, we should skip the form selector and take them to configuration of their materialization.

Steps to recreate:

1. Go to an account with no materializations
2. Select New Materialization
3. Go to Selector screen with nothing to select except "new materialization"

This effectively has the user click "New Materialization" twice unnecessarily

kiahna-tucker:
The title for this defect is misleading, but understandably so given what can be observed in the application. The hydrator for the "existing entity check" store -- which determines whether the page should be shown -- is **only** filtering out `ops` tasks, meaning tasks under the `demo/` tenant are included in the response.

The component responsible for populating (and filtering) the list of existing entity cards on this page does correctly filter out tasks under the `ops` and `demo/` tenants, hence the list of existing tasks appears to be empty.

PR #542, opened 2023-04-04
Omit demo tasks from existing entity check hydrator

kiahna-tucker:
## Changes

Omit tasks under the `demo/` prefix from the existing task check performed by the store hydrator. Currently, this check is only performed in the materialization creation workflow.

## Tests

## Issues

#540

### DIFF
--- a/src/api/liveSpecsExt.ts
+++ b/src/api/liveSpecsExt.ts
@@ -269,7 +269,7 @@ const getLiveSpecsByCatalogNames = async (
 const getLiveSpecsByConnectorId = async (
     specType: EntityWithCreateWorkflow,
     connectorId: string,
-    prefixFilter?: string
+    prefixFilters?: string[]
 ) => {
     const taskColumns: string =
         specType === 'capture'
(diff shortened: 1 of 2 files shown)


src/api/liveSpecsExt.ts
```diff
@@ -284,12 +284,14 @@ const getLiveSpecsByConnectorId = async (
...
-        );
+    if (prefixFilters && prefixFilters.length > 0) {
+        prefixFilters.forEach((prefix) => {
+            queryBuilder = queryBuilder.not(
+                'catalog_name',
+                'ilike',
+                `${prefix}/%`
+            );
+        });
     }
```
travjenkins, line 295: This is good with me. 

Thinking... I wonder if there is any impact of handling it like this vs using the `or` function.

src/components/shared/Entity/ExistingEntityCards/Store/create.ts
```diff
@@ -61,7 +61,7 @@ const getInitialState = (
                 const { data, error } = await getLiveSpecsByConnectorId(
                     entityType,
                     connectorId,
-                    'ops'
+                    ['ops', 'demo']
```
travjenkins, line 64: I am wondering if we should eventually make this more "dynamic"... not totally sure how we would but might be nice.

travjenkins (approved):
LGTM

Issue #513, opened 2023-03-13
OAuth does not consistently set client-related endpoint configuration properties


kiahna-tucker:
## Bug

Supplying credentials via OAuth does not consistently set client-related endpoint configuration properties (e.g., client ID, client secret), ultimately preventing the user from successfully publishing their changes.

### Recreation Steps:

1. Enter the edit workflow for a task using a connector with OAuth as an authentication option (e.g., Google Sheets).

1. Edit the data in the endpoint configuration form.

1. Use OAuth to resupply your credentials and observe that an error icon remains in the _Endpoint Config_ form section header.

1. Click the _Next_ CTA and observe that a page-level error appears claiming errors exist within the endpoint configuration form. The reported erroring fields often (if not always) correspond to the schema properties an alternative authentication option (e.g., the service account key option for a Google Sheets connector).

1. Change the selected tab in the _Authentication_ section of the endpoint config form and observe that all _Endpoint Config_ form section error indicators disappear.

1. Reselect the OAuth tab in the _Authentication_ section of the endpoint config form and observe that a page-level error appears claiming errors exist within the endpoint configuration form. The reported erroring fields often (if not always) correspond to the expected schema properties for the OAuth authentication option.

1. Use OAuth to resupply your credentials and observe that all _Endpoint Config_ form section error indicators disappear.

1. Click the _Next_ CTA.

1. View the task specification in the _Specification Editor_ form section and observe that the client-related endpoint configuration properties contain empty strings.

**NOTE:** This defect can be observed in the create workflow as well; recreation steps pending.

## Expected

Supplying credentials via OAuth **should** consistently set client-related endpoint configuration properties (e.g., client ID, client secret).

## Screenshots

**Capture Edit | Endpoint config form data altered and OAuth credentials supplied**

<img width="1019" alt="issue-000-credentials_not_set_via_oauth-form_data_changed" src="https://user-images.githubusercontent.com/77648584/224830899-1ab33282-f280-46e2-bfc9-309094a3795e.PNG">

<br />

**Capture Edit | Next CTA clicked after endpoint config form data altered**

<img width="1019" alt="issue-000-credentials_not_set_via_oauth-next_clicked_after_form_data_changed" src="https://user-images.githubusercontent.com/77648584/224831179-5a7b0046-50c2-41bf-9ade-a6f872a0dc16.PNG">

<br />

**Capture Edit | Alternate authentication option selected**

[...]

PR #549, opened 2023-04-05
Getting OAuth and Resource Config errors to play nice together

travjenkins:
## Changes

1. Refactoring to reduce how big the OAuth renderer is
2. Putting the old KeyValueList back in
3. Changes how OAuth renderer handles default values
4. Reset and hydrate resource config when connector changes

## Tests

Manually tested

## Issues

Fixes: https://github.com/estuary/ui/issues/513
Fixes: https://github.com/estuary/ui/issues/571

## Content

None

## Screenshots

Materialize a captures collections works after changing connector
image
image

Capture Create store is correct
image
image


src/components/collection/DataPreview/TableView.tsx
travjenkins, line 1: Some general clean up of logs

src/components/shared/Entity/EndpointConfig/Form.tsx
```diff
@@ -77,7 +71,7 @@ function EndpointConfigForm({ readOnly }: Props) {
                     cells={materialCells}
                     config={defaultOptions}
                     readonly={readOnly || isActive}
-                    validationMode={showValidationVal}
+                    validationMode="ValidateAndShow"
```
travjenkins, line 74: The function never listened to the value so just passing the value

src/components/shared/KeyValueList.tsx
travjenkins, line 1: Refactored this out to prepare for when virtualized stuff will work

src/components/shared/KeyValueList/index.tsx
travjenkins, line 1: Putting back in the non virtualized one until we can sort out the issues

src/context/Workflow.tsx
```diff
@@ -19,4 +19,17 @@ const useEntityWorkflow = () => {
     return useContext(WorkflowContext);
 };
 
-export { WorkflowContextProvider, useEntityWorkflow };
+const useEntityWorkflow_Editing = () => {
```
travjenkins, line 22: Little helper hook that I think can be used all over the app to fetch if workflows are in edit. I did not switch all components to this to prevent needing to test too much

src/forms/renderers/OAuth/index.tsx
```diff
@@ -95,146 +78,50 @@ const OAuthproviderRenderer = ({
...
-            payload.state,
-            payload.code,
-            endpointConfigData
-        );
+    const { hasAllRequiredProps, showAuthenticated } = useAllRequiredPropCheck(
+        data,
+        options,
+        onChangePath,
+        discriminatorProperty
+    );
```
travjenkins, line 99: Moved the logic over to a hook to keep things a bit cleaned inside this huge file.

src/forms/renderers/OAuth/index.tsx
```diff
@@ -95,146 +78,50 @@ const OAuthproviderRenderer = ({
...
-                ...fakeDefaults,
+    // OAuth Stuff
+    const { errorMessage, openPopUp, loading } = useOauthHandler(
+        provider,
+        (tokenResponse) => {
+            // Order matters here
+            //  1. Get the data we have so far
+            //  2. Merge in the injected values as those MUST match what the server is expecting
+            //  3. Add in the response from the server for the access token
+            const updatedCredentials = {
```
travjenkins, line 109: Moved the logic over to a hook to keep things a bit cleaned inside this huge file.

src/forms/renderers/OAuth/useOauthHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+        }
+    };
+
+    // handler for the useOauth stuff
+    const onSuccess = async (payload: any) => {
+        const preparedData = endpointConfigData;
+        preparedData[CREDENTIALS] = {
+            ...endpointConfigData[CREDENTIALS],
+            ...INJECTED_VALUES,
+        };
```
travjenkins, line 42: This is the only real change to this logic... we not make sure the injected value is here. 

Probably does not matter as I don't think the accessToken endpoint uses them but wanted to be careful.

src/forms/renderers/OAuth/index.tsx
travjenkins: The success handler here merges in the injected values again so that they are there when truly needed. This fixes the issue from https://github.com/estuary/ui/issues/513

src/utils/sops-utils.ts
```diff
@@ -26,8 +26,12 @@ const copyEncryptedEndpointConfig = (
                 : '';
 
         if (isPlainObject(value)) {
+            // Need to generate the new nested object in the template
+            endpointConfigTemplate[truncatedKey || key] = {};
```
travjenkins, line 30: If this is nested then I feel like we need to populate it.

src/lang/en-US.ts
```diff
@@ -926,6 +926,7 @@ const NewTransform: ResolvedIntlConfig['messages'] = {
 };
 
 const CustomRenderers: ResolvedIntlConfig['messages'] = {
+    'oauth.error.credentialsMissing': `need to complete OAuth`,
```
kiahna-tucker, line 929: Should this message be written in sentence case?

src/context/Workflow.tsx
kiahna-tucker: Name alternatives to consider: `useEntityWorkflow_Edit`, `useWorkflow_EntityEdit`, or `useEntityEditWorkflow`. _Thinking out loud._

kiahna-tucker (approved):

src/lang/en-US.ts
travjenkins: No - it is matching these kind of errors

image

Issue #552, opened 2023-04-06
Update GitPod UX


travjenkins:
https://github.com/estuary/flow-gitpod-base/commit/760639e41678331cb84060edcfaa89be41ae9991
- [X] ? Update the ReadMe in the GitPod project ?

https://github.com/estuary/ui/pull/556 (prev. https://github.com/estuary/ui/pull/553)
- [x] Show a close button on the modal
- [x] Change CTA to be named "Transformation"
- [X] Make it clear the user should use GitPod after the window is opened
- [x] Better template spec for sql
- [x] Better template spec for typescript

travjenkins:
Closing as we have completed enough of this from the UI/Dashboard standpoint. More work is being done in the `flow-gitpod-base` repo.

PR #553, opened 2023-04-06
fix: Update gitpod specs

jshearer:
## Changes

We no longer need to add the existing specs to the draft, but instead create a new draft collection

### TODO:
* Figure out the right template for a typescript derivation vs a SQL one, and wire that up to the state of the toggle so we actually give you the template derivation that you picked in the UI

## Implements part of:
https://github.com/estuary/ui/issues/552

### DIFF
--- a/src/components/transformation/create/index.tsx
+++ b/src/components/transformation/create/index.tsx
@@ -22,7 +22,6 @@ import {
 } from '@mui/material';
 import { createEntityDraft } from 'api/drafts';
 import { createDraftSpec } from 'api/draftSpecs';
-import { getLiveSpecsByCatalogNames } from 'api/liveSpecsExt';
 import { createRefreshToken } from 'api/tokens';
 import useCombinedGrantsExt from 'hooks/useCombinedGrantsExt';
 import useLiveSpecs from 'hooks/useLiveSpecs';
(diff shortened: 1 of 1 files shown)


src/components/transformation/create/index.tsx
```diff
@@ -148,26 +147,40 @@ function TransformationCreate() {
                 );
             }
             const draftId: string = draft.data[0].id;
-            const specsByName = await getLiveSpecsByCatalogNames(
-                null,
-                Array.from(selectedCollectionSet)
+            const spec = {
```
travjenkins, line 150: This function is slowly growing... can we move this over into a stand alone hook we could use?

src/components/transformation/create/index.tsx
```diff
@@ -148,26 +147,40 @@ function TransformationCreate() {
...
+                        your_key: {
+                            type: 'string',
+                        },
+                    },
+                    type: 'object',
+                    required: ['your_key'],
+                },
+                derive: {
+                    using: {
+                        sqlite: { migrations: ['migration_1.sql'] },
```
jgraettinger, line 163:
```
{
    "schema": {
      "type": "object",
      "properties": {
        "your_key": {
          "type": "string"
        }
      },
      "required": [
        "your_key"
      ]
    },
    "key": [
      "/your_key"
    ],
    "derive": {
      "using": {
        "sqlite": {
          "migrations": [
            "BASE.migration.0.sql"
          ]
        }
      },
      "transforms": [
        {
          "name": "TRANSFORM",
          "source": "demo/wikipedia/recentchange",
          "lambda": "BASE.lambda.TRANSFORM.sql"
        }
      ]
    }
}
```

This is a canonical SQLite template to generate ☝️ 
This is a canonical TypeScript template to generate 👇 

```
{
    "schema": {
      "type": "object",
      "properties": {
        "your_key": {
          "type": "string"
        }
      },
      "required": [
        "your_key"
      ]
    },
    "key": [
      "/your_key"
    ],
    "derive": {
      "using": {
        "typescript": {
          "module": "BASE.ts"
        }
      },
      "transforms": [
        {
          "name": "TRANSFORM",
          "source": "demo/wikipedia/recentchange"
        }
      ]
    }
  }
```

@jshearer TRANSFORM and BASE are templated parameters for the base name (ex, BASE of `johnny/testing` is `testing`) and the transform name respectively. Using these names exactly is desired because it means that, if the spec is authored, published, and then pulled again, the filenames will stay consistent.

I tested these by creating a draft and manually updating it via:

```
# Inspect raw draft specs
$ RUST_LOG=info ~/flowctl raw get --table draft_specs --query draft_id=eq.09:68:5a:e4:74:60:08:00 | jq '.'

# Update draft spec
$ RUST_LOG=info ~/flowctl raw update --table draft_specs --query id=eq.09:68:73:65:9e:60:50:00 --body "$(<path/to/row.json)"
```

Note that in row.json above, you need to wrap everything in the table column names `{"catalog_name": "johnny/testing", "spec": { ... }}` etc.

PR #556, opened 2023-04-07
Gitpod dialog UX Tweaks 1

travjenkins:
## Changes

1. Update content from `collection` to `transformation` to reduce possible confusion
2. Add a dedicated close button to the dialog
3. Add more fine grained errors around URL missing vs Window not opening
4. Update the modal when GitPod is opened

## Tests

Manually tested

## Issues

Contributes to: https://github.com/estuary/ui/issues/552

## Content

Updating content that was discussed with Dave during 1on1

## Screenshots

Updated button to make it clear this is for a transformation
image

Updated content in the modal
image

Handle if the URL could not be generated
image

Handle if the window was not able to be opened
image

After GitPod window is opened
image

Gitpod files for SQL transformation
image

Gitpod files for TS transformation
image

### DIFF
--- a/src/components/derivation/Create/index.tsx
+++ b/src/components/derivation/Create/index.tsx
@@ -16,18 +16,18 @@ function DerivationCreate() {
     // it's closed, so you don't reopen it and have your previous
     // selections still selected, which would be unexpected.
     const [newCollectionKey, setNewCollectionKey] = useState(0);
-    const [showBackdrop, setShowBackdrop] = useState(false);
+    const [showConfirmation, setShowConfirmation] = useState(false);
 
     const closeDialog = () => {
         navigate(authenticatedRoutes.collections.fullPath);
-        setShowBackdrop(false);
+        setShowConfirmation(false);
         setNewCollectionKey((k) => k + 1);
     };
 
     return (
         <Dialog
             open
-            fullWidth={!showBackdrop}
+            fullWidth={!showConfirmation}
             maxWidth="lg"
             onClose={closeDialog}
             aria-labelledby={ARIA_LABEL_ID}
@@ -36,7 +36,7 @@ function DerivationCreate() {
                 <FormattedMessage id="newTransform.modal.title" />
             </DialogTitleWithClose>
             <DialogContent>
-                <Collapse in={showBackdrop}>
+                <Collapse in={showConfirmation}>
                     <AlertBox
                         short
                         severity="info"
@@ -49,12 +49,12 @@ function DerivationCreate() {
                         <FormattedMessage id="newTransform.info.gitPodWindowMessage" />
                     </AlertBox>
                 </Collapse>
-                <Collapse in={!showBackdrop}>
+                <Collapse in={!showConfirmation}>
                     <TransformationCreate
                         key={newCollectionKey}
                         postWindowOpen={(gitPodWindow) => {
                             if (gitPodWindow) {
-                                setShowBackdrop(true);
+                                setShowConfirmation(true);
                             }
                         }}
                     />
--- a/src/components/derivation/Create/index.tsx
+++ b/src/components/derivation/Create/index.tsx
@@ -1,5 +1,6 @@
-import { Dialog, DialogContent } from '@mui/material';
+import { Collapse, Dialog, DialogContent, Typography } from '@mui/material';
 import { authenticatedRoutes } from 'app/routes';
+import AlertBox from 'components/shared/AlertBox';
 import DialogTitleWithClose from 'components/shared/Dialog/TitleWithClose';
 import TransformationCreate from 'components/transformation/create';
 import { useState } from 'react';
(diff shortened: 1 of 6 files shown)


src/components/derivation/Create/index.tsx
```diff
@@ -13,22 +16,49 @@ function DerivationCreate() {
...
+                        severity="info"
+                        title={
+                            <Typography>
+                                <FormattedMessage id="newTransform.info.gitPodWindowTitle" />
+                            </Typography>
+                        }
+                    >
+                        <FormattedMessage id="newTransform.info.gitPodWindowMessage" />
+                    </AlertBox>
+                </Collapse>
```
travjenkins, line 51: Show some kind of confirmation when the window has been opened in case a user does not directly go to GitPod.

src/components/shared/Dialog/TitleWithClose.tsx
travjenkins, line 1: Shareable dialog title with a close button in the corner. Normally our dialogs close with a verbal cta on the bottom. However, that does not work very well with the transformation create design. In addition, makes it a _little_ harder to share the close functionality.

src/components/transformation/create/generateTransformSpec.ts
travjenkins, line 1: Helper to generate the specs we need. Decided to stick with a plain function over a hook since it does not _really_ need any state.

src/components/shared/Dialog/TitleWithClose.tsx
travjenkins, line 1: Shareable dialog title with a close button in the corner. Normally our dialogs close with a verbal cta on the bottom. However, that does not work very well with the transformation create design. In addition, makes it a little harder to share the close functionality.

src/components/transformation/create/index.tsx
```diff
@@ -373,16 +369,32 @@ function TransformationCreate() {
...
                         <LoadingButton
-                            disabled={!!entityNameError || !!submitButtonError}
                             fullWidth
                             variant="contained"
                             loading={urlLoading}
+                            disabled={
+                                !!entityNameError ||
+                                !!submitButtonError ||
+                                urlLoading
+                            }
```
travjenkins, line 379: Adding the loading check to reduce the chance of double clicks

src/components/transformation/create/index.tsx
```diff
@@ -373,16 +369,32 @@ function TransformationCreate() {
...
+                                        '_blank'
+                                    );
+
+                                    if (!gitPodWindow || gitPodWindow.closed) {
+                                        displayError(
+                                            intl.formatMessage({
+                                                id: 'newTransform.errors.gitPodWindow',
+                                            })
+                                        );
+                                    }
```
travjenkins, line 395: Same check we do with OAuth. Isn't perfect but helps in a situation where the Browser is blocking stuff.

src/components/transformation/create/index.tsx
```diff
@@ -391,12 +403,14 @@ function TransformationCreate() {
...
-                            >
-                                <FormattedMessage id="newTransform.instructions" />
+                        <Stack spacing={1}>
+                            <Typography variant="caption">
+                                <Box>
+                                    <FormattedMessage id="newTransform.instructions1" />
+                                </Box>
+                                <Box>
+                                    <FormattedMessage id="newTransform.instructions2" />
+                                </Box>
```
travjenkins, line 413: Changed the color as it was too hard to read.
Made these be on two lines due to how the words wrapped was very odd.

travjenkins, line 413: Changed the color as it was too hard to read.
Made these be on two lines due to how the words wrapped was very odd.

src/lang/en-US.ts
```diff
@@ -900,23 +900,27 @@ const Docs: ResolvedIntlConfig['messages'] = {
...
+    'newTransform.collection.label': `Transformation Name`,
     'newTransform.errors.collection': `Select A Source Collection`,
     'newTransform.errors.name': `Name Your Transform`,
     'newTransform.errors.prefixMissing': `No prefix selected`,
     'newTransform.errors.namePattern': `Name does not match pattern`,
     'newTransform.errors.nameInvalid': `Invalid entity name`,
     'newTransform.errors.nameMissing': `Missing entity name`,
-    'newTransform.errors.gitPod': `Failed to open GitPod`,
-    'newTransform.stepper.step1.label': `Select your collection`,
+    'newTransform.errors.urlNotGenerated': `We failed to generate the proper URL to start GitPod. ${Error['error.tryAgain']}`,
```
travjenkins, line 914: Making this error match what was created for OAuth

src/lang/en-US.ts
```diff
@@ -900,23 +900,27 @@ const Docs: ResolvedIntlConfig['messages'] = {
...
+    'newTransform.errors.gitPodWindow': `Failed to open GitPod. Your browser may be blocking it from opening. Please ensure your browser allows pop-ups.`,
```
travjenkins, line 915: New errors match the ones made for OAuth... just with `oauth` replaced with `GitPod`

jgraettinger:
👍 I just ran the workflow for both SQL and Typescript => Gitpod with no issues (though the image inside of gitpod still needs work... 🤔 )

jgraettinger:
Some UX feedback (/cc @dyaffe)

I think we are all fine with "+ New Transformation" being the call-to-action in the collections page: 
image
It's close enough to what's happening and is a clear PLG call-to-action.

Once you're in that modal, though, we need to be more precise about using the correct terminology.

image
=> "Derive a new Collection"

image
=> "Select source collections"

image
=> "Write transformation**s**" ? (pluralized).

image
=> "Collection Name" or "Derived Collection Name"

Similarly: 
image
=> "Name your Derived Collection"

For reference, I've pasted the current modal below.

Otherwise, this looks great! I've tested the workflows and it's doing the right things.

image

jgraettinger (approved):
LGTM % noted text updates.

Reviewed UX, not UI code.

jgraettinger:
I'd like to demo this in the morning, if it's possible to merge beforehand. Thanks

PR #566, opened 2023-04-14
oauth2: get code_verifier and pass it to access_token request

mdibaiee:
## Changes

- This is part of https://github.com/estuary/flow/issues/1003
- Sister pull-request: https://github.com/estuary/flow/pull/1004
- We basically have an additional security parameter that the OAuth edge function returns to the browser when initiating the request, and the front-end needs to pass it back in when requesting access token
- Not sure if `useState` is a good way of holding onto the `code_verifier` here 🤔 

## Tests

Tested locally

## Issues

- https://github.com/estuary/flow/issues/1003

## Content

N/A

## Screenshots

N/A


src/forms/renderers/OAuth.tsx
```diff
@@ -170,11 +170,12 @@ const OAuthproviderRenderer = ({
     };
 
     // handler for the useOauth stuff
-    const onSuccess = async (payload: any) => {
+    const onSuccess = async (payload: any, codeVerifier: string) => {
```
travjenkins, line 173: Pass the verifier to the success function

src/forms/renderers/OAuth.tsx
```diff
@@ -219,7 +220,11 @@ const OAuthproviderRenderer = ({
             );
         } else if (!isEmpty(fetchAuthURL.data)) {
             // This kicks off the call and the success is handled with the onSuccess/onError
-            getAuth(fetchAuthURL.data.url, fetchAuthURL.data.state);
+            getAuth(
+                fetchAuthURL.data.url,
+                fetchAuthURL.data.state,
+                fetchAuthURL.data.code_verifier
+            );
```
travjenkins, line 227: Now when we have the URL we pass along the verifier. We could probably change this to just pass the single `fetchAuthURL.data` object given how much is used but did not want to change all that for this.

travjenkins (approved):
LGTM

PR #572, opened 2023-04-17
Performance / Support tweaks

travjenkins:
## Changes

1. Disable response/request body filtering for LogRocket
2. Add fonts to the app
3. Enable Gzip

## Tests

Manual

## Issues

N/A

## Content

N/A

## Screenshots

`compress-create-react-app` output
image


src/index.tsx
```diff
@@ -1,3 +1,4 @@
+import 'index.css';
```
travjenkins, line 1: Load in the css so we get the font files.

skord (approved):
Nginx Conf looks good to me.

jshearer (approved):
So I think this is fine, but why do we need `compress-create-react-app`? Shouldn't nginx's gzip deal with serving compressed assets if the client requests them?

skord:
@jshearer I can't speak to that one directly, but anything that can be gzipped ahead of time should probably be gzipped ahead of time as an optimization.

PR #578, opened 2023-04-21
Endpoint config and Resource config rehydration changes

travjenkins:
## Changes

1. Make resource config only carry forward collections if the materialization is rehydrating. Otherwise - the collections (outside of preselected with URL params) are cleared out.
2. Have Endpoint config reset the "needs server" and encrypted config when connector changes

## Tests

Manually

## Issues

Caught from a slack convo and LogRocket

## Content

N/A

## Screenshots

N/A


kiahna-tucker (approved):
Approved because this PR accomplishes everything it sets out to accomplish. I would like to discuss the UX tradeoffs in greater detail today.

Issue #575, opened 2023-04-20
UI can request outdated versions of the app


travjenkins:
**Bug**
A user that has been using the UI for a long time they can run an old version of the UI. This was always the case before code splitting. However, now that the app is split this can cause weird issues for the user where a part of the app fails.

**Expected**
The user loads the latest UI as much as possible. We will need to figure how we want to handle this. It could be we force reload the page or we ask the user to. Also, there is a chance we could sometimes do this in the background.

**Screenshots**
![image](https://user-images.githubusercontent.com/270078/233415551-f80d48f2-5334-45cd-a896-348b482188df.png)

**Desktop (please complete the following information):**
N/A

**Additional context**
N/A

travjenkins:
Good Enough For Now:

We will present a "reload" cta to the user when this happens. Eventually we should be proactive with this and have the UI check if there is a new version it could be pulling.

Also, we will prefetch the application to reduce the change of this kind of issue.

PR #580, opened 2023-04-21
LogRocket error boundary events and offer a reload

travjenkins:
## Changes

1. Right now the error boundary kind of eats errors. This can make it difficult for us to find out when something is wrong. Adding an even for this so we can provide better support.
2. Offering an option to reload the app when a chunk cannot be found.
3. Make special event for PaymentMethod error
4. Keep payment information header when error boundary is showing
5. Small content update

## Tests

Manually

## Issues

Fixes (enough for now): https://github.com/estuary/ui/issues/575

## Content

New message about loading issues

## Screenshots

Pop up when an error for fetching chunk
image

Requested change
image


src/components/capture/useDiscoverCapture.ts
```diff
@@ -69,7 +68,7 @@ import {
 } from 'utils/workflow-utils';
 
 const trackEvent = (payload: any) => {
-    LogRocket.track(CustomEvents.CAPTURE_DISCOVER, {
+    logRocketEvent(CustomEvents.CAPTURE_DISCOVER, {
```
travjenkins, line 71: This will be all over this PR. Moving this call to a single function so we can make it extra safe.

src/components/admin/Billing/index.tsx
```diff
@@ -211,12 +212,31 @@ function AdminBilling() {
...
+                            <>
+                                <Typography
+                                    component="span"
+                                    variant="h6"
+                                    sx={{
+                                        alignItems: 'center',
+                                    }}
+                                >
+                                    <FormattedMessage id="billing.payment_methods.header" />
                                 </Typography>
```
travjenkins, line 224: Copied from the component so the header is always there

src/components/admin/Billing/index.tsx
```diff
@@ -211,12 +212,31 @@ function AdminBilling() {
...
+                            </>
                         }
+                        onError={(errorLoadingPaymentMethods) => {
+                            logRocketEvent(
+                                CustomEvents.ERROR_BOUNDARY_PAYMENT_METHODS,
+                                {
+                                    stack: errorLoadingPaymentMethods.stack,
+                                }
+                            );
+                        }}
```
travjenkins, line 239: Using the error function to call logrocket with a custom event

src/context/Router/index.tsx
travjenkins, line 1: Impacts perf a bit but a good balance between loading everything lazy vs loading all at once. Until we have a more fleshed out solution to ensure the latest split is loaded I think this is alright.

kiahna-tucker (approved):

PR #581, opened 2023-04-21
fix: Gracefully degrade if stripe is catastrophically down, and fix a couple of other nits

jshearer:
Screen Shot 2023-04-21 at 15 58 41

Also make the input form readonly while submitting, and move stripe initialization to inside the `<PaymentMethods>` component so that any errors it causes propagate correctly to the `<ErrorBoundary>`


travjenkins (approved):
lgtm

PR #582, opened 2023-04-21
fix: Loading spinner bug, and use `FormattedMessage` instead of `MessageWithLink`

jshearer:
## Changes

* Fix intermittent bug where full-page loading spinner shows up on Billing because of subtile misunderstanding of SWR hook
* Replace `MessageWithLink` with `FormattedMessage`, which was causing a bunch of errors about missing doc links, which makes sense since we don't have a docs pager for billing yet


travjenkins (approved):
lgtm

PR #586, opened 2023-04-24
Adjust when the organization name on the onboarding page is validated

kiahna-tucker:
## Changes

Validate the contents of the organization name field on the _Onboarding_ page whenever the value changes.

## Tests

1. Enter an invalid character in organization name field and observe that the field assumes an error state. Attempt to submit the form (via the enter key or clicking the CTA) and observe that the form cannot be submitted.

1. Enter a valid string in the organization name field, do not complete the survey, and observe that the form can be submitted.

1. Enter a valid string in the organization name field, complete the survey, and observe that the form can be submitted.


travjenkins (approved):
lgtm

PR #590, opened 2023-04-28
Billing UX Tweaks 2

kiahna-tucker:
## Changes

1. Replace all line charts on the _Billing_ page with a variation of a bar chart. The _Data Volume by Month_ and _Tasks by Month_ views are stacked bar charts and the _Data Volume by Task_ view is a standard bar chart.

1. Reduce the number tasks shown on the _Data Volume by Task_ chart from ten to three. In other words, this chart now presents the top three, data processing tasks over the given interval of time (i.e., the last six months, for the time being).

1. Introduce a tenant selector on the _Billing_ page to scope the contents on the page to the selected tenant.

1. Remove the tenant selector that was conditionally presented in the _Payment Information_ section.

1. Source more content in the _Payment Information_ section from the language file.

1. Use the shared, table loading state component when the payment methods for the selected tenant are being fetched.

1. Correct the incomplete type annotations in the `PaymentMethods` component and, consequently, add temporary error handling for the service that fetches the payment method records. See comment left in the file for more detail. No changes were made to service implementations.

1. Identify the start date and time of a given billing period when a cell in the _month_ column of the billing history table is hovered.

## Tests

## Issues

## Content

## Screenshots


src/components/admin/Billing/TenantOptions.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <AutocompletedField
+            label={intl.formatMessage({
+                id: 'common.tenant',
+            })}
+            options={tenants}
+            defaultValue={tenants[0]}
+            changeHandler={(_event, value) => {
+                setSelectedTenant(value);
+
+                resetBillingSelectTableState();
```
kiahna-tucker, line 45: Resetting the select table store state does not reset the hydration-related state properties. Is this intentional?

Also, I am surprised that the `EntityTable` component does not _reliably_ present a loading state when the query changes. The component listens to the `query.loading` state property, but the listener (i.e., `isValidating`) does not always align with the value of the state slice. When paired with the ripple effects of polling, the _billing history_ table never assumes an intermediate, loading state and, instead, presents stale data while waiting for new query data.

src/components/admin/Billing/PaymentMethods.tsx
kiahna-tucker, line 1: I did not want to alter the service calls in the absence of proper test data and given how close we are to our next major deadline. Therefore, I opted to work around the services as implemented to provide some basic error handling and corrected incomplete type annotations.

src/components/admin/Billing/graphs/graph.css
travjenkins, line 1: This needs removed and converted to SX props within the component

src/components/admin/Billing/TenantOptions.tsx
travjenkins: Yeah - for the same reason as below. Allow for serving cached data. However, we are not using that functionality right now so we could easily just have it set the hydration props in the table as well.

That is by design so that a loading state of the table is not constantly flashed and we can serve up cached data while waiting on new data.

src/components/admin/Billing/graphs/DataByMonthGraph.tsx
```diff
@@ -165,38 +170,60 @@ function DataByMonthGraph() {
...
-                                        date.includes(config.axisValueLabel)
-                                    ) ?? config.axisValueLabel;
-
-                            return `${tooltipTitle}<br />${config.marker} ${dataVolume}`;
-                        } else {
-                            return undefined;
-                        }
+                            if (content) {
+                                content = `${content}
+                                            <div class="tooltipItem">
```
travjenkins, line 187: Once the styling is moved into `sx` we can share these class names so they are always in sync.

src/utils/billing-utils.ts
```diff
@@ -116,30 +118,34 @@ export const formatBillingCatalogStats = (
             const totalCost = evaluateTotalCost(dataVolume, taskCount);
 
             if (billingRecordIndex === -1) {
```
travjenkins, line 120: This block is pushing how far we can go with just duplicating stuff to save time. That object is pretty big and contains logic. This eventually needs to be shared.

travjenkins (changes requested):
Overall this is alright but have some changes called out.

src/components/admin/Billing/graphs/graph.css
kiahna-tucker: This is the recommended way to share styles between eCharts components. As far as I am aware, it is impossible to interpolate an object into the string which contains HTML for the tooltip, in this case. I did attempt to do so. If you have accomplished this yourself or are aware of a code sample, please share a link.

src/components/admin/Billing/index.tsx
```diff
@@ -34,38 +26,42 @@ import {
...
+        billingStats,
+        error,
+        isValidating: isValidatingStats,
+    } = useBillingCatalogStats();
+
+    useEffect(() => {
+        if (!isValidatingGrants) {
+            setTenants(combinedGrants);
+        }
+    }, [setTenants, combinedGrants, isValidatingGrants]);
```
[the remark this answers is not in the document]
kiahna-tucker: Firstly, this property is shared. _Need to switch contexts. Will return to this comment._

travjenkins (approved):
lgtm

PR #595, opened 2023-05-04
Show discover errors and better formatting

travjenkins:
## Changes

1. Allow key value list to take in components as title and val
2. Have draft errors wrap on newlines for errors
3. Have the overflow get handled with scrolling
4. Flipping the DraftErrors title and val so the layout makes a bit more sense
5. 

## Tests

Manually

## Issues

From Google docs backlog

## Content

N/A

## Screenshots

New error format for draft errors
image

### DIFF
--- a/src/components/capture/useDiscoverCapture.ts
+++ b/src/components/capture/useDiscoverCapture.ts
@@ -9,6 +9,7 @@ import {
     useEditorStore_isSaving,
     useEditorStore_persistedDraftId,
     useEditorStore_resetState,
+    useEditorStore_setDiscoveredDraftId,
     useEditorStore_setId,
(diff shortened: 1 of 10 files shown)


jgraettinger:
Adding a function for unpacking scopes. Example when testing a capture with multiple errors (which have scopes with fragment-encoded JSON pointers):

image

The control plane will sometimes send _just_ a catalog name as the scope, as is the case with discovers. That was done to make them less ugly in the UI, and over time we can transition to those being full URLs as well, but they're at least handled reasonably by this PR:

image

travjenkins (commented):
Pushed a quick change to convert double equals to triple

Issue #560, opened 2023-04-11
Nested error windows


jman0119:
**Bug**
I was trying to create a capture, and then received errors in this nested window when I hit Next.

**Expected**
Errors to display not in nested windows.

**Screenshots**
Include a screenshot of what you saw when the bug happened.
![image](https://user-images.githubusercontent.com/39440655/231307049-58e2526a-edc3-47bf-8c30-1a747679edc6.png)

**Desktop (please complete the following information):**
 - OS: [e.g. Windows] Windows
 - Browser [e.g. chrome, safari] Chrome

**Additional context**

PR #597, opened 2023-05-05
Add schema evolution to the UI

travjenkins:
## Changes

1. When a test/save fails check if there were incompatible collections
2. Show a new error section when incompatible changes are caught
3. Wire up button to enable users to call "evolve schemas"

## Tests

Manually tested

## Issues

Fixes https://github.com/estuary/ui/issues/560

## Content

A lot of new content provided by Dave

## Screenshots

Have a new section showing the incompatible collection issues
image

While the schema evolution is running we display similar to save/test
image

If evolution works we fetch latest collections and hide error
image

If a HUGE server/network issue happens trying to start the update
image

If the poller fails to get a status
image


src/components/capture/Edit.tsx
travjenkins, line 1: `Cleanup jobFailed`

src/components/capture/useDiscoverCapture.ts
```diff
@@ -130,94 +111,15 @@ function useDiscoverCapture(
...
-            evaluateDiscoveredCollections,
-            lastPubId,
-            resetCollections,
-            setDiscoveredCollections,
-            setDraftId,
-            setEncryptedEndpointConfig,
-            setPersistedDraftId,
-        ]
-    );
+    const storeDiscoveredCollections = useStoreDiscoveredCaptures();
```
travjenkins, line 117: All the stuff was moved into a function. This was so the evolved schemas could get updated just like the discovered ones. Most everything was moved as it.

src/components/capture/useDiscoverCapture.ts
```diff
@@ -130,94 +111,15 @@ function useDiscoverCapture(
...
 
     const jobFailed = useCallback(
-        (errorTitle: string) => {
+        (error) => {
```
travjenkins, line 120: `Cleanup jobFailed`
This happens elsewhere. Passing in the object each time so we can control the title and message better.

src/components/capture/useDiscoverCapture.ts
```diff
@@ -248,7 +150,11 @@ function useDiscoverCapture(
                     })
                     .order('created_at', { ascending: false }),
                 async (payload: any) => {
-                    await storeDiscoveredCollections(payload.draft_id);
+                    await storeDiscoveredCollections(
+                        payload.draft_id,
+                        entityType,
+                        callFailed
+                    );
```
travjenkins, line 157: Passing these props just because they are required now since this is being created with a hook. Instead of passing these into the hook I decided to pass to the function to reduce the number of "memo" runs would be needed.

src/components/capture/useDiscoverCapture.ts
```diff
@@ -264,16 +170,22 @@ function useDiscoverCapture(
...
-                        jobFailed(payload.error);
+                        jobFailed(DEFAULT_POLLER_ERROR);
                     } else {
-                        jobFailed(`${messagePrefix}.test.failedErrorTitle`);
+                        jobFailed({
+                            title: 'discovery.failed.title',
+                            error: {
+                                message: 'discovery.failed.message',
+                            },
+                        });
```
travjenkins, line 180: `Cleanup jobFailed`
Now have a default job status poller error settings object. Eventually I want to get almost all the `jobFailed` functions shared but that was a bit too much with this PR.

src/components/editor/Bindings/Store/hooks.ts
travjenkins, line 1: `store changes`

src/components/materialization/Edit.tsx
travjenkins, line 1: `Cleanup jobFailed`

src/components/shared/Entity/Actions/SchemaEvolution.tsx
travjenkins, line 1: This all started with a copy of discover and then tweaks. A future thing we need to do is basically make this "Start async task, watch async task, handle failures" stuff more shared and have a pattern for how we do it.

src/components/shared/Entity/Actions/SchemaEvolution.tsx
```diff
@@ -0,0 +1,169 @@
...
+                setIncompatibleCollections([]);
+            },
+            async (payload: any) => {
+                if (payload.error === JOB_STATUS_POLLER_ERROR) {
+                    jobFailed(DEFAULT_POLLER_ERROR);
+                } else {
+                    jobFailed({
+                        ...DEFAULT_POLLER_ERROR,
+                        message: `entityEvolution.serverUnreachable`,
+                    });
```
travjenkins, line 109: Want to use the default heading but use a custom message.

src/components/shared/Entity/Actions/SchemaEvolution.tsx
```diff
@@ -0,0 +1,169 @@
...
+                }
+            }
+        );
+    };
+
+    const save = async (event: React.MouseEvent<HTMLElement>) => {
+        event.preventDefault();
+
+        updateFormStatus(FormStatus.SCHEMA_EVOLVING);
```
travjenkins, line 118: Added a new status to make things a bit more clear even though we do not trigger much off the status

src/components/shared/Entity/Edit/index.tsx
travjenkins, line 1: Showing the section _below_ the normal ones so we can reuse that section if the schema evolution stuff has problems.

src/components/shared/Entity/Error/index.tsx
```diff
@@ -30,10 +30,13 @@ function EntityError({ logToken, error, title, draftId }: Props) {
         <HeaderSummary severity="error" title={title}>
             <Stack direction="column" spacing={2}>
                 <Box>
-                    <Error error={error} hideTitle={true} />
+                    <Error error={error} hideTitle={true} noAlertBox />
```
travjenkins, line 33: Prevents the "nested errors" issue

src/components/shared/Entity/Error/index.tsx
```diff
@@ -30,10 +30,13 @@ function EntityError({ logToken, error, title, draftId }: Props) {
...
 
                     {idForDraftErrors ? (
-                        <DraftErrors draftId={idForDraftErrors} />
+                        <>
+                            <Divider />
+                            <DraftErrors draftId={idForDraftErrors} />
+                        </>
```
travjenkins, line 39: Throwing in a divider for design purpose. Now that the above does not have a box having these errors indented looked a bit better after a line on the page.

src/components/shared/Entity/HeaderSummary.tsx
travjenkins, line 1: Small tweaks to the design of the error

src/components/shared/Entity/IncompatibleCollections/CollectionsList.tsx
travjenkins, line 1: Basically the same code from the collection editor PR with some tweaks. If we continue with this approach we'll probably need to make these shared.

src/components/shared/Error.tsx
travjenkins, line 1: Moved this into the folder `Error`

src/components/shared/Error/Details.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useIntl } from 'react-intl';
+import { hasLength } from 'utils/misc-utils';
+import { ErrorDetails } from './types';
+
+interface Props {
+    error?: ErrorDetails;
+}
+
+// This is not visible to users. This can be used while
+//  developing new code to see all the details supabase is returning
```
travjenkins, line 11: Important note on why this is still here

src/components/shared/Error/Message.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { FormattedMessage, useIntl } from 'react-intl';
+import { logRocketConsole } from 'services/logrocket';
+import { ErrorDetails } from './types';
+
+interface Props {
+    error?: ErrorDetails;
+}
+
+function Message({ error }: Props) {
+    logRocketConsole('Error message displayed', error);
```
travjenkins, line 12: Threw this in so while we debug in prod we could still get access to all the error details (in case there are any details missing)

src/components/shared/Error/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+        return null;
+    }
+
+    if (noAlertBox) {
+        return (
+            <Box sx={{ p: 1, width: '100%' }}>
+                <Message error={error} />
+            </Box>
+        );
+    }
```
travjenkins, line 24: New functionality... allow this to be used with the entire AlertBox wrapping it. Mainly just for the header errors in create/edit

src/hooks/useShardsList.ts
travjenkins, line 1: Made the calls to logging to LogRocket a reusable function to be extra extra safe.

src/lang/en-US.ts
```diff
@@ -860,7 +864,8 @@ const OAuth: ResolvedIntlConfig['messages'] = {
 };
 
 const Supabase: ResolvedIntlConfig['messages'] = {
-    'supabase.poller.failed': `We encountered a problem retrieving the status of this action. Please check your network connection and try again.`,
+    'supabase.poller.failed.title': `Unable To Reach Server`,
+    'supabase.poller.failed.message': `We encountered a problem retrieving the status of this action. Please check your network connection and try again.`,
```
travjenkins, line 868: Added a new title to make it clear what happened and make it more general

src/utils/name-utils.ts
travjenkins, line 1: New function from Phil. This will update the collection name. Eventually we will allow customers to update the names on their own.

src/services/supabase.ts
```diff
@@ -289,6 +290,15 @@ export const jobSucceeded = (jobStatus?: JobStatus) => {
...
 export const JOB_STATUS_POLLER_ERROR = 'supabase.poller.failed';
+export const DEFAULT_POLLER_ERROR_TITLE_KEY = 'supabase.poller.failed.title';
+export const DEFAULT_POLLER_ERROR_MESSAGE_KEY =
+    'supabase.poller.failed.message';
+export const DEFAULT_POLLER_ERROR = {
+    title: DEFAULT_POLLER_ERROR_TITLE_KEY,
+    error: {
+        message: DEFAULT_POLLER_ERROR_MESSAGE_KEY,
+    },
+};
```
travjenkins, line 301: Making some shared setting for `jobFailed` related work

src/components/shared/Entity/Actions/SchemaEvolution.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    );
+                    setFormState({
+                        logToken: response.data[0].logs_token,
+                        showLogs: false,
+                    });
+                }
+            }
+        } else {
+            logRocketEvent('Schema:Evolution:Missing draftId');
+            onFailure({
```
psFried, line 150: `Missing Draft Id` seems like a confusing error message to result from this. It kinda seems like we'd want to treat this as some type of "internal error". WDYT? I'm assuming this would be very uncommon in any case.

psFried (approved):
LGTM :rocket:
This looks great.

src/components/shared/Entity/Actions/SchemaEvolution.tsx
travjenkins: It has never happened as far as we know and also not really sure how this would happen in the real world but we just put it there cause it _could_ happen.

PR #605, opened 2023-05-09
add connector selector to TestJsonForms page

mdibaiee:
## Changes

## Tests

## Issues

## Content

## Screenshots


src/stores/DetailsForm/Store.ts
```diff
@@ -213,6 +213,14 @@ export const getInitialState = (
...
                 }
             } else {
+                const { setDetails_connector } = get();
+                setDetails_connector({
+                    id: connectorId,
+                    iconPath: '',
+                    imageName: '',
+                    imagePath: '',
+                    connectorId,
+                });
```
mdibaiee, line 223: This is a hacky solution I introduced so I can hydrate just the `connectorId` without having to fetch anything else about the connector. The idea is when we want to test a new connector, we create a row for it in the `connectors` table, but no `connector_tags` row. We specify `oauth2_client_id` and `oauth2_client_secret` in `connectors` row and that is used to drive the OAuth flow.

I'm open to updating this logic to be perhaps more specific for the TestJsonForms page instead of being in this `else` block, or perhaps somehow moving it somewhere else so it doesn't interfere with the general logic at all. Let me know what you think.

src/hooks/useConnectors.ts
```diff
@@ -0,0 +1,31 @@
...
+    id: string;
+    title: { 'en-US': string };
+    image_name: string;
+}
+
+export const CONNECTOR_QUERY = `
+    id, title, image_name
+`;
+
+function useConnectors() {
```
mdibaiee, line 14: A hook for fetching `connectors` directory from their own table, without any association with `connector_tags`

src/stores/DetailsForm/Store.ts
travjenkins: We should probably move this given we really want to reduce potential impact that the text page has on the application. 

I will need to think about this a bit.

mdibaiee: @travjenkins I introduced a new Workflow value `test_json_forms` and added an if condition in the `hydrateState` to only apply this logic if `workflow == 'test_json_forms'`, do you think this is sufficient isolation of this logic?

travjenkins: Yeah - that is enough. I was thinking of that as one of the possible options and that seems easiest. Also makes it clear for future work that this store is used in a test page.

travjenkins (approved):
lgtm

src/hooks/useConnectors.ts
travjenkins: Added this comment into the code to make sure this knowledge wasn't lost

PR #613, opened 2023-05-13
prod: propagate grant token in OIDC redirect URL

kiahna-tucker:
## Changes

Include the grant bearer token in the redirect URL of the OIDC authentication options.

## Tests


dyaffe (approved):
Approving this since it looks straightforward and theoretically can't break anything.

Issue #616, opened 2023-05-16
Make Entity Names More Obvious


dyaffe:
**Problem**
If someone only has access to a single tenant (which represents 99% of our users), we should automatically prefix whatever they type as any name field with their tenant name.

We've seen a few users have trouble with our naming scheme and we might as well try to make it is as simple as possible.

If a user has access to more than one, we can keep our current behavior.

PR #618, opened 2023-05-17
Entity name UX Tweaks 1

travjenkins:
## Changes

1. Default the input to the only prefix is there is only one

## Tests

Manually tested create and edit

## Issues

https://github.com/estuary/ui/issues/616

## Content

n/a

## Screenshots

Only allowed prefix will be prefilled
image

Has multiple prefix so not prefilled
image


kiahna-tucker (approved):

Issue #573, opened 2023-04-19
Postfix connector name on materializations


dyaffe:
**Problem**
We should postfix connector names on materializations for symmetry with captures

travjenkins:
We should try to do this the same way we handle Captures. Possibly move that code into a shared location as well.

PR #622, opened 2023-05-18
Add image name to end of Materializations / Prod Collection columns issue

travjenkins:
## Changes

1. Create a shared hook for Capture and Materializations
2. Have Materializations suffix the name with image name
3. Fix Collection columns

## Tests

Manually tested create and edit

## Issues

Fixes https://github.com/estuary/ui/issues/573

## Content

None

### DIFF
--- a/src/components/tables/Collections/useCollectionColumns.tsx
+++ b/src/components/tables/Collections/useCollectionColumns.tsx
@@ -44,7 +44,7 @@ const useCollectionColumns = (): ColumnProps[] => {
     return useMemo(() => {
         if (hasDetails) {
             const response = [...defaultColumns];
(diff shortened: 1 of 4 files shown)


src/components/tables/Collections/useCollectionColumns.tsx
travjenkins, line 1: Fixes issue Phil noticed with Columns

src/hooks/useEntityNameSuffix.ts
travjenkins, line 1: Took logic from discover hook and packaged here. Passing when make it easy. Considered making this return a function but feel like just returning the name we want to use is better as it makes is just use the name everywhere we can.

kiahna-tucker (approved):

PR #628, opened 2023-05-23
fix: OAuth credential nesting

jshearer:
## Changes

I _believe_ this is a subtle bug, but it's entirely possible that this is actually desired behavior that I just don't understand. Thoughts?

`schema` appears to be the whole connector schema, for example:
```json
{
    "$schema": "http://json-schema.org/draft/2020-12/schema",
    "properties": {
        "credentials": {
            "properties": {
                "client_id": {
                    "type": "string",
                    "secret": true
                },
                "client_secret": {
                    "type": "string",
                    "secret": true
                },
                "access_token": {
                    "type": "string",
                    "secret": true
                }
            },
            "additionalProperties": false,
            "type": "object",
            "title": "Authentication",
            "x-oauth2-provider": "slack"
        }
    },
    "type": "object",
    "required": [
        "credentials"
    ],
    "title": "Slack Connection"
}
```

So what's happening here is we're generating defaults for this schema, which in this case just ends up being
```json
{
  "credentials": {}
}
```
And then setting those as the value under `onChangePath`, which itself is `"credentials"`, resulting in a data that looks like this
```json
{
  "credentials": {
    "credentials": {}
  }
}
```

Instead, we want to generate defaults just for `onChangePath` (`"credentials")`, which should just be `{}`, and then set _those_ as the value at `onChangePath`.


travjenkins (approved):
lgtm

PR #633, opened 2023-05-24
Integrate the billing RPC

kiahna-tucker:
## Changes

1. Update the billing history table, data volume by month chart, and tasks by month chart so that data is sourced from the `billing_report` RPC.

1. Fetch billing data for previous billing cycles only when initializing the _Billing_ page.

1. Fetch billing data for the current billing cycle every thirty seconds.

## Screenshots

pr_screenshot-633-billing_rpc-page_loading

<br />

pr_screenshot-633-billing_rpc-default

<br />

pr_screenshot-633-billing_rpc-no_data_usage

### DIFF
--- a/src/components/admin/Billing/index.tsx
+++ b/src/components/admin/Billing/index.tsx
@@ -14,7 +14,7 @@ import AlertBox from 'components/shared/AlertBox';
 import BillingHistoryTable from 'components/tables/Billing';
 import { eachMonthOfInterval, format, startOfMonth, subMonths } from 'date-fns';
 import useBillingCatalogStats from 'hooks/billing/useBillingCatalogStats';
-import useBillingHistory from 'hooks/billing/useBillingHistory';
+import useBillingRecord from 'hooks/billing/useBillingRecord';
 import usePageTitle from 'hooks/usePageTitle';
 import { useEffect, useMemo, useState } from 'react';
 import { ErrorBoundary } from 'react-error-boundary';
@@ -73,8 +73,8 @@ function AdminBilling() {
         isValidating: isValidatingStats,
     } = useBillingCatalogStats();
 
-    const { billingHistory, isValidating: isValidatingHistory } =
-        useBillingHistory(currentMonth);
+    const { billingRecord, isValidating: isValidatingRecord } =
+        useBillingRecord(currentMonth);
 
     useEffect(() => {
         if (!isValidatingStats && billingStats) {
@@ -139,16 +139,16 @@ function AdminBilling() {
     useEffect(() => {
         if (
             historyInitialized &&
-            !isValidatingHistory &&
-            hasLength(billingHistory)
+            !isValidatingRecord &&
+            hasLength(billingRecord)
         ) {
-            updateBillingHistory(billingHistory);
+            updateBillingHistory(billingRecord);
         }
     }, [
         updateBillingHistory,
-        billingHistory,
+        billingRecord,
         historyInitialized,
-        isValidatingHistory,
+        isValidatingRecord,
     ]);
 
     useUnmount(() => resetBillingState());
--- a/src/components/admin/Billing/index.tsx
+++ b/src/components/admin/Billing/index.tsx
@@ -17,16 +17,18 @@ import useBillingCatalogStats from 'hooks/billing/useBillingCatalogStats';
 import useBillingRecord from 'hooks/billing/useBillingRecord';
 import usePageTitle from 'hooks/usePageTitle';
 import { isArray, isEmpty } from 'lodash';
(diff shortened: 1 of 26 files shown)


src/components/admin/Billing/index.tsx
```diff
@@ -35,25 +42,43 @@ function AdminBilling() {
...
 
+    const selectedTenant = useBilling_selectedTenant();
     const setBillingHistory = useBilling_setBillingHistory();
+    const updateBillingHistory = useBilling_updateBillingHistory();
     const setDataByTaskGraphDetails = useBilling_setDataByTaskGraphDetails();
 
     const resetBillingState = useBilling_resetState();
 
+    const [historyInitialized, setHistoryInitialized] =
+        useState<boolean>(false);
```
travjenkins, line 53: This does not need typed

src/components/admin/Billing/index.tsx
```diff
@@ -35,25 +42,43 @@ function AdminBilling() {
...
+
+    const currentMonth = useConstant(() => {
+        const today = new Date();
+
+        return startOfMonth(today);
+    });
```
travjenkins, line 59: Not for this PR... but we should look into creating a hook or something for all our date/time needs so they all reference a single copy given how often we are calculating this stuff now.

src/components/admin/Billing/index.tsx
```diff
@@ -67,10 +92,65 @@ function AdminBilling() {
...
+    useEffect(() => {
+        if (
+            selectedTenant &&
+            dateRange.length > 0 &&
+            !hydrated &&
+            !historyInitialized
+        ) {
+            setHistoryInitialized(true);
+
+            void getBillingHistory(selectedTenant, dateRange).then(
```
travjenkins, line 107: Anything driving including the `void` here? Since we have the `then` that shouldn't be needed.

src/hooks/billing/useBillingRecord.ts
```diff
@@ -0,0 +1,33 @@
+import { BillingRecord, getBillingRecord } from 'api/billing';
+import { useSelectNew } from 'hooks/supabase-swr/hooks/useSelect';
+import { useBilling_selectedTenant } from 'stores/Billing/hooks';
+import { hasLength } from 'utils/misc-utils';
+
+const INTERVAL = 30000;
```
travjenkins, line 6: Please make a shareable `const` for this. Since we keep using 30,000 as a default we might as well keep it in a single place.

src/hooks/billing/useBillingRecord.ts
```diff
@@ -0,0 +1,33 @@
...
+    const { data, error, mutate, isValidating } = useSelectNew<BillingRecord>(
+        hasLength(selectedTenant)
+            ? getBillingRecord(selectedTenant, currentMonth)
+            : null,
+        {
+            errorRetryCount: 3,
+            errorRetryInterval: INTERVAL / 2,
+            refreshInterval: INTERVAL,
+            revalidateOnFocus: false,
+        }
```
travjenkins, line 22: Same here... we can make this shareable.

src/stores/Tables/Billing/Hydrator.tsx
```diff
@@ -32,9 +35,11 @@ export const BillingHistoryTableHydrator = ({
...
     const hydrateState = useBillingTable_hydrateContinuously();
 
-    const { billingHistory, error, isValidating } = useBillingHistory({
-        query,
-    });
+    const {
+        billingRecord: billingHistory,
+        error,
+        isValidating,
+    } = useBillingRecord('');
```
travjenkins, line 42: Is this call why we handle when a `date` or a `string` is passed to this hook?

travjenkins (commented):
Overall this is fine - some minor tweaks

src/components/admin/Billing/index.tsx
kiahna-tucker: Of course not. This is a classic instance of an oversight, intensified when programming without vision.

src/components/admin/Billing/index.tsx
kiahna-tucker: It is widely understood that the type can be inferred. Adding a type annotation to all `useState` hooks is a personal preference for reasons that I do not care to explain. If you are adamant that you would like it removed from your codebase, I shall do so.

src/stores/Tables/Billing/Hydrator.tsx
kiahna-tucker: No, it is because time can be represented as a `string` or a `Date` and I wanted the query builder to handle both cases.

travjenkins (approved):
looks good

Issue #638, opened 2023-05-25
Don't show regex when creating a capture or materialization


dyaffe:
**Problem**
Humans really don't understand regex's.  It is probably better if instead of showing the regex, we show a static field like "Must match patter "organization/name".

travjenkins:
Parking some details:

The enable we'll need to look into wiring up custom `ajv` errors.

Library : https://github.com/ajv-validator/ajv-errors

Discussion (kinda old) : https://jsonforms.discourse.group/t/custom-error-messages/140

travjenkins:
Would could also just tweak this in the catalog name renderer to save time (maybe... not sure how easy this would be just for that renderer)

Issue #627, opened 2023-05-23
Default value not filled when sending form submission request


mdibaiee:
**Bug**
- When a new page is opened for a connector, and the connector has default values for inputs, the default value shows in the inputs, but when clicking "Next", the request sent to config-encryption has an empty `{}` config object.

**Expected**
- The request also has default values filled in

**Screenshots**
![image](https://github.com/estuary/ui/assets/2807772/b578bbae-3d4c-45f6-bf26-8469fe5d3a7a)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Firefox

**Additional context**

travjenkins:
Debug notes: the data is there in the store as expected. However, not making it to the call to the server. Investigating.

Issue #624, opened 2023-05-19
Improve Header UI Entry


dyaffe:
**Problem**
Currently when someone uses the CSV parser and enters headers (steps below), they have to enter headers individually.  This can be problematic because users routinely have 100s of headers.  It would be nice to have a bulk entry option like a string that accepts comma separated values or anything that we can do easily in our current UI to improve the experience:

1. Go to any file store capture (Ex. SFTP)
2. Click parser -> CSV -> headers
3. Add a bunch of headers
4. Save
5. Edit the same specification and they will be gone and you'll have to do it again.

travjenkins:
https://github.com/estuary/ui/issues/621 will be handle with this one. Please see that one for other details.

travjenkins:
Debug notes:

1. `indexOfFittingSchema` is coming back empty for some reason. Gotta figure out why.

PR #641, opened 2023-05-30
Fix several JSON Forms UX bugs and feature requests 

travjenkins:
## Changes

1. [627] Entity Create - When opening create we now default the `serverUpdateRequired` to true
1. [642] JSON Forms - Minor refactoring to reduce duplication of when we add the required option to an element
2. [642] JSON Forms - Add the required option for a normal group layout. Previously we just added to "special cases" like oAuth or anyOf/oneOf.
3. [624] Sops Utils - updated the encrypted config copies to no longer directly update the template object
4. [624] Sops Utils - made the encrypted config copier recursion work correctly. It was not populating the object by default.
5. [638] Catalog Name Renderer - added some code to replace the normal error if it contains the pattern for validation

## Tests

Manually tested with the Slack Materialization, Google Sheets Capture, and S3 Capture

## Issues

Fixes https://github.com/estuary/ui/issues/624
Fixes https://github.com/estuary/ui/issues/627
Fixes https://github.com/estuary/ui/issues/642
Fixes https://github.com/estuary/ui/issues/638

## Content

N/A 

## Screenshots

Section displays open
image

New Catalog Name Error
image


kiahna-tucker (approved):

PR #645, opened 2023-06-02
Endpoint Config form rendering tweaks

travjenkins:
I started working on some larger issues and stumbled on these tweaks while doing that work. I have shelved the other stuff but still want these to get through as I think they add value stand alone.


kiahna-tucker (approved):

Issue #518, opened 2023-03-14
Display determinate progress bar during discovery and publication creation


dyaffe:
**Proposal** 
We should add a loading bar when doing a discover or publish that takes the same amount of time as the backend timeout (I think it's 30s).

This will help users understand the max time that a publish or discover can take.

kiahna-tucker:
FYI: This is a duplicate of issue #488.

PR #650, opened 2023-06-06
Add deterministic load bar

travjenkins:
## Changes

1. Add a  "timer bar" for discover that will slowly fill up over 30 seconds
2. Update header so the new time bar is not always shown
3. New hooks that are somewhat helpful

## Tests

Manually tested

## Issues

Implements https://github.com/estuary/ui/issues/518

## Content

None


kiahna-tucker (commented):
There is a production defect in the capture create workflow that becomes _slightly_ more noticeable with this PR: captures have to be discovered twice before they can be published.

The way we are setting `entityNameChanged` and `endpointConfigServerUpdateRequired` are causing the `EntityCreate` effect below to reset the draft ID (i.e., the `id` editor state property).

```ts
  useEffect(() => {
      const resetDraftIdFlag =
          entityNameChanged ||
          endpointConfigServerUpdateRequired ||
          resourceConfigServerUpdateRequired;

      setDraftId(resetDraftIdFlag ? null : persistedDraftId);
  }, [
      setDraftId,
      endpointConfigServerUpdateRequired,
      entityNameChanged,
      persistedDraftId,
      resourceConfigServerUpdateRequired,
  ]);
```

After initial discovery, the draft ID is reset because `entityNameChanged` **and** `endpointConfigServerUpdateRequired` are `true`.

Additionally, it should be noted that `entityNameChanged` never seems to change value in the capture create workflow and it does not change value more than once in the materialization create workflow

* **Capture Create:** If you alter the catalog name, the _Output Collections_ and _Advanced Specification  Editor_ sections disappear **only** because `resourceConfigServerUpdateRequired` is `true`. And consequently, if the change(s) made to the catalog name are undone, the aforementioned sections do not reappear which forces the user to have to rediscover and configure their resources _despite_ the perfectly viable previous draft.

* **Materialization Create:** If you alter the catalog name, the  _Advanced Specification  Editor_ section disappear because the flag is aptly set. However, if the change(s) made to the catalog name are undone, the aforementioned section does not reappear which forces the superfluous patch of the existing `draft_specs` entry _despite_ the perfectly viable previous draft. 

If you want to address these two, production bugs separately, that is fine. Let me know and I can approve your PR.

Issue #647, opened 2023-06-05
Update entity name in create flow


travjenkins:
There is user confusion around how to properly enter entity name. We would like to split out the tenant/organization part of the name into a stand alone input. Then the rest of the name will be entered by the user. This will follow design approaches similar to transformation creation.

PR #651, opened 2023-06-08
Make object name consistent

travjenkins:
## Changes

1. Create a shared object name input
2. Wired in Grants, Access Links, Entity, and transform create with new input
3. Make the error summary shared between all create flows

## Tests

Manually tested

## Issues

Implements https://github.com/estuary/ui/issues/647

## Content
 
Some minor tweaks using what was there as building blocks

### Transformations
image
image
image
image
image

### Invitations
image
image
image

### Grants
image
image
image


src/components/transformation/create/index.tsx
```diff
@@ -168,54 +122,25 @@ function TransformationCreate({ postWindowOpen }: Props) {
...
-                                    ),
+                            <Box
+                                sx={{
+                                    p: 2,
                                 }}
-                            />
+                            >
+                                <PrefixedName
+                                    standardVariant
+                                    size="medium"
```
kiahna-tucker, line 132: **Personal Note:** all instances of `PrefixedName` in this component should have a size of `small`. Future rollback.

kiahna-tucker (approved):
Approved with moderate testing. The defects called out in this [PR](https://github.com/estuary/ui/pull/650) persist but may be out of scope for this PR.

src/components/transformation/create/index.tsx
travjenkins: I don't think so - in the old design a small input looks pretty weird. That's why it is set to medium.

PR #655, opened 2023-06-13
Enable publications in updated derivation create workflow

kiahna-tucker:
## Changes

1. **Use the shared, _Save and Publish_ button component (i.e., _src/components/shared/Entity/Actions/SaveButton.tsx_) in the updated derivation create workflow.** This integration required that the details form, form state, and resource config stores be initialized when the derivation create workflow is entered because of the selectors used in a child component of the aforementioned button (i.e., `EntityCreateSave` defined in _src/components/shared/Entity/Actions/Save.tsx_). The form state store is the only store whose state is altered after initialization in the derivation create workflow.

1. Enable publications in the updated derivation create workflow via the _Proceed to GitPod_ or _Save and Publish_ CTA.

1. Default all transformation lambdas to an empty string when the updated derivation create workflow is entered.

1. Default the `shuffle` property of a transformation to `any` if unspecified.

1. Allow the SQL Monaco editor to patch transformation lambdas with an empty string.

1. Reduce the size of the header and breadcrumb text in the shared, draft errors component (i.e., _src/components/shared/Entity/Error/DraftErrors.tsx_).

1. Consolidate all derivation create workflow helper functions into a single file, _src/utils/derivation-utils.ts_.

## Tests

## Issues

## Content

## Screenshots


src/components/shared/Entity/Actions/SaveButton.tsx
```diff
@@ -15,14 +15,15 @@ import { FormStatus } from 'stores/FormState/types';
...
     callFailed: Function;
-    disabled: boolean;
-    taskNames: string[];
     logEvent:
         | CustomEvents.CAPTURE_CREATE
+        | CustomEvents.COLLECTION_CREATE
         | CustomEvents.MATERIALIZATION_CREATE
         | CustomEvents.CAPTURE_EDIT
         | CustomEvents.MATERIALIZATION_EDIT;
+    disabled?: boolean;
```
kiahna-tucker, line 24: The `disabled` property was made optional because logic has not been defined to conditionally set it and I did not want to pass a fixed boolean _initially_. In retrospect, it does not make sense to make this property optional given it will likely be conditionally set in the near future.

src/components/transformation/create/DerivationEditor/index.tsx
```diff
@@ -96,16 +142,28 @@ function DerivationEditor() {
...
+                >
+                    <GitPodButton
+                        postWindowOpen={postWindowOpen}
+                        buttonVariant="outlined"
+                    />
 
-                <Button>Publish</Button>
+                    <EntitySaveButton
+                        callFailed={helpers.callFailed}
+                        taskNames={
```
kiahna-tucker, line 158: This condition _should_ be able to be removed since the default value of the `catalogName` transform create state property was changed from `null` to an empty string. Will update.

src/components/transformation/create/GitPodButton.tsx
```diff
@@ -66,30 +79,70 @@ function GitPodButton({
...
+                    catalog_name: catalogName,
+                    spec_type: 'collection',
+                });
+            } else if (sourceCollectionSet) {
+                const draft = await createEntityDraft(catalogName);
+
+                if (draft.error) {
+                    throw new Error(
+                        `[${draft.error.code}]: ${draft.error.message}, ${draft.error.details}, ${draft.error.hint}`
+                    );
```
travjenkins, line 113: Not for this PR - but I think we should handle the errors a bit different. Returning a string that contains all the props is nice but most of these are not truly usable. We should make this just handle errors the same way other create flows do.

src/components/transformation/create/GitPodButton.tsx
```diff
@@ -108,16 +161,24 @@ function GitPodButton({
...
-                const [token, draftId] = await Promise.all([
+                const [token, evaluatedDraftId] = await Promise.all([
                     createRefreshToken(false, '1 day'),
                     generateDraftWithSpecs(),
                 ]);
 
+                if (!evaluatedDraftId) {
+                    throw new Error(
+                        intl.formatMessage({
+                            id: 'newTransform.errors.urlNotGenerated',
```
kiahna-tucker, line 172: Since we refrain from exposing the concept of a draft to the user, I felt this error message could be reused in this scenario:

`We failed to generate the proper URL to start GitPod. Try again and if the issue persists please contact support.`

src/utils/derivation-utils.ts
travjenkins, line 1: Not for this PR: but noticed we are repeating some patterns of the "file names" and we should probably make those shared via some function that generates them. Also, could help keeping all those patterns together for if they ever need to change.

kiahna-tucker, line 1: There is quite _a lot_ going on in this file. I imagine there is a way to share code between these functions. Taking a second look now.

src/utils/derivation-utils.ts
```diff
@@ -1,44 +1,191 @@
...
+
+    switch (language) {
+        case 'typescript':
+            return generateTsTemplate(strippedEntityName, sourceCollections);
+        default:
+            return generateSqlTemplate(
+                strippedEntityName,
+                sourceCollections,
+                options
+            );
```
travjenkins, line 145: No change needed: I think as we get a more mature we need to look into having defaults vs throwing errors. Balancing between having verbose code and ensuring stuff does not "fail silently"

kiahna-tucker (commented):
Found a couple items to look into before this PR is merged.

travjenkins (approved):
Overall this is okay - one major change that needs to happen sometime soon is stop handling both `set` and `array` for the collections. We should just have a single solution instead of supporting multiple types.

PR #656, opened 2023-06-13
prod: update derivation and catalog name in transform create store

kiahna-tucker:
## Changes

1. Store the complete value of `PrefixedName` in the `catalogName` transform create store property.

1. Store the value of the input component used in `PrefixedName` in the `name` transform create store property (which corresponds to the derivation name, a subset of the catalog name).

## Tests


travjenkins (approved):
lgtm

PR #660, opened 2023-06-15
Trigger billing loading states when selected tenant changes

kiahna-tucker:
## Changes

1. Present the loading states for components on the _Billing_ page when the selected tenant changes.

1. Adjust the billing-related catalog stats queries so that data is fetched when a new billing cycle begins.

## Tests

* Validate that the components on the _Billing_ page enter a loading state when the page is initially loaded.

* Validate that the components on the _Billing_ page enter a loading state when the selected tenant changes.

**NOTE:** I am unable to generate catalog stats locally. The second scenario needs to be tested with at least one account that has catalog stats before this PR can be merged.

## Screenshots


travjenkins (approved):
lgtm

PR #665, opened 2023-06-19
Allow non-required props as keys and display prop type in selector

travjenkins:
## Changes

1. Allow the keys to use fields that are not required
2. Group the fields if they must/may exist
3. Show the field description
4. Show the field type
5. Update content

## Tests

Manually tested capture create/edit
Manually tested materialization create/edit

## Issues

Requested in Slack

## Content

Updates from conversations

## Screenshots

image


kiahna-tucker (approved):
Approved with moderate testing of capture workflows. I have some _gripes_ with some of the patterns below but my opposition is non-blocking.

Issue #663, opened 2023-06-19
Expired JWT token infinite loop messaging


travjenkins:
**Bug**
When a user gets logged out for an expired token they will get a constant loop where the "token expired" error keeps popping up over and over

**Expected**
They get logged out and see one (1) "token expired" error.

**Screenshots**
![image](https://github.com/estuary/ui/assets/270078/52d99bc4-08e7-4989-9d9f-02b3e016f459)

**Desktop (please complete the following information):**
N/A - doubt this is related

**Additional context**
N/A

travjenkins:
This might be due to multiple browser/tabs being opened while the token expires. Then, after expiring we try to log out users but there can be issues with that. Possibly related: https://github.com/orgs/supabase/discussions/889#discussioncomment-4856808

PR #671, opened 2023-06-26
Login / Logout experience tweaks

travjenkins:
## Changes

1. Add a new page `/logout` that will not try to log the user back in. So if a user is logged out the app will not try to immediately log back in.

## Tests

Manually tested logging in and out

## Issues

Progress on https://github.com/estuary/ui/issues/663

## Content

N/A

## Screenshots

N/A


kiahna-tucker (approved):
Approving with moderate testing in the interest of time. I tested the happy path of flows in the application that involve signing into and/or out of the application. Tests include but are not limited to the following:

* simple log in of an existing user

* simple log out of an existing user

* creation of a new tenant as a new user

* redemption of a tenant invitation

* sharing data between two tenants

I have not run through all of the code files yet.

PR #673, opened 2023-06-27
Create a page-based derivation create workflow

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/895

## Changes

1. Create a page-based derivation create workflow.

1. Add basic derivation schema validation. Erroneous scenarios include: the Flow-templated key (i.e., `your_key`) is present in the schema.

1. Add basic derivation database validation. Erroneous scenarios include: a transform lambda consisting of an empty string, a migration consisting of only an empty string.

1. Model the form section containing the transform and migration configuration after the _Output Collections_ section of existing task workflows.

1. Default the transform name to a _versioned_ variant of the evaluated source collection table name. For example, the name of a transform with the source collection `melk/waffles/events` would be `events`. A second transform with the source collection `melk/pancakes/events` would have the name `events_v1`.

1. Display the transform name on each list item (as opposed to the _faux_ filename).

1. Enable the creation of multiple transforms using the same source collection. Currently, one instance of a transform using a specific source collection can be added at a time.

1. Remove individual transforms (and migration) via a CTA on the list item.

1. Limit the number of migrations that can be created in the derivation create workflow to one.

1. Consolidate the Monaco editor component used in the SQL editor with that used elsewhere in the application.

1. Keep the _specification_ editor in sync with the _schema_ and _SQL_ editors. Valid changes made in the _schema_ and _SQL_ editors are reflected in the _specification_ editor. Valid changes made to the derivation schema in the _specification_ editor are reflected by the contents of the _Schema_ section.

**NOTE:** The layout of this page-based derivation create workflow is still knowingly in a transitional state. A core, future change will be partitioning the transformation and migration configuration views.

## Tests

## Content

## Screenshots

### DIFF
--- a/src/components/derivation/Create/CreateAlternate.tsx
+++ b/src/components/derivation/Create/CreateAlternate.tsx
@@ -16,6 +16,7 @@ import PatchDraftButton from 'components/transformation/create/PatchDraftButton'
 import DerivationSchema from 'components/transformation/create/Schema';
 import usePageTitle from 'hooks/usePageTitle';
 import { useNavigate } from 'react-router';
+import { useUnmount } from 'react-use';
 import { CustomEvents } from 'services/logrocket';
 import {
     useFormStateStore_error,
@@ -28,6 +29,7 @@ import { FormStatus } from 'stores/FormState/types';
 import {
     useTransformationCreate_catalogName,
     useTransformationCreate_emptySQLExists,
+    useTransformationCreate_resetState,
     useTransformationCreate_schemaUnedited,
 } from 'stores/TransformationCreate/hooks';
 
@@ -55,6 +57,7 @@ function DerivationCreateAlternate() {
     const catalogName = useTransformationCreate_catalogName();
     const emptySQLExists = useTransformationCreate_emptySQLExists();
     const schemaUnedited = useTransformationCreate_schemaUnedited();
+    const resetTransformationCreateState = useTransformationCreate_resetState();
 
     const helpers = {
         callFailed: (formState: any) => {
@@ -65,8 +68,6 @@ function DerivationCreateAlternate() {
             });
         },
         exit: () => {
-            resetEditorStore();
-            resetFormState();
             navigate(authenticatedRoutes.collections.fullPath);
         },
     };
@@ -82,6 +83,12 @@ function DerivationCreateAlternate() {
             }
         },
     };
+
+    useUnmount(() => {
+        resetEditorStore();
+        resetFormState();
+        resetTransformationCreateState();
+    });
 
     return (
         <DraftSpecEditorHydrator
--- a/src/components/derivation/Create/index.tsx
+++ b/src/components/derivation/Create/index.tsx
@@ -3,17 +3,17 @@ import { authenticatedRoutes } from 'app/routes';
 import AlertBox from 'components/shared/AlertBox';
 import DialogTitleWithClose from 'components/shared/Dialog/TitleWithClose';
 import TransformationCreate from 'components/transformation/create';
-import { LocalZustandProvider } from 'context/LocalZustand';
-import { useMemo, useState } from 'react';
+import { useState } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { useNavigate } from 'react-router-dom';
-import { TransformCreateStoreNames } from 'stores/names';
-import { createTransformationCreateStore } from 'stores/TransformationCreate/Store';
+import { useTransformationCreate_resetState } from 'stores/TransformationCreate/hooks';
 
 const ARIA_LABEL_ID = 'derivation-create-dialog';
 
 function DerivationCreate() {
     const navigate = useNavigate();
+
+    const resetTransformationCreateState = useTransformationCreate_resetState();
 
     // There is _probably_ a better way to do this, but the idea is
     // to reset the state of the NewCollection component every time
@@ -26,15 +26,8 @@ function DerivationCreate() {
         navigate(authenticatedRoutes.collections.fullPath);
         setShowConfirmation(false);
         setNewCollectionKey((k) => k + 1);
+        resetTransformationCreateState();
     };
-
-    const createLocalStore = useMemo(
-        () =>
-            createTransformationCreateStore(
-                TransformCreateStoreNames.TRANSFORM_CREATE
-            ),
-        []
-    );
 
     return (
         <Dialog
@@ -64,19 +57,17 @@ function DerivationCreate() {
                 </Collapse>
 
                 <Collapse in={!showConfirmation}>
-                    <LocalZustandProvider createStore={createLocalStore}>
-                        <TransformationCreate
-                            key={newCollectionKey}
-                            postWindowOpen={(gitPodWindow) => {
-                                // If there is a window object we know the browser at least let us open it up.
-                                //  This does not 100% prove that GitPod loaded correctly
-                                if (gitPodWindow) {
-                                    setShowConfirmation(true);
-                                }
-                            }}
-                            closeDialog={closeDialog}
-                        />
-                    </LocalZustandProvider>
+                    <TransformationCreate
+                        key={newCollectionKey}
+                        postWindowOpen={(gitPodWindow) => {
+                            // If there is a window object we know the browser at least let us open it up.
+                            //  This does not 100% prove that GitPod loaded correctly
+                            if (gitPodWindow) {
+                                setShowConfirmation(true);
+                            }
+                        }}
+                        closeDialog={closeDialog}
+                    />
                 </Collapse>
             </DialogContent>
         </Dialog>
--- a/src/stores/TransformationCreate/hooks.ts
+++ b/src/stores/TransformationCreate/hooks.ts
@@ -263,3 +263,10 @@ export const useTransformationCreate_setCatalogName = () => {
         (state) => state.setCatalogName
     );
 };
+
+export const useTransformationCreate_resetState = () => {
+    return useZustandStore<
+        TransformCreateState,
+        TransformCreateState['resetState']
+    >(TransformCreateStoreNames.TRANSFORM_CREATE, (state) => state.resetState);
+};


src/hooks/useDraftSpecEditor.ts
```diff
@@ -3,52 +3,73 @@ import { AllowedScopes } from 'components/editor/MonacoEditor/types';
...
+        async (
             newVal: any,
             catalogName: string,
-            specType: string,
-            propUpdating?: AllowedScopes
+            specType: Entity,
+            propUpdating?: AllowedScopes | string
         ) => {
-            if (draftSpec) {
+            if (mutate && draftSpec) {
```
travjenkins, line 62: Let's get this function converted over to a non-nesting approach.

src/components/editor/MonacoEditor/index.tsx
```diff
@@ -8,35 +15,40 @@ import {
...
 import { AllowedScopes } from './types';
 
-type onChange = (
+type EditorChangeHandler = (
     newVal: any,
     path: string,
-    specType: string,
-    scope?: AllowedScopes
+    specType: Entity,
+    scope?: AllowedScopes | string
```
travjenkins, line 38: Either the AllowedScopes or the string type needs removed. Allowing any string defeats the purpose of having a known list of scopes.

src/components/editor/MonacoEditor/index.tsx
```diff
@@ -222,23 +277,27 @@ function MonacoEditor({
...
+                        py: 0.5,
+                        px: 1,
+                        display: 'flex',
+                        justifyContent: 'space-between',
+                        alignItems: 'center',
                     }}
                 >
+                    <Typography sx={{ fontWeight: 500 }}>
+                        {editorLabel}
+                    </Typography>
```
travjenkins, line 295: If we're gonna label this we should eventually make this more like a fieldset or something for the editor. Not really sure the best way to make this accessible for Monaco.

src/components/editor/MonacoEditor/index.tsx
kiahna-tucker: This was something I planned to revisit before merging. The editor has gone through many iterations and I decided to make a final call once the component was in a stable state. Leaning towards removing `AllowedScopes` altogether.

src/components/transformation/create/Config/SQLEditor/useSQLEditor.ts
```diff
@@ -0,0 +1,142 @@
...
+            attributeId,
+            attributeType,
+            draftId,
+            draftSpec,
+            migrations,
+            mutateDraftSpecs,
+            transformConfigs,
+        ]
+    );
+
```
travjenkins, line 109: I think these effects are the same as in the `useDraftSpecEditor` if these are going to be combines in the future please add a comment stating that. Otherwise, please make 'em share this functionality.

src/components/derivation/Create/CreateAlternate.tsx
```diff
@@ -0,0 +1,135 @@
...
+            setFormState({
+                status: FormStatus.FAILED,
+                exitWhenLogsClose: false,
+                ...formState,
+            });
+        },
+        exit: () => {
+            resetEditorStore();
+            resetFormState();
+            navigate(authenticatedRoutes.collections.fullPath);
```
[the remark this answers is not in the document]
kiahna-tucker: Sure. Older workflows use the highlighted approach, so I followed suit with the understanding that the longstanding pattern would be updated at one time.

src/hooks/useDraftSpecEditor.ts
```diff
@@ -3,64 +3,86 @@ import { AllowedScopes } from 'components/editor/MonacoEditor/types';
...
-        localScope: useLocalScope,
+        localScope,
     });
     const draftId = useEditorStore_persistedDraftId();
 
-    // Fetch the draft specs for this draft
-    const { draftSpecs, isValidating, mutate } = useDraftSpecs(draftId, {
-        specType: entityType,
-        catalogName: entityName,
+    const draftSpecs = useEditorStore_queryResponse_draftSpecs({ localScope });
```
travjenkins, line 34: Not for this pr but with the query in the store already we should really stop duplicating these into the `specs` as it does not seem to solve any problems.

kiahna-tucker: As discussed, there will be a PR dedicated to overhauling the editor store. This will be taken care of at that time.

src/stores/TransformationCreate/Store.ts
```diff
@@ -1,33 +1,67 @@
...
-import { intersection } from 'lodash';
+import { intersection, omit } from 'lodash';
 import { TransformCreateStoreNames } from 'stores/names';
+import { hasLength } from 'utils/misc-utils';
 import { devtoolsOptions } from 'utils/store-utils';
 import { create, StoreApi } from 'zustand';
 import { devtools, NamedSet } from 'zustand/middleware';
 import { TransformCreateState } from './types';
 
+const evaluateSelectedAttribute = (
```
travjenkins, line 10: Please add some comments to this function to walk through what is happening in it.

src/stores/TransformationCreate/Store.ts
```diff
@@ -94,6 +128,14 @@ const getInitialState = (
                     };
                 });
 
+                state.emptySQLExists =
+                    Object.values(state.transformConfigs).some(
+                        ({ lambda }) => !hasLength(lambda)
+                    ) ||
+                    Object.values(state.migrations).some(
+                        (migration) => !hasLength(migration)
+                    );
```
travjenkins, line 137: These repeated checks need to move into a shared function.

src/stores/TransformationCreate/Store.ts
kiahna-tucker: I will not at present because the function will be removed shortly.

src/components/transformation/create/InitializeDraftButton.tsx
```diff
@@ -74,9 +79,28 @@ function InitializeDraftButton({
...
+                        const tableName = stripPathing(collection);
+
+                        if (Object.hasOwn(latestTransformVersions, tableName)) {
+                            latestTransformVersions[tableName] += 1;
+                        } else {
+                            latestTransformVersions = {
+                                ...latestTransformVersions,
+                                [tableName]: 0,
+                            };
+                        }
```
travjenkins, line 96: What is driving this approach? Is something wrong with just using the block notation for both parts of the if/else block? Then the object could become a `const`

```javascript
if (Object.hasOwn(latestTransformVersions, tableName)) {
    latestTransformVersions[tableName] += 1;
} else {
    latestTransformVersions[tableName] = 0;
}
```

src/components/transformation/create/InitializeDraftButton.tsx
```diff
@@ -101,7 +126,13 @@ function InitializeDraftButton({
...
                     setPersistedDraftId(draftId);
 
-                    setOpenSQLEditor(true);
+                    if (setSQLEditorOpen) {
+                        setSQLEditorOpen(true);
+                    }
+
+                    // TODO (transform): Replace this with the navigate to create workflow hook and the production-ready URL
+                    //   when it is time to launch this feature.
+                    navigate(authenticatedRoutes.beta.new.fullPath);
```
travjenkins, line 135: What are your thoughts on what this will look like when ready? Right now you get stuck when reloading on the `/new` path.

src/components/transformation/create/Schema/Header.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const schemaUnedited = useMemo(() => {
+        if (currentCatalog && Object.hasOwn(currentCatalog.spec, 'schema')) {
+            return Object.hasOwn(
+                currentCatalog.spec.schema.properties,
+                'your_key'
+            );
+        } else {
+            return false;
+        }
+    }, [currentCatalog]);
```
travjenkins, line 21: I am not sure exactly how I feel about this being in the head itself. Are you thinking this validation would eventually move somewhere else or stay here?

src/components/transformation/create/PatchDraftButton.tsx
```diff
@@ -0,0 +1,138 @@
...
+    // Transformation Create Store
+    const catalogName = useTransformationCreate_catalogName();
+    const setSelectedAttribute = useTransformationCreate_setSelectedAttribute();
+
+    const setCatalogUpdating = useTransformationCreate_setCatalogUpdating();
+
+    const transformConfigs = useTransformationCreate_transformConfigs();
+    const migrations = useTransformationCreate_migrations();
+
+    const updateDerivationSpec = useCallback(async (): Promise<void> => {
```
travjenkins, line 45: We don't need to change this now but before this goes live

This should use the non-nesting approach.

src/components/transformation/create/Config/Tabs.tsx
travjenkins, line 1: Just thinking - we should really make our own reusable component for tabs given how often we're doing this now.

src/components/transformation/create/Config/UpdateDraftButton.tsx
```diff
@@ -75,18 +78,25 @@ function UpdateDraftButton({ selectedCollections, setDialogOpen }: Props) {
...
                 draft_id: draftId,
                 catalog_name: catalogName,
             });
 
-            if (draftSpecResponse.error) {
+            if (draftSpecResponse.error || !mutateDraftSpecs) {
                 setCatalogUpdating(false);
                 // Set error state
             } else {
+                await mutateDraftSpecs();
```
travjenkins, line 92: Is this getting error handled somewhere else?

src/components/editor/LiveSpec.tsx
```diff
@@ -1,13 +1,26 @@
...
 
 function LiveSpecEditor({ localZustandScope }: Props) {
+    const currentCatalog = useEditorStore_currentCatalog({
+        localScope: localZustandScope,
+    });
+
+    const specAsString = useMemo(
+        () => stringifyJSON(currentCatalog?.spec ?? null),
+        [currentCatalog?.spec]
+    );
```
travjenkins, line 18: Not for this pr - we needs to make a specs utils kinda thing soon.

src/components/transformation/create/Schema/Header.tsx
kiahna-tucker: It should be in the store because the primary CTAs need to be disabled when the flag is true. Moving it now.

src/components/transformation/create/InitializeDraftButton.tsx
kiahna-tucker: It would function in a similar way to the edit workflows, using the drafted collection to prefill the form. The draft ID would be passed as a URL parameter as per usual.

src/components/editor/Store/create.ts
```diff
@@ -105,18 +124,58 @@ const getInitialState = <T>(
...
+                    }
+
+                    state.invalidEditors = invalidEditors.filter(
+                        (path) => path !== value
+                    );
+                }),
+                false,
+                'Stale Status Removed'
+            );
+        },
```
travjenkins, line 177: I feel like this function should not exist _in this way_. I feel like this update should be auto handled whenever `removeAttribute` is called.

src/components/editor/Store/create.ts
```diff
@@ -147,6 +217,9 @@ const getInitialState = <T>(
 
 export const createEditorStore = <T>(key: string) => {
     return create<EditorStoreState<T>>()(
-        devtools((set) => getInitialState<T>(set), devtoolsOptions(key))
+        devtools(
+            (set, get) => getInitialState<T>(set, get),
```
travjenkins, line 221: Thinking out loud - starting to think we should just always send `set` and `get` to all stores when starting 'em up and just mark them as unused.

kiahna-tucker: Agreed. That has been my position.

src/components/derivation/Create/CreateAlternate.tsx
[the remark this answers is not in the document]
travjenkins: I tackled updating a lot of the older ones and thought they were all converted. If there are ones hanging around that's a bug. Please open up a ticket for the ones you found.

src/components/transformation/create/InitializeDraftButton.tsx
travjenkins: Okay.. so I am imagining something like : `collections/new?draftid=...`

Would we need to write a guard that pushes the consumer to the previous page if the draft isn't there? Or just land 'em on a page with a message of "this draft no longer exists"?

src/components/editor/Store/create.ts
```diff
@@ -105,18 +124,58 @@ const getInitialState = <T>(
...
             set(
                 produce((state) => {
-                    state.isSaving = newVal === EditorStatus.SAVING;
-                    state.isEditing = newVal === EditorStatus.EDITING;
-                    state.status = newVal;
+                    if (Object.hasOwn(state.statuses, path)) {
+                        state.statuses[path] = value;
+                    } else {
+                        state.statuses = { ...state.statuses, [path]: value };
+                    }
```
travjenkins, line 134: Same question here as in elsewhere in the PR... not totally sure why this approach is being taken.

src/components/editor/Store/create.ts
kiahna-tucker: I feel the same way. This was something I would have discussed in dailies this past Monday. _Checkpoint comment. Need to pivot to something else..._

src/components/editor/Store/create.ts
```diff
@@ -105,18 +124,58 @@ const getInitialState = <T>(
...
+                    );
+
+                    state.invalidEditors = Object.entries(state.statuses)
+                        .filter(
+                            ([_key, status]) =>
+                                status === EditorStatus.INVALID ||
+                                status === EditorStatus.SAVE_FAILED ||
+                                status === EditorStatus.OUT_OF_SYNC
+                        )
+                        .map(([key, _status]) => key);
```
travjenkins, line 154: Thinking out loud - this one wouldn't be as easy to pull off but we should probably make some kind of reusable store helper/extension/something to handle status checking and make it work for singular or multiples like here.

src/components/editor/Store/types.ts
```diff
@@ -1,5 +1,15 @@
...
-import { DraftSpec } from 'hooks/useDraftSpecs';
+import { PostgrestError } from '@supabase/postgrest-js';
+import { SuccessResponse } from 'hooks/supabase-swr';
+import { DraftSpec, DraftSpecQuery } from 'hooks/useDraftSpecs';
+import { KeyedMutator } from 'swr';
+
+export interface DraftSpecResponse {
+    draftSpecs: DraftSpecQuery[];
+    isValidating: boolean;
+    mutate: KeyedMutator<SuccessResponse<DraftSpecQuery>> | null;
```
travjenkins, line 10: Not sure on this one yet... but given how often we have to check for this function I am thinking we should not make this null and rather just default it to a noop instead.

src/components/editor/Store/create.ts
kiahna-tucker: The answer is assistive technologies, my friend. _Hence_ the need for me to run through each file before merging, but I am not going to hold up a PR review for minor edits.

src/components/shared/Entity/CatalogEditor.tsx
```diff
@@ -23,7 +23,10 @@ function CatalogEditor({ messageId }: Props) {
     const formStatus = useFormStateStore_status();
     const formActive = useFormStateStore_isActive();
 
-    if (draftId && formStatus !== FormStatus.INIT) {
+    if (
+        draftId &&
+        (formStatus !== FormStatus.INIT || entityType === 'collection')
+    ) {
```
travjenkins, line 29: Please add a comment to call out why collections are skipped here.

src/components/schema/KeyAutoComplete/index.tsx
```diff
@@ -47,9 +47,10 @@ function KeyAutoComplete({ disabled, onChange, value }: Props) {
     const [localCopyValue, setLocalCopyValue] = useState<string[]>([]);
     const [inputValue, setInputValue] = useState('');
 
-    // We only want to all edit during Capture create/edit.
+    // We only want to all edit during capture create/edit and derivation create.
     const entityType = useEntityType();
-    const editKeyAllowed = entityType === 'capture';
+    const editKeyAllowed =
+        entityType === 'capture' || entityType === 'collection';
```
travjenkins, line 53: Thinking out loud - we should really look into "inverting" this pattern. All over the place we fetch out the type and then do some logic with it. I wonder if we made a context where you fetch what the entity type can do would be better. Have not put a lot of thought into this idea yet.

src/components/transformation/create/Config/SQLEditor/index.tsx
travjenkins, line 1: Thinking out loud - we might wanna take the time to make a reusable component that handles the loading / error / form / default kinda view. However, not really sure that would make life any easier as they are all _ever so slightly_ different.

src/components/transformation/create/Config/Header.tsx
```diff
@@ -0,0 +1,42 @@
...
+    // Transformation Create Store
+    const emptySQLExists = useTransformationCreate_emptySQLExists();
+    const shuffleKeyErrorsExist =
+        useTransformationCreate_shuffleKeyErrorsExist();
+
+    return (
+        <>
+            {emptySQLExists ||
+            shuffleKeyErrorsExist ||
+            invalidEditors.length > 0 ? (
```
travjenkins, line 25: Thinking... still need to decide on how to handle this situation. Since we only check the length we could just store off a simple "hasInvalidEditors" for this to reduce renders but not sure if the trade off is worth it. 

I think potentially that is the best approach once we get something wired into Zustand to handle computed values and then all those little "helper" store props are handled automagically for us.

kiahna-tucker: I was thinking the latter myself. Extending the store did not seem worthwhile because it is yet another related attribute to maintain that is (allegedly) relatively cheap.

src/components/transformation/create/Config/catalog/CatalogList.tsx
travjenkins, line 1: Is the thought this'll get used as the basis for the new catalog list / selectors / etc. or just making a custom one for transforms/migrations?

src/components/transformation/create/Config/catalog/CatalogList.tsx
```diff
@@ -0,0 +1,118 @@
...
+    addButtonClickHandler,
+    extendList = true,
+    height,
+}: Props) {
+    const theme = useTheme();
+
+    return (
+        <List
+            disablePadding
+            subheader={
```
travjenkins, line 45: Eventually the subheader should probably endup as a component

src/components/transformation/create/Config/catalog/CatalogListItem.tsx
travjenkins, line 1: I assume this'll also get broken up a bit and stuff shared.

src/components/transformation/create/Config/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+                </Typography>
+
+                <ListAndDetails
+                    list={<TransformList />}
+                    details={<DerivationCatalogEditor />}
+                    backgroundColor={
+                        alternativeReflexContainerBackground[theme.palette.mode]
+                    }
+                    displayBorder={true}
+                    height={550}
```
travjenkins, line 29: I see `550` is used in two places... can we throw that into a shared spot?

src/context/Theme.tsx
```diff
@@ -415,6 +415,11 @@ export const dataGridListStyling: SxProps<Theme> = {
     },
 };
 
+export const disabledButtonText = {
+    light: 'rgba(0, 0, 0, 0.26)',
+    dark: 'rgba(255, 255, 255, 0.3)',
+};
```
travjenkins, line 421: What is driving needing custom disabled styling that isn't applied to all buttons?

src/hooks/useInitializeCollectionDraft.ts
```diff
@@ -57,12 +57,15 @@ function useInitializeCollectionDraft() {
...
-    // Draft Editor Store
+    // Global Draft Editor Store
     const draftId = useEditorStore_persistedDraftId();
     const setDraftId = useEditorStore_setId();
 
     const setPersistedDraftId = useEditorStore_setPersistedDraftId();
 
+    // Local Draft Editor Store
+    const setLocalDraftId = useEditorStore_setId({ localScope: true });
+
```
travjenkins, line 68: Nothing to change... but this _feels_ weird to me... kind of like a code smell or something. Where we are fetching stuff for the global version of a store and then also fetching from the local copy. Cannot think of a reason why this is bad just feels wonky to me.

travjenkins (approved):
We have some stuff to discuss on nitty-gritty details but nothing that jumps out right now. I regression tested a tiny bit but mainly was playing around with functionality as I was looking over code. Will do some more testing tomorrow but seems okay for now.

src/components/transformation/create/InitializeDraftButton.tsx
kiahna-tucker: The answer to a related comment applies here:

> The answer: typescript code presets on assistive technologies. Hence, I perform an exhaustive review of my own code changes before merging the PR to catch any superfluous syntax introduced by the assistive technology.

src/components/derivation/Create/CreateAlternate.tsx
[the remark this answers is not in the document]
kiahna-tucker: Reference: [issue 684](https://github.com/estuary/ui/issues/684).

PR #674, opened 2023-06-29
Show evolution actions

psFried:
## Changes

This updates the handling of incompatible collections from failed publications to account for the new behavior of the evolutions and publications handlers. The agent's behavior around the evolution of collections has changed in a few ways recently, and this change brings the UI in line with that.  Previously, the evolutions handler would automatically choose between creating new collections or just new materialization bindings, based on the presense of the `x-infer-schema` annotation.  This was a flawed heuristic, and this commit sets things more or less straight (at least until we identify other problems).

The UI will now determine the specific action that's required in order to remediate a propsed change that's been deemed incompatible. It does this based on a new field that's been added to the `IncompatibleCollection`s in the publication `job_status`.  The new field is called `requires_recreation`, and its value is an array of `reason`s why a collection should be re-created. If that field is present and non-empty, then the UI will pass a `new_name` for that collection when creating the `evolutions` job.  If `requires_recreation` is empty, then no `new_name` will be set and the `evolutions` job will only update materialization bindings.

The end result can be somewhat more complicated for users to reason about, since some collections may need re-created while others may only need new materialization bindings. The wording of various evolution-related messages has been updated.  The UI now displays a proposed 'action' alongside each each incompatible collection. For collections that need re-created, there's also a reason.  Collections that _aren't_ being re-created could in theory also have a reason, but the error messages being returned by connectors don't currently seem appropriate for display there, so I left them out.

## Tests

I created captures and then manually edited the collection specs to trigger publication failures due to various types of incompatible changes. For example, changing the collection key or changing the `type` of a materialized field.

I also shut down the agent and manually pasted in some `job_status`es that had up to 100 entries in `incompatible_collections`, to make sure that performance is acceptable with many naughty collections (it is, at least on a super fancy m1 mac 🙃). I honestly don't expect to see very many `incompatible_collections` in practice, since even in databases with hundreds of tables it _seems_ unlikely to make incompatible changes to very many of them at once. But I'm calling that out as an assumption that could be proven wrong, which may motivate changing this to a virtualized list of some sort.

## Issues

estuary/flow#1063

## Content

yup, most of the evolutions-related content got updated. I ran it past @dyaffe on a screenshare, and he suggested changing `Apply evolution actions` to just `Apply`, so there's still some screenshots with the old wording because I'm lazy.

## Screenshots

Screen Shot 2023-06-29 at 4 37 47 PM

Screen Shot 2023-06-29 at 4 37 05 PM

Screen Shot 2023-06-29 at 12 36 32 PM

Screen Shot 2023-06-29 at 12 37 24 PM


src/lang/en-US.ts
```diff
@@ -1078,9 +1078,35 @@ const SchemaEditor_Collection: ResolvedIntlConfig['messages'] = {
...
+    'entityEvolution.error.title': `Changes rejected due to incompatible collection updates`,
+    'entityEvolution.error.message': `The proposed collection changes would break downstream tasks. You can click '${CTAs['cta.evolve']}' below to automatically update your draft with the following recommended actions.`,
+    'entityEvolution.error.note': `Note: This may result in additional cost as new versions are backfilled.`,
+
+    // Single quotes are special and must be doubled: https://formatjs.io/docs/core-concepts/icu-syntax#quoting--escaping
+    'entityEvolution.action.recreateOneBinding.description': `the Materialization ''{materializationName}'' will be updated to materialize the collection into a new resource`,
+    'entityEvolution.action.recreateBindings.description': `{materializationCount} {materializationCount, plural,
+        one {Materialization}
+        other {Materializations}
+    } will be updated to materialize the collection into new resources`,
```
travjenkins, line 1090: @psFried  - added support for plurals vs singular here. I know this is actually handled within the component but just trying to always make this a good habit so if anything changes later the messages won't get messed up.

Thee library has a lot of options but the main 3 we're trying to handle  are: `zero` (sometimes), `one`, and `other`. 
https://formatjs.io/docs/core-concepts/icu-syntax/#plural-format

src/components/shared/Entity/IncompatibleCollections/Description.tsx
travjenkins, line 1: @psFried - I pulled the description into a component. How you had it before was fine. Mainly because if variables this needs are the same the `CollectionsList` needs and so when one renders the other will as well... so keeping them together makes sense.

However, I felt the description stuff was _juuuuuust_ on the line of being big enough to be in a stand alone component. Also, I could really see this being used elsewhere (in like a history view where we explain what we updated).

src/components/shared/Entity/IncompatibleCollections/index.tsx
```diff
@@ -37,8 +39,9 @@ function IncompatibleCollections() {
                         </Typography>
                     </Box>
 
-                    <CollectionsList />
-
+                    <Box sx={{ maxHeight: 200, overflow: 'auto' }}>
+                        <CollectionsList />
+                    </Box>
```
travjenkins, line 44: @psFried - Added to make the list scroll when it got too long. I know we do not do this up above with errors but with the two sections sometimes displaying at the same time (at least when I tested) the page was getting a little ridiculously big. Really just a personal tweak.

travjenkins (approved):
LGTM. Tested out with my local postgres with 1k collections and messed with the keys of like 15 all in one go and it worked fine.

src/lang/en-US.ts
psFried: SGTM, thanks. And makes sense that we'd want to support plurals universally here, so we never have to wonder whether a message does so or not.

PR #683, opened 2023-07-12
Deps upgrade: a lot of them

travjenkins:
### Deps
Updated basically all dependencies that we can. The few that are left as because there is some non-trivial amount of effort to get them upgraded. Mainly JsonForms and MUI due to changes in how the DatePicker works.

This work can be handled in the near future.

### .npmrc
Removed the local copy of the file to make installs on GitHub Actions work better. Also, should help other devs from running into issues where this is overriding the file in the home directory

### Read Me
Cleaned up the read me a bit to make it more useful for the current state of the app

## Tests

Went through create and edit for entities
Viewed as many pages as possible
Able to run local builds

## Issues

[Dependabot](https://github.com/estuary/ui/security/dependabot)

## Content

N/A

## Screenshots

N/A


README.md
```diff
@@ -4,24 +4,29 @@ The web UI for Estuary Flow.
...
+[Github docs](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-npm-registry#authenticating-with-a-personal-access-token).
 
-For more details on UI commands please look at `build.md`.
+# How to start/run?
 
-# What is the status?
+Run: `npm start`
 
-The status very much a "work in progress"... or like one level above a proof of concept.
+The UI is built to hit our own instance of Supabase. That means you will need to have that running locally as well.
```
kiahna-tucker, line 25: > The UI is built to hit our own instance of Supabase.

Should the company be explicitly mentioned here, replacing the pronominal adjective _our_?

> That means you will need to have that running locally as well.

I think it would be helpful to provide a link to the local Supabase setup documentation here. I would consider this a candidate for the aforementioned dependency section, if you choose to go that route.

README.md
```diff
@@ -4,24 +4,29 @@ The web UI for Estuary Flow.
...
 
 # Building Blocks
 
-This project was started using [Create React App](https://create-react-app.dev/) and [added TypeScript](https://create-react-app.dev/docs/adding-typescript).
-
-As we are currently using [MUI](https://mui.com/core/) as our main component library so we used [their typescript template](https://github.com/mui-org/material-ui/tree/master/examples/create-react-app-with-typescript) as a reference for this project. We combined some of their tsconfig settings with those found from [typescript-cheatsheets](https://github.com/typescript-cheatsheets/react#troubleshooting-handbook-tsconfigjson).
+-   [MUI](https://mui.com/core/) for components
+-   [Zustand](https://github.com/pmndrs/zustand) for local state
+-   [JSON Forms](https://github.com/eclipsesource/jsonforms) for many forms
+-   [SWR](https://github.com/vercel/swr) for server calls
```
kiahna-tucker, line 32: Any reason why we would not want to mention Apache ECharts here?

src/components/tables/RowActions/DisableEnable/Button.tsx
```diff
@@ -7,7 +7,7 @@ import { specContainsDerivation } from 'utils/misc-utils';
 import { DisableEnableButtonProps } from './types';
 
 const updateShardDisabled = (draftSpec: any, enabling: boolean) => {
-    draftSpec.shards = draftSpec.shards ?? {};
+    draftSpec.shards ??= {};
```
kiahna-tucker, line 10: What in the magical operator is this? Haha I will definitely be revisiting documentation to see when this was added. This has to mean that `||=`  and `&&=` are valid operators as well... and maybe even oddballs such as `>>=` and `%=`. Wild.

kiahna-tucker (approved):
Approved with suggestions and moderate testing.

README.md
```diff
@@ -4,24 +4,29 @@ The web UI for Estuary Flow.
 
 # How to install?
 
-You'll need to make sure the UI ahas the depencies installed. For the UI this should be straight forward and running `npm install` in the UI root directory.
+Make sure Node and npm are installed
```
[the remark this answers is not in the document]
travjenkins: I was thinking the same... but skipped so I'll add

I think the only one is really estuary/flow and supabase. Is there anything else?

README.md
```diff
@@ -4,24 +4,29 @@ The web UI for Estuary Flow.
...
 
-You'll also need internet connectivity so you can reach our Supabase API.
+Run: `npm install`
 
-# How do I run the UI?
+## I got a 401 error for `@estuary/flow-web`.
```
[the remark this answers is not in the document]
travjenkins: I'll make it troubleshooting. Felt weird making it that with a single topic but makes it clear where to add future ones.

README.md
travjenkins: I had the company but removed as it felt weird to say "Estuary's instance" because I thought it made it feel like it was made to point to production only.

Added that list

README.md
travjenkins: That's fair... I mainly was picking ones that were really important and that one counts... I'll add.

src/components/tables/RowActions/DisableEnable/Button.tsx
travjenkins: Yup... I guess it has been long enough that these can be safely used now. I was holding off for awhile just to be safe but we do really stick to pretty modern browsers so very happy to have these :D .

PR #688, opened 2023-07-14
Remove unused database call

travjenkins:
## Changes

1. No longer make a call to liveSpecs that populates nothing and shows loading. This very slightly alters the UX but I think it feels better overall.
2. No longer pass in loading/skeleton. This can easily be handled by the parent if needed.

## Tests

1. Manually tested create and edit for all entity types. Mainly ran with network throttling on so I could make sure the slow loading experience was still okay.

## Issues

Part of the "Bindings Selector" work

## Content
N/A
## Screenshots
N/A


kiahna-tucker (approved):
Approved with moderate manual testing.

Issue #697, opened 2023-07-26
Some charts unable to render due to date issues


travjenkins:
**Bug**
A user was unable to see charts of usage for entities

**Expected**
They see the graph as usual

**Screenshots**
Timezone: india
![image](https://github.com/estuary/ui/assets/270078/45fc5e21-8b34-4fe3-9f77-3e11f88166a4)
![image](https://github.com/estuary/ui/assets/270078/b97ec399-f688-4256-ac5a-c486bb126a97)

Timezone: EDT
![image](https://github.com/estuary/ui/assets/270078/d131a81f-248d-44d0-ae55-2b732ca4ea57)
![image](https://github.com/estuary/ui/assets/270078/1f78b804-ec3d-4672-8258-3fa64cc7910a)

Reference for times lining up around midnight in India
![image](https://github.com/estuary/ui/assets/270078/3de46e6b-40c9-4919-8a45-d9c5fde5bcf3)

```
...the graph library loaded, but instead of times on the X axis, there was “Invalid Data” or “Invalid Date” in place of the times
```

**Desktop (please complete the following information):**
Doubt this matters but will still look into it

**Additional context**
The user was working from India... thinking this might be a date/internationalization issue.

travjenkins:
The "indian timezone" is offset by 30 minutes so none of the data lines up. Looking into how to handle this.

PR #699, opened 2023-07-26
Fix hourly graphs for timezones

travjenkins:
## Changes

The hourly chart now uses the dataset approach and uses the timestamp the server returns as the main category. It is always stored, sorted, etc. in GMT and only formatted when displayed to the user.

Update the markline label to not clash

## Tests

Tested with timezones in New York, India, Australia

## Issues

Quick fix for https://github.com/estuary/ui/issues/697

## Content

## Screenshots

We will now display all the data we have. If some is missing nothing will show. This enables us to later more easily add a slider
image

Full six hours
image

Kolkuta, India
image

Broken Hill, Australia
image

Updated the markline to not clash as bad
image

image

### DIFF
--- a/src/components/graphs/DataByHourGraph/index.tsx
+++ b/src/components/graphs/DataByHourGraph/index.tsx
@@ -1,5 +1,6 @@
 import { useTheme } from '@mui/material';
 import { defaultOutlineColor, eChartsColors } from 'context/Theme';
+import { format, sub } from 'date-fns';
 import { EChartsOption } from 'echarts';
 import { BarChart } from 'echarts/charts';
 import {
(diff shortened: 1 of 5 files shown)


src/components/graphs/DataByHourGraph/index.tsx
```diff
@@ -118,26 +96,161 @@ function DataByHourGraph({ range, stats = [] }: Props) {
...
+
         const option: EChartsOption = {
             animation: false,
             darkMode: theme.palette.mode === 'dark',
             legend: legendConfig,
+            series: [bytesSeries, docsSeries],
+            useUTC: true,
+            // Setting dataset here because setting in a stand alone set option cause the chart to go blank
+            dataset: {
+                dimensions: ['timestamp', 'bytes', 'docs'],
```
kiahna-tucker, line 194: These strings are repeated throughout the file. Could we store them off in a constant for ease of reference (and maintenance)?

src/components/graphs/DataByHourGraph/index.tsx
```diff
@@ -200,127 +313,17 @@ function DataByHourGraph({ range, stats = [] }: Props) {
...
-            docsSeries.data?.push({ itemStyle, value: docs });
-        });
-
-        // Update mychart series so new data goes in
-        myChart?.setOption({
-            series: [bytesSeries, docsSeries],
-        });
-    }, [hours, intl, myChart, scopedDataSet]);
-
     return <div id="data-by-hour" style={{ height: 350 }} />;
```
kiahna-tucker, line 327: This chart container ID should also be stored in an constant for ease of reference (and maintenance).

kiahna-tucker (approved):

PR #706, opened 2023-07-31
Field selection UX improvements

kiahna-tucker:
## Changes

1. Correct the field selection change debouncing logic.

1. Make the fields column of the field selection table "sticky" so that it can be referenced when the table is horizontally scrolled.

1. Extend the set of form state statuses with an _updating_ and _updated_ status for asynchronous processes.

1. Set the icons used in the MUI checkbox component in the theme.

1. Add a CTA to the field selection section that initiates a dry run publication to populate the field selection table. The dry run publication logs dialog closes immediately once the process is complete.

## Tests

## Issues

## Content

## Screenshots

fs1

fs2

fs3

fs5


src/components/shared/Entity/Actions/Save.tsx
```diff
@@ -142,10 +144,10 @@ function EntityCreateSave({ disabled, dryRun, onFailure, logEvent }: Props) {
             async (payload: any) => {
                 trackEvent(logEvent, payload);
 
-                const imcompatibleCollections =
+                const incompatibleCollections =
```
kiahna-tucker, line 147: Fixed a typo in the word _incompatible_.

src/context/Theme.tsx
```diff
@@ -70,9 +72,15 @@ export const sample_grey = {
...
-export type SemanticColor = '#40B763' | '#F5D75E' | '#CA3B55' | '#4FD6FF';
+export type SemanticColor =
+    | '#40B763'
+    | '#2A7942'
+    | '#F5D75E'
+    | '#CA3B55'
+    | '#4FD6FF';
 
 export const successMain: SemanticColor = '#40B763';
+export const successDark: SemanticColor = '#2A7942';
```
kiahna-tucker, line 83: The _main_ semantic color used to indicate success does not meet the AA level contrast. Even when adjusted in accord with the contrast threshold (i.e., the calculated `theme.palette.success.dark`), the contrast was still unsatisfactory. Therefore, the dark variant of the "success" semantic color was explicitly defined, using the _main_ variant as a base.

src/components/editor/Bindings/FieldSelection/useFieldSelection.ts
travjenkins, line 1: When a user makes a single change this hook is calling mutate twice

kiahna-tucker:
> We should probably disable the button if the field must be included. Otherwise you can click on it and cause calls to be made.

The _include_ button should be disabled only when the recommended flag is not set (i.e., `false`) and the field is required. The disabled state of an active toggle state is a little hard to distinguish from that of an inactive toggle button. It is in accord with MUI's disabled button styling but I cannot imagine someone with less than optimal eyesight could distinguish which option is selected.

delete

src/components/capture/RediscoverButton.tsx
```diff
@@ -1,21 +1,18 @@
...
 }
 
-function RediscoverButton({
-    entityType,
-    disabled,
-    callFailed,
-    postGenerateMutate,
-}: Props) {
+function RediscoverButton({ entityType, disabled, postGenerateMutate }: Props) {
+    const { callFailed } = useEntityWorkflowHelpers();
```
travjenkins, line 14: We fetch this same function for both here and the `generate` button. We should just have the `useDiscoverCapture` hook pull this out and not have it passed.

src/components/shared/Entity/hooks/useEntityWorkflowHelpers.ts
travjenkins, line 1: To be safe we should wrap these in `callbacks`

src/components/editor/Bindings/FieldSelection/index.tsx
```diff
@@ -87,7 +94,7 @@ const mapConstraintsToProjections = (
         };
     });
 
-function FieldSelectionViewer({ collectionName }: Props) {
+function FieldSelectionViewer({ collectionName, connectorsExist }: Props) {
```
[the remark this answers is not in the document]
kiahna-tucker: Ah, this was something I intended to bring up yesterday. At least one connector must exist in order to render the resource configuration form section, therefore I think it is safe to set the `disabled` property of the `EntityTestButton` component instance explicitly to `false`. This would remove the need for the highlighted `connectorsExist` property entirely. How does that sound to you?

src/components/editor/Bindings/FieldSelection/useFieldSelection.ts
travjenkins, line 1: The main update function wipes out `fields` if it already existed. I remember that we were okay with something like that from a previous conversation but I cannot fully recall if it was this one. Just wanted to double triple check.

src/components/tables/cells/fieldSelection/FieldActions.tsx
```diff
@@ -20,57 +24,99 @@ interface Props {
...
+                            selectedValue !== 'include' || includeRequired
+                                ? 'include'
+                                : null;
+
+                        setSingleSelection(
+                            field,
+                            selectedValue === 'include' && recommendFields
+                                ? 'default'
+                                : singleValue
+                        );
```
travjenkins, line 97: This chunk that is duplicated should be shared.

src/components/tables/cells/fieldSelection/OutlinedToggleButton.tsx
```diff
@@ -17,39 +24,64 @@ function OutlinedToggleButton({
...
     onChange,
     onClick,
 }: Props) {
-    const selectionSaving = useBindingsEditorStore_selectionSaving();
+    const defaultStateSx: SxProps<Theme> = coloredDefaultState
+        ? {
+              border: (theme) => primaryColoredOutline[theme.palette.mode],
+              color: (theme) => theme.palette.primary.main,
+          }
+        : {};
```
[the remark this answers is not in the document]
kiahna-tucker: > There is logic populating the sx... where one is at the top and the other ones down below are done inline

Clarifying statement from an external thread.

src/utils/workflow-utils.ts
```diff
@@ -156,3 +157,24 @@ export const modifyExistingCaptureDraftSpec = async (
...
+export const evaluateRecommendedIncludedFields = (
+    constraintType: ConstraintTypes
+): boolean => {
+    return (
+        constraintType === ConstraintTypes.FIELD_REQUIRED ||
+        constraintType === ConstraintTypes.LOCATION_REQUIRED ||
+        constraintType === ConstraintTypes.LOCATION_RECOMMENDED ||
+        constraintType === ConstraintTypes.FIELD_OPTIONAL
+    );
+};
```
travjenkins, line 180: If the `contraintType` checks are pretty locked down then I think we should was have `evaluateRecommendedIncludedFields` call `evaluateRequiredIncludedFields` to cover the field and location checks. That way they are always in sync.

travjenkins (approved):
Nit picky stuff only. The major call out about duplicate calls is the only worry but that can wait until we're ready to launch this feature.

src/components/editor/Bindings/FieldSelection/index.tsx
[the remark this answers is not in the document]
travjenkins: That sounds okay. The thing that confused me while skimming the code was the name... I was a bit confused why field selection cared about connectors for the reason you state.

So I guess either add a comment or update the name within here.

PR #707, opened 2023-08-01
Quick compare added to history

travjenkins:
## Changes

Adding a quick spec compare to the history view

## Tests

Manually tested

## Issues

Request made during offsite

## Content

N/A

## Screenshots

image


kiahna-tucker (approved):
Approved without manual testing. Reserving any visual callouts because this seems satisfactory to a broader audience as-is.

Issue #666, opened 2023-06-20
Disable bindings instead of removing them


psFried:
This relates to the ongoing work on automated schema evolution described in estuary/flow#1063. estuary/flow#1077 introduces the ability to mark individual capture and materialization bindings as disabled. It's important that we represent bindings as being disabled because otherwise an automated re-discover operation might re-add bindings that a user had previously deselected.

The desired behavior is for the UI to always mark bindings as disabled when a user deselects them. The UI should always leave all bindings in place in the spec. To disable a capture binding, set its `target` to an explicit `null` value (the field is still required). To disable a materialization binding, its `resource` should be set to an explicit `null` value.

Issue #536, opened 2023-03-30
Removing all Bindings then adding one removes config options


dyaffe:
**Bug**
Removing all Bindings then adding one removes config options

**Fix**
We should retain config options when a user removes all bindings then adds one.

Issue #447, opened 2023-01-16
Binding selector removes collections after adding a new one


dyaffe:
**Bug**
Editing a materialization then attempting to just add a binding removes all other bindings in the UI.

**Expected**
Adding a binding should just add that binding.

**Screenshots**
Include a screenshot of what you saw when the bug happened.

![Screenshot 2023-01-16 at 12 10 26 PM](https://user-images.githubusercontent.com/8095717/212733633-e51d975c-6b04-43cc-bb35-7b368291f026.png)
![Screenshot 2023-01-16 at 12 10 35 PM](https://user-images.githubusercontent.com/8095717/212733637-47079e8e-5732-4638-a643-307d11f3d419.png)

travjenkins:
Going to dig into this. It is really weird - I tested this right away when this was opened and saw it happen but now having issues recreating it.

travjenkins:
Cool... so this happens when an entity has a binding that is a collection that is old enough that it is not returned as part of the first 1k collections we get back.

I think we'll need to fix this along with https://github.com/estuary/ui/issues/358

PR #716, opened 2023-08-11
New bindings editor

travjenkins:
### Create/Edit flows
1. Setting server update flag so that next button does not need to be clicked twice
1. Implemented new bindings search and selector to fix several major defects
1. Lots of refactoring and store performance related changes
1. Added numbers to the section titles hoping users will follow them and fill out all sections

### Create/Edit Materializations
1. Added in warning message on the log dialogs if there are no bindings
1. Hide field selection when binding is disabled as you cannot pick fields
1. Updated the query that fetches details for collections from a wide `select *` to just selecting the single field we actually use.

### Tables
1. Added rowsPerPage as a setting to pass to entity table
1. Added a reset function for entity table queries

### New Create Transform
1. Got the new transform create close-ish but did not complete integrating with latest selector

### General
1. Updated to the latest data grid version and made updates to fixing breaking changes
1. Testing out a new pattern of directly using `useStore` instead of having everything in a hooks file

## Tests

Manually tested

## Issues

Fixes https://github.com/estuary/ui/issues/666
Fixes https://github.com/estuary/ui/issues/536
Fixes https://github.com/estuary/ui/issues/447
Fixes https://github.com/estuary/ui/issues/358

## Content

## Screenshots


src/components/transformation/create/legacy/LanguageSelector.tsx
```diff
@@ -22,7 +22,7 @@ function LegacyLanguageSelector() {
     const setLanguage = useTransformationCreate_setLanguage();
 
     return (
-        <LegacyStepWrapper last>
+        <LegacyStepWrapper last="true">
```
travjenkins, line 25: Fixes dom related error as `last` was actually added to the dom and mui was complaining.

src/hooks/useServerUpdateRequiredMonitor.ts
```diff
@@ -0,0 +1,66 @@
...
+    useResourceConfig_setServerUpdateRequired,
+} from 'stores/ResourceConfig/hooks';
+import { hasLength } from 'utils/misc-utils';
+import {
+    getCollectionName,
+    getCollectionNameProp,
+    getDisableProps,
+} from 'utils/workflow-utils';
+
+const useServerUpdateRequiredMonitor = (draftSpecs: DraftSpecQuery[]) => {
```
travjenkins, line 16: Moved into a shared hook as a couple places needed this and needed to make sure that this was always getting set the same way.

kiahna-tucker (approved):
Approved with manual testing in the interest of time, given this has been reviewed by the team holistically. There are a handful of call-outs that can be addressed as fast-follows and I will layer in review comments today.

PR #718, opened 2023-08-13
Expose capture auto discovery settings

kiahna-tucker:
## Changes

1. Allow the `autoDiscover` property of a capture specification to be set in the _Endpoint Config_ section of capture workflows.

## Tests

## Issues

## Content

## Screenshots

### DIFF
--- a/src/stores/EndpointConfig/Store.ts
+++ b/src/stores/EndpointConfig/Store.ts
@@ -20,8 +20,8 @@ import { JsonFormsData, Schema } from 'types';
 import { hasLength } from 'utils/misc-utils';
 import { parseEncryptedEndpointConfig } from 'utils/sops-utils';
 import { devtoolsOptions } from 'utils/store-utils';
-import { create, StoreApi } from 'zustand';
-import { devtools, NamedSet } from 'zustand/middleware';
+import { StoreApi, create } from 'zustand';
+import { NamedSet, devtools } from 'zustand/middleware';
 import { EndpointConfigState } from './types';
 
 const STORE_KEY = 'Endpoint Config';
@@ -44,29 +44,33 @@ const populateErrors = (
 
 const getInitialStateData = (): Pick<
     EndpointConfigState,
+    | 'addNewBindings'
     | 'encryptedEndpointConfig'
+    | 'endpointCanBeEmpty'
     | 'endpointConfig'
     | 'errorsExist'
     | 'endpointConfigErrors'
     | 'endpointSchema'
+    | 'evolveIncompatibleCollections'
     | 'hydrated'
     | 'hydrationErrorsExist'
     | 'previousEndpointConfig'
     | 'publishedEndpointConfig'
     | 'serverUpdateRequired'
-    | 'endpointCanBeEmpty'
 > => ({
+    addNewBindings: false,
     encryptedEndpointConfig: { data: {}, errors: [] },
+    endpointCanBeEmpty: false,
     endpointConfig: { data: {}, errors: [] },
     errorsExist: true,
     endpointConfigErrors: [],
     endpointSchema: {},
+    evolveIncompatibleCollections: false,
     hydrated: false,
     hydrationErrorsExist: false,
     previousEndpointConfig: { data: {}, errors: [] },
     publishedEndpointConfig: { data: {}, errors: [] },
     serverUpdateRequired: false,
-    endpointCanBeEmpty: false,
 });
 
 const getInitialState = (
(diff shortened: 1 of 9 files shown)


psFried:
Just based on the description and screenshots, this looks directionally right to me. One detail I'd like to call out is that, technically, we don't have to set `addNewBindings` whenever a user opts in to schema evolution. It's perfectly valid to set `autoDiscover: {}`, which would opt in to keeping existing bindings up-to-date without automatically adding new ones.

I'll suggest adding another check box called "Add new bindings", and grouping it with "Include breaking changes" to only show it when evolution is enabled.

src/stores/EndpointConfig/Store.ts
```diff
@@ -245,6 +245,28 @@ const getInitialState = (
...
+                //         ) && autoDiscoverySettings.addNewBindings
+                //     );
+
+                //     setEvolveIncompatibleCollections(
+                //         Object.hasOwn(
+                //             autoDiscoverySettings,
+                //             'evolveIncompatibleCollections'
+                //         ) && autoDiscoverySettings.evolveIncompatibleCollections
+                //     );
+                // }
```
travjenkins, line 268: Please add a comment explaining the plan for this code

kiahna-tucker: No, it clearly needs to be removed.

src/components/capture/AutoDiscoverySettings.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+            resizeObserver.observe(containerEl);
+        }
+
+        return () => {
+            if (resizeObserver !== null && containerEl) {
+                resizeObserver.unobserve(containerEl);
+            }
+        };
+    }, [setDisplayDirection, containerEl]);
```
travjenkins, line 173: Why are we handling this? Was there a requirement that these display in this way?

kiahna-tucker: It was discussed in the running thread. I can let the options scrunch into themselves if you prefer.

src/components/capture/useAutoDiscovery.ts
```diff
@@ -0,0 +1,90 @@
...
+    const setSettingsActive = useSchemaEvolution_setSettingsActive();
+    const setSettingsSaving = useSchemaEvolution_setSettingsSaving();
+
+    // TODO (schema evolution): Determine a comfortable debounce interval. A second or less feels too quick
+    //   but five seconds feels too long.
+    const debouncedUpdate = useRef(
+        debounce(() => {
+            setSettingsActive(false);
+            setSettingsSaving(true);
+        }, 3000)
```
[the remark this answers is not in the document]
kiahna-tucker: Cool. Consider it done.

src/components/capture/AutoDiscoverySettings.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+            const bindingSetting: boolean =
+                Object.hasOwn(autoDiscoverySettings, 'addNewBindings') &&
+                autoDiscoverySettings.addNewBindings;
+
+            const evolutionSetting: boolean =
+                Object.hasOwn(
+                    autoDiscoverySettings,
+                    'evolveIncompatibleCollections'
+                ) && autoDiscoverySettings.evolveIncompatibleCollections;
```
travjenkins, line 80: These should be replaced with a usage of the `Boolean` constructor.

kiahna-tucker: As you wish

src/components/capture/AutoDiscoverySettings.tsx
```diff
@@ -0,0 +1,224 @@
...
+    const settingsSaving = useSchemaEvolution_settingsSaving();
+    const setSettingsSaving = useSchemaEvolution_setSettingsSaving();
+
+    const updateAutoDiscoverySettings = useAutoDiscovery();
+
+    const settingsExist = useMemo(
+        () =>
+            draftSpecs.length > 0 &&
+            Object.hasOwn(draftSpecs[0].spec, 'autoDiscover') &&
+            isObject(draftSpecs[0].spec.autoDiscover),
```
travjenkins, line 62: `isObject` should handle a missing object fine so we should remove the `hasOwn' check and just use `isObject` with an optional chain.

src/components/capture/AutoDiscoverySettings.tsx
```diff
@@ -0,0 +1,226 @@
...
+    useEffect(() => {
+        setAutoDiscover(settingsExist, { initOnly: true });
+
+        if (settingsExist) {
+            const autoDiscoverySettings = draftSpecs[0].spec.autoDiscover;
+
+            const bindingSetting = Boolean(
+                Object.hasOwn(autoDiscoverySettings, 'addNewBindings') &&
+                    autoDiscoverySettings.addNewBindings
+            );
```
travjenkins, line 73: Sorry - I meant just calling the constructor with the object.

```
const bindingSetting = Boolean(autoDiscoverySettings?.addNewBindings);
```

travjenkins (approved):
LGTM - only concern is that the switches are displayed while stuff is still loading in... we might want to hide or disable those as the specs are still loading in.

PR #723, opened 2023-08-18
Display complete billing history for all tenants in table

kiahna-tucker:
## Changes

The billing history table on the _Billing_ page should display the four, most recent billing records. This PR corrects the defect preventing tenants that have existed for less than four billing cycles from seeing all available records reflected in the table.

## Tests

1. Validate that all available billing records appear on the billing history table for an account that has existed for less than four billing cycles.

1. Validate that all available billing records appear on the billing history table for an account that has existed for four billing cycles.

1. Validate that all available billing records appear on the billing history table for an account that has existed for more than four billing cycles.

## Issues


travjenkins (approved):
lgtm

PR #724, opened 2023-08-22
Quick fix for incorrect entity name when pasted in... again

travjenkins:
## Changes

1. Removing the use of a helper function as that messes up the state. This is a temp fix as we need to properly add computed values ASAP.

## Tests

Manual

## Issues

Caught in a slack channel again


kiahna-tucker (approved):
There are a couple defects encountered when you change the capture name. Looking into whether they are introduced by this PR (unlikely), exacerbated by this PR (potentially, given the new delay), or existed in production unnoticed.

Update: This is a defect in production. `useDraftSpecs` should have all its input parameters as dependencies, at the very least the catalog name should be added to the dependency array.

PR #727, opened 2023-08-25
Use a time range to fetch monthly catalog stats

kiahna-tucker:
## Changes

Use a time range to fetch monthly catalog stats shown in the entity tables.

## Tests

### DIFF
--- a/src/api/stats.ts
+++ b/src/api/stats.ts
@@ -1,5 +1,6 @@
 import { PostgrestFilterBuilder } from '@supabase/postgrest-js';
 import {
+    addMonths,
     endOfWeek,
     startOfHour,
     startOfMonth,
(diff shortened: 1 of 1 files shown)


src/api/stats.ts
```diff
@@ -143,13 +145,15 @@ const getStatsByName = (names: string[], filter?: StatsFilter) => {
         // Month Range
         case 'thisMonth':
             queryBuilder = queryBuilder
-                .eq('ts', formatToGMT(startOfMonth(today)))
+                .gte('ts', formatToGMT(startOfMonth(today)))
```
jgraettinger, line 148: this is kinda smelly, I don't know why a time range would be required.

For reference, here's how our reporting is mapping a `Date` into truncated time points: https://github.com/estuary/flow/blob/master/ops-catalog/catalog-stats.ts#L54-L79

PR #728, opened 2023-08-25
New bindings editor: UX 1

travjenkins:
## Changes

1. Sort the collections by published date like we do today
2. Wrap the entity names so the table does not scroll

## Tests

Manually tested materialization create

## Issues

Noticed during production validation

## Content

N/A

## Screenshots

Default table state when an entity name is long
image

Default table state when all entity names are short
image


kiahna-tucker (approved):
Approved without manual testing

PR #730, opened 2023-08-25
Better handling of details entity name custom errors and access token issues

travjenkins:
### Entity name custom errors
1. Move custom errors setting into the store
2. Do not show forms until they are hydrated. This was causing issues where the store was in the midst of hydrating/defaulting AND the form was defaulting values.

### Access token issues
1. When accessing login check if there is a grant token and ensure we go to welcome
2. Show a notification if a token was skipped 
3. Minor refactoring for the grant guard

## Tests

Manually tested create/edit of Captures and Materializations

## Issues

Noticed in Slack

## Content

One message is now hardcoded in the store

## Screenshots

user visiting a single use URL that was already submitted AND they did not have access (no change)
image

When an existing user attempts to get access a second time (new)
image


kiahna-tucker (approved):

kiahna-tucker (approved):
Approved in the interest of time.

Issue #735, opened 2023-09-07
Calls to patch the draft specification due to field selection changes can be cancelled


dyaffe:
**Bug**
Twice in the past week, I've seen [this](https://app.logrocket.com/tue20y/estuarytest/s/5-49d731c9-919d-4868-b37b-33ba0ab0b52a/47?t=1694091869136.2302) happen when adding exclusions to fields in materializations.  

Note: 
1. The modal goes away before saving
2. The actual publication doesn't happen or at minimum the exclusion doesn't take effect

It did eventually work when I updated it through the "Advanced Specification Editor".

cc @kiahna-tucker

kiahna-tucker:
I'm hesitant to call this a field selection bug because the functionality is still intact. The debounce interval for the clicks of the field selection action buttons is too long in practice. In the linked session, you clicked a field selection action button and then clicked Save & Publish within the debounce interval, ultimately cancelling the call to patch the specification. Shortening the interval should resolve this issue with some impact to the UX.

dyaffe:
Interesting -- what is our current "debounce interval" and what would the negative effects on UX be for shortening it?

I guess our UI doesn't generally lend itself to quick actions and that's what we're seeing here.  If you want to go in and do something fast like exclude a single field, it would make sense that you proceed quickly, but most other tasks are more slow?

dyaffe:
Would it be possible to allow someone to make a bunch of changes and only change the form when they click "Save and Publish" or "Test"?  I'm sure there are reasons that is a bad idea and you've already thought them through but just asking.

travjenkins:
@dave -we could and that is how we handle stuff like endpoint config but it does lead to some annoying state management issues in the UI.

If we can keep calling the db then if the user reloads or something the work is not lost. Also, in the future if we allow folks to pick up where they left off calling the DB is basically the only solution there.

PR #738, opened 2023-09-11
Reduce field selection debounce interval

kiahna-tucker:
## Issues

#735 

## Changes

Reduce the field selection `debounce` interval to ensure the drafted specification is patched.

## Tests

* Click a field selection action button and confirm that a competing call cannot be initiated that would result in the cancellation of patching the drafted specification.


travjenkins (approved):
lgtm

PR #739, opened 2023-09-11
Distinguish triggers for dry run publication log dialogs

kiahna-tucker:
## Issues

#712 

## Changes

1. Create a component for the _See Fields_ CTA that is fashioned after the `EntityTestButton` component.

1. Extend the shared, publication button component props (i.e., `EntityCreateSave`) to include a flag which will force the log dialog closed when set. This flag only affects the behavior of the log dialog opened when a dry run publication is initiated.

1. Add the requested header content to the language file for future use.

**NOTE:** Some refactoring is needed to make the log dialog more composable. Presently, the _See Fields_ CTA makes use of the log dialog instance in the `EntityTestButton` component since it is opened when the form status is _testing_ or _tested_.

## Tests

* Validate that the behavior and content of the log dialog opened by the _Save & Publish_ CTA are unchanged.

* Validate that the log dialog opened by the _Test_ CTA must be explicitly closed by the user.

* Validate that the log dialog opened by the _See Fields_ CTA automatically closes when the dry run publication succeeds.


travjenkins (approved):
The code is fine with me but all the dialogs close right away when there is an error. Which I think is right since we show the error and logs inline. 

However, I see the comment that test should only close when the user says so... want to make sure that is only true for a successful run

Other than that LGTM

kiahna-tucker:
> The code is fine with me but all the dialogs close right away when there is an error. Which I think is right since we show the error and logs inline.

If memory serves, this behavior existed since the origin of the log dialog and this PR was not trying to change that. Essentially, the log dialog opened when the _See Fields_ CTA is clicked should automatically close when the call concludes. The condition mentioned in the PR description was what was explicitly stated when the behavior was requested. I merely copied and pasted it from my notes from that meeting.

Issue #747, opened 2023-09-14
Key not updating properly when schema is edited


travjenkins:
**Bug**
When editing the schema the key list does not update properly

**Expected**
The keys update properly

**Screenshots**
![image](https://github.com/estuary/ui/assets/270078/9a09fecb-6454-4b64-b8e1-99b2cae7f7a6)

PR #748, opened 2023-09-14
Correct schema editing

travjenkins:
## Issues

Fixes https://github.com/estuary/ui/issues/747

### 747
- Do NOT directly update the state object
- Used a ref so we can compare without causing renders
- Only update spec when things have actually changed

### Misc
- Show a loading page when the inferred schema is being processed
- Disable schema edit button while form is active

## Tests

Manually tested

## Content

N/A

## Screenshots

image


kiahna-tucker (approved):

Issue #749, opened 2023-09-14
Remove UI cleaning up disabled collections during create


travjenkins:
The control plane will now clean up collections that are disabled. The UI no longer needs to clean these up.

PR #750, opened 2023-09-14
Removing the UI clean up of Collections

travjenkins:
## Issues

Fixes https://github.com/estuary/ui/issues/749

### 749
- Revert the code back to how it was before and no longer clean up disabled collections

## Tests

Manually tested capture create

## Content

N/A

## Screenshots

N/A


kiahna-tucker (approved):

Issue #756, opened 2023-09-20
Refreshing collections loses recent changes to bindings


travjenkins:
**Bug**
![image](https://github.com/estuary/ui/assets/270078/7b181407-8839-409b-bc1d-d482c1a666ce)

**Expected**
The recent changes to bindings are not overwritten

PR #757, opened 2023-09-20
Let refresh button update draft before discovery

travjenkins:
## Issues

Fixes https://github.com/estuary/ui/issues/756

## Changes

- 756
    - Refresh button will now update the draft before kicking off a new discovery
    - Allow the storing of collections to skip their update
    - Break up of `useCaptureDiscover` function into multiple hooks to reduce cyclomatic complexity
    - Prevent editor from making duplicate updates
    
- Misc
    - The timed progress bar resets when it hits the wait time
    - If the bar is changing variants we now debounce those changes to make the changes a bit less aggressive

## Tests

Manually tested create/edit of Capture

## Content

N/A

#### Enable/Disable is respected
Edited a capture and disabled the one binding
image

Clicked on refresh
image

Which updated the draft
image

Causing the rediscovery to keep the settings
image

#### Config changes are overwritten

Update the stream name
image

Draft spec is updated
image

The discovery comes back with original default name
image


src/components/capture/useCaptureConfigEncrypt.ts
travjenkins, line 1: With these new hooks there is no major logic changes compare to the chunks of code they were taken from. The only big change was making sure the return value is something that the "parent" function can use to know if things went well or not.

src/hooks/useServerUpdateRequiredMonitor.ts
travjenkins, line 1: This stuff will be implemented later in a stand alone pr

src/components/editor/MonacoEditor/index.tsx
```diff
@@ -131,6 +131,13 @@ function MonacoEditor({
 
     const updateValue = useCallback(
         (isUndo: boolean) => {
+            if (!isUndo && status !== EditorStatus.EDITING) {
+                logRocketConsole('editor:update:skipped', {
+                    status,
+                });
+                return;
+            }
```
travjenkins, line 139: This should help reduce duplicate updates flowing through the editor. Adding the logging to make sure it is always known when this happens in case there is an issue found.

src/components/shared/Entity/Actions/SchemaEvolution.tsx
```diff
@@ -98,6 +101,10 @@ function SchemaEvolution({ onFailure }: Props) {
                 // Now that the collections are updated we can clear out the list so
                 //  the error section is no longer displayed.
                 setIncompatibleCollections([]);
+
+                if (mutate_advancedEditor) {
+                    await mutate_advancedEditor();
+                }
```
travjenkins, line 107: Very similar to the `postGenerateMutate` function

src/hooks/useServerUpdateRequiredMonitor.ts
```diff
@@ -16,6 +16,8 @@ import {
 const useServerUpdateRequiredMonitor = (draftSpecs: DraftSpecQuery[]) => {
     const entityType = useEntityType();
 
+    // Need to fetch if we're in edit or not
```
kiahna-tucker, line 19: Is this comment leftover from an older piece of code, associated with the TODO comment below, an independent TODO comment, or something that should be addressed at this time? I am having difficulty distinguishing the intent.

kiahna-tucker (commented):
Finishing manual testing. Overall, the changes look fine but I did need clarity on one comment.

kiahna-tucker (approved):
Approved with moderate testing of the capture workflows. The progress indicator can shift from a determinate state to an indeterminate state then back to a determinate state (often completely filled), remaining on screen for a few seconds after the final transition before disappearing. I have noticed this most consistently when updating the endpoint configuration (after running discovery once, in the create workflow) and then refreshing the capture bindings. Is this a known UX quirk? If so, is it something we would like to address now or in a fast-follow?

Personally, given the value added by these changes, I think a fast-follow would be acceptable since this is a more mild UX issue.

PR #761, opened 2023-09-21
prod: preserve the selected binding while processing field selection updates

kiahna-tucker:
## Issues

## Changes

-   {issue_number}
    -   _list your changes_

## Tests

Manually tested

## Screenshots

### DIFF
--- a/src/components/collection/Selector/List/index.tsx
+++ b/src/components/collection/Selector/List/index.tsx
@@ -3,9 +3,9 @@ import {
     DataGrid,
     GridColDef,
     GridFilterModel,
-    gridPaginatedVisibleSortedGridRowIdsSelector,
     GridRowId,
     GridRowSelectionModel,
+    gridPaginatedVisibleSortedGridRowIdsSelector,
     useGridApiRef,
 } from '@mui/x-data-grid';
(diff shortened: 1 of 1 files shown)


src/components/collection/Selector/List/index.tsx
```diff
@@ -81,10 +83,18 @@ function CollectionSelectorList({
...
+
+    // Resource Config Store
     const collections = useResourceConfig_collections();
     const currentCollection = useResourceConfig_currentCollection();
 
-    const selectionEnabled = currentCollection && setCurrentCollection;
+    const selectionEnabled =
+        currentCollection &&
+        setCurrentCollection &&
+        formStatus !== FormStatus.UPDATING;
```
kiahna-tucker, line 96: We can restrict this condition further by applying the form status condition only when the entity type is a materialization. Presently, the binding selector will also be locked while the capture auto discovery settings are updated.

travjenkins: Reducing scope to just materializations kinda makes sense in terms of risk but then I think it makes the app kinda weird feeling. I think a consistently "worse" experience is better than a inconsistent experience.

travjenkins (approved):
lgtm - only request is that we make the debounce lower to decrease chance of this happening.

Issue #759, opened 2023-09-21
This Month and Last Month stats show zero


mdibaiee:
**Bug**
- "This Month" and "Last Month" stats show zero
- The issue seems to be caused by the request being sent having the wrong `ts`, see these sample requests:

This Month:
```
https://eyrcnmuzzyriypdajwdk.supabase.co/rest/v1/catalog_stats?select=catalog_name%2Cgrain%2Cts%2Cbytes_written_by_me%2Cdocs_written_by_me%2Cbytes_read_by_me%2Cdocs_read_by_me%2Cbytes_written_to_me%2Cdocs_written_to_me%2Cbytes_read_from_me%2Cdocs_read_from_me&catalog_name=in.%28Podder%2Fpodderdb%2FPodderDB%2Fdetaileddemographics_v2%2CPodder%2Fpodderdb%2FPodderDB%2Freviews_v2%2CPodder%2Fpodderdb%2FPodderDB%2Fpodcasts%2CPodder%2Fpodderdb%2FPodderDB%2Fplays%2CPodder%2Fpodderdb%2FPodderDB%2Fepisodes%2CPodder%2Fpodderdb%2FPodderDB%2Faudiences%29&order=catalog_name.asc.nullslast&ts=eq.2023-08-31+00%3A00%3A00%2B00&grain=eq.monthly
```

Last Month:
```
https://eyrcnmuzzyriypdajwdk.supabase.co/rest/v1/catalog_stats?select=catalog_name,grain,ts,bytes_written_by_me,docs_written_by_me,bytes_read_by_me,docs_read_by_me,bytes_written_to_me,docs_written_to_me,bytes_read_from_me,docs_read_from_me&catalog_name=in.(Podder/podderdb/PodderDB/detaileddemographics_v2,Podder/podderdb/PodderDB/reviews_v2,Podder/podderdb/PodderDB/podcasts,Podder/podderdb/PodderDB/plays,Podder/podderdb/PodderDB/episodes,Podder/podderdb/PodderDB/audiences)&order=catalog_name.asc.nullslast&ts=eq.2023-07-31 00:00:00+00&grain=eq.monthly
```

Notice that the `ts` in these requests is the last day of a month prior to what is expected. These requests were sent in September, from Ireland (UTC+1).

**Expected**
- Show correct monthly stats

**Screenshots**
[...]

PR #763, opened 2023-09-21
Switch to the `utc` object from Date-Fns

travjenkins:
## Issues

Fixes https://github.com/estuary/ui/issues/759

## Changes

- 759
    - Use the [utc](https://github.com/date-fns/utc) object for dates. Was originally concerned a bit as this is not mentioned in their docs when discussing timezone conversion. However, I can see how maybe they do not consider simply switching the local time to UTC as a "timezone conversion".
- Misc
    - Found a duplicated key while testing the app. Fixed it by adding a special key for when things are "nested"

## Tests

Manually tested
- Stats in table and on details page
- While testing logged the dates and compared to World Time Buddy

## Screenshots

image

Today
image

Yesterday
image

This Week
image

Last Week
image

This Month
image

Last Month
image

Detail Stats 6 hours
image

Billing Table
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
             "name": "estuary-ui",
             "version": "0.2.0-alpha",
             "dependencies": {
+                "@date-fns/utc": "^1.0.0",
                 "@dnd-kit/core": "^6.0.8",
                 "@dnd-kit/modifiers": "^6.0.1",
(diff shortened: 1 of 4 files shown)


kiahna-tucker (approved):
Approved with manual, local testing.

PR #769, opened 2023-09-22
Field selection UX improvements 2

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/771

## Changes

*   771
    * Set the `recommendFields` store property to `true` when initializing the field selection view and the `fields` stanza is not present in the binding configuration.
    
* Misc.
    * Check for the existence of inferred field types when rendering field selection table rows.

## Tests

* 771
    * Executed the recreation steps listed in the originating [issue](https://github.com/estuary/ui/issues/771) and confirmed that the defective scenario cannot be reproduced.

* Misc.
    * Populated the field selection table for a binding whose schema contains a **single** field that cannot exist and confirmed that the table renders without error.
    * Populated the field selection table for a binding whose schema contains a **multiple** fields that cannot exist and confirmed that the table renders without error.

## Screenshots

_N/A_


travjenkins (approved):
lgtm

Issue #784, opened 2023-10-11
Do not nag users that are paying outside of payment providers


travjenkins:
Some of our users are manually paying for the product outside of the payment provider. These customers can get flagged in the UI warning for missing payment methods. These users are confused and concerned when they see such warning.

We will be adding a new column to the tenants table that will store if they `pays_externally`

PR #785, opened 2023-10-11
Checking what payment provider tenant is using before warning payment methods

travjenkins:
Adding the types for that
Checking the new column to see if we need to fetch payment methods

## Issues

https://github.com/estuary/ui/issues/784

## Changes

784
- Added a new column to fetch from `tenants`
- Skip the call for payment methods for tenants that pay externally

Misc
- Stop fetching payment methods if there is no trial started
- Replaced checking all `tenants` with checking all `payment methods` so we do not scare users

## Tests

Manually tested
- Added the column manually and tested against it

## Screenshots
N/A

All external means no calls to billing to get payment methods
image
image

2 in stripe means two calls
image
image
image

2 in stripe but not in trial means no calls
image
image

### DIFF
--- a/src/api/billing.ts
+++ b/src/api/billing.ts
@@ -144,7 +144,10 @@ export const getPaymentMethodsForTenants = async (
     let count = 0;
 
     tenants.some((tenantDetail) => {
-        if (tenantDetail.pays_externally || !tenantDetail.trial_start) {
+        if (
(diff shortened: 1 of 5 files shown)


jshearer (approved):
LGTM after https://github.com/estuary/flow/pull/1232 is merged

PR #788, opened 2023-10-13
Quick fix for lingering store hydration

travjenkins:
## Issues

Found in slack channel

## Changes

Misc
- Added a flag of `active` to stores with hydration to store off if the store is being used. This is just a quick fix to prevent store's from being hydrated after the user leaves the pages that use them.

- Saw an issue with journalData spamming calls during network issues so making that better

## Tests

Manually tested
- Slowed the network and would enter capture create (with a connector) and quick leave. Then keep an eye on the store and see if the store ended up hydrated.

### DIFF
--- a/src/components/shared/Entity/ExistingEntityCards/Store/create.ts
+++ b/src/components/shared/Entity/ExistingEntityCards/Store/create.ts
@@ -58,7 +58,7 @@ const getInitialState = (
         },
 
         hydrateState: async (entityType, connectorId) => {
-            if (connectorId) {
+            if (get().active && connectorId) {
                 const { setCreateNewTask, setHydrationErrorsExist } = get();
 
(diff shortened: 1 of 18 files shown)


src/components/shared/Entity/ExistingEntityCards/Store/create.ts
```diff
@@ -58,7 +58,7 @@ const getInitialState = (
         },
 
         hydrateState: async (entityType, connectorId) => {
-            if (connectorId) {
+            if (get().active && connectorId) {
```
kiahna-tucker, line 61: Do you prefer this approach over destructuring?

kiahna-tucker (approved):
Approved without manual testing. Would you prefer to no longer destructure the state object in actions? Is this a pattern you are attempting to introduce here?

PR #790, opened 2023-10-16
Better error handling for publishing

travjenkins:
## Issues

From Slack conversation

## Changes

Slack
- If there is a network issue then let the polled try one more time
- Add a little bit more logging for LogRocket
- Handle when the status could be `draftEmpty`
- Query for the publication based on the returned `id`

## Tests

Manually tested

Published and paused the debugged...
 - alter the `job_status` on the `publications` table
 - disable the url `localhost:5431/rest/v1/publications` in chrome (before and after inserting publication)

## Screenshots

N/A


kiahna-tucker (approved):

Issue #789, opened 2023-10-13
Automatically re-discover after a binding is enabled


dyaffe:
**Bug**
When a user edits a capture selects "enable" on a binding that was previously disabled, they can not proceed without clicking refresh.  

This is extremely unintuitive to the point that I would classify it as a bug (though if you want to move it, feel free).

Instead, we should do a "refresh" after a user enables a binding on the "Next" step to prevent them from having to.

travjenkins:
Implementation note
We need to keep a list of what is enabled/disabled and when they go to click on next we see if there is anything going from disabled -> enabled and kick the refresh as an extra step.

PR #791, opened 2023-10-18
Rediscover capture when previously disabled collections are enabled

travjenkins:
## Issues

https://github.com/estuary/ui/issues/789

## Changes

- 789
    - Keep track of what collections were disable during the last load/discover
    - Keep track of what collections were enabled and will require rediscovery
    - Keep track of when rediscovery is needed
    - Have the capture's generate button pass if rediscovery is needed

## Tests

Manually tested

- Capture create
- Capture edit

## Screenshots

Enabling a previously disabled
image

Disabling it again is recorded
image

When clicking next we run a discover
image

And show the "discover" loading bar
image

### DIFF
--- a/src/components/capture/useDiscoverStartSubscription.ts
+++ b/src/components/capture/useDiscoverStartSubscription.ts
@@ -21,7 +21,6 @@ import {
 } from 'stores/EndpointConfig/hooks';
 import { useFormStateStore_setFormState } from 'stores/FormState/hooks';
 import { FormStatus } from 'stores/FormState/types';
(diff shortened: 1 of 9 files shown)


src/components/capture/useDiscoverStartSubscription.ts
travjenkins, line 1: I previously checked in some code for this when I thought I could add it quickly and just checked in some defaulting stuff. So this is why I am removing it now.

kiahna-tucker (approved):

PR #795, opened 2023-10-18
Prevent full page spinner from flashing on load

travjenkins:
## Issues

Long standing issue

## Changes

OnLoadSpinner
- Create a context that can show a full page spinner while the app starts
- Allow default value to be set so OAuth can be extra safe and not use this
- Added in component that will turn off/on the spinner
- Added the "turning off" component to full page wrapper and app layout as that seems to cover all the pages that can load while logging in.

Guards
- The guards no longer need to show their own spinner (except grants) so they now return null during load
- Removed some `suspense` uses as it was not doing anything right now

FullPageSpinner
- Have the full page spinner no longer show right away to give the network a second and not always flash the indicator to people
- Allow a delay to be passed in in case we need to skip the wait / make it longer

index.html
- Cleaned up some comments. Considered defaulting the background color but since we dark and light mode this would not always make the app look like it is loading faster.

## Tests

Manually tested

Both new and existing users...
- Logging in
- Granting tenant access (logged in and logging in)
- Granting tenant access with fake token
- Granting tenant access with used token
- Google Sheets OAuth

## Screenshots

Kind of hard to snag a screenshot and don't wanna include a video. Basically when the app is loading you should see one fullpage spinner during login


src/context/OnLoadSpinner/OnLoadSpinner.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useOnLoadSpinner } from './OnLoadSpinnerContext';
+
+interface Props extends BaseComponentProps {
+    display: boolean;
+}
+
+function OnLoadSpinner({ children, display }: Props) {
+    const { setLoading } = useOnLoadSpinner();
+
+    console.log('OnLoadSpinner', display);
```
kiahna-tucker, line 12: Flagging the presence of a logging statement in the event it should be removed.

kiahna-tucker (approved):

PR #802, opened 2023-10-25
Do not require `lastPubId` for details page

travjenkins:
## Issues

## Changes

-   {issue_number}
    -   _list your changes_

## Tests

Manually tested

## Screenshots

### DIFF
--- a/src/components/editor/Store/LiveSpecsHydrator.tsx
+++ b/src/components/editor/Store/LiveSpecsHydrator.tsx
@@ -9,13 +9,11 @@ import { BaseComponentProps } from 'types';
 import { hasLength } from 'utils/misc-utils';
 
 interface Props extends BaseComponentProps {
-    lastPubId: string;
     localZustandScope: boolean;
     collectionNames?: string[];
 }
 
 function LiveSpecsHydrator({
-    lastPubId,
     localZustandScope,
     collectionNames,
     children,
@@ -30,14 +28,11 @@ function LiveSpecsHydrator({
     const setId = useEditorStore_setId({ localScope: localZustandScope });
 
     useEffect(() => {
-        setId(lastPubId);
-    }, [lastPubId, setId]);
-
-    useEffect(() => {
         if (hasLength(publicationSpecs)) {
             setSpecs(publicationSpecs);
+            setId(publicationSpecs[0].last_pub_id);
         }
-    }, [publicationSpecs, setSpecs]);
+    }, [publicationSpecs, setId, setSpecs]);
 
     // TODO (details) make this error handling better
     // 1. Store this in the store
--- a/src/components/shared/Entity/Details/index.tsx
+++ b/src/components/shared/Entity/Details/index.tsx
@@ -24,9 +24,6 @@ function EntityDetails() {
 
     // Fetch params from URL
     const catalogName = useGlobalSearchParams(GlobalSearchParams.CATALOG_NAME);
-    const lastPubId = useGlobalSearchParams(GlobalSearchParams.LAST_PUB_ID);
-
-    console.log('lastPubId', lastPubId);
 
     return (
         <LocalZustandProvider createStore={localStore}>
@@ -34,7 +31,7 @@ function EntityDetails() {
                 collectionNames={[catalogName]}
                 localZustandScope={true}
             >
-                <ShardHydrator lastPubId={lastPubId} catalogName={catalogName}>
+                <ShardHydrator catalogName={catalogName}>
                     <Box>
                         <Stack spacing={2} sx={{ m: 1 }}>
                             <Stack direction="row" spacing={1}>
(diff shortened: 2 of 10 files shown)


src/components/editor/Store/LiveSpecsHydrator.tsx
```diff
@@ -29,15 +27,12 @@ function LiveSpecsHydrator({
...
     const setId = useEditorStore_setId({ localScope: localZustandScope });
 
-    useEffect(() => {
-        setId(lastPubId);
-    }, [lastPubId, setId]);
-
     useEffect(() => {
         if (hasLength(publicationSpecs)) {
             setSpecs(publicationSpecs);
+            setId(publicationSpecs[0].last_pub_id);
```
travjenkins, line 33: If we just set the `id` in the hydrator then we do not have to pass it around as much

src/components/shared/Entity/Details/index.tsx
travjenkins, line 1: No longer need to pass around the `lastPubId`

src/hooks/useLiveSpecs.ts
travjenkins, line 1: Want to query for the `last_pub_id` so that we can use the query response to populate the `id` in the store

src/api/liveSpecsExt.ts
```diff
@@ -351,17 +347,14 @@ export interface LiveSpecsExtQuery_ByLiveSpecId {
     spec_type: Entity;
     last_pub_id: string;
     spec: any;
+    connector_id: string;
```
travjenkins, line 350: Need this so we can pass it along during hydration (instead of passing it in the URL)

src/components/shared/Entity/hooks/useEntityWorkflowHelpers.ts
```diff
@@ -142,20 +149,29 @@ function useEntityWorkflowHelpers() {
...
+            pubId,
+            entityType
         );
-    }, [exit, navigate, pubId]);
+
+        const liveSpecId = liveSpecResponse.data?.[0]?.live_spec_id;
+
+        if (!liveSpecId) {
+            // TODO need to show an error here
+            console.error('uh oh');
```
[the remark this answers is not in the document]
travjenkins: removed

src/components/editor/Store/LiveSpecsHydrator.tsx
travjenkins: Yeah - the concept of `id` has caused a few issues while working on this but yeah... didn't want to boil the ocean on this one.

src/components/materialization/Edit.tsx
```diff
@@ -23,6 +23,7 @@ import { EndpointConfigHydrator } from 'stores/EndpointConfig/Hydrator';
 import ResourceConfigHydrator from 'stores/ResourceConfig/Hydrator';
 
 function MaterializationEdit() {
+    console.log('MaterializationEdit');
```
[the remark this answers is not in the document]
travjenkins: removed

kiahna-tucker (approved):

src/api/publicationSpecsExt.ts
```diff
@@ -27,3 +27,16 @@ export const getPublicationHistoryByCatalogName = (catalogName: string) => {
...
+    pubId: string | null, // Do not actually pass null... just making typing easiser
+    entityType: Entity
+) => {
+    return supabaseClient
+        .from<PublicationSpecsExt_PublicationHistory>(
+            TABLES.PUBLICATION_SPECS_EXT
+        )
+        .select(`live_spec_id`)
+        .eq('pub_id', pubId)
+        .eq('spec_type', entityType);
```
psFried, line 41: The combination of `pub_id` and `spec_type` isn't enough to uniquely identify a publication spec. Consider the `pub_id` similar to a `draft_id`, in that it can identify any number or combination of specs. So you could have multiple captures in a publication, for example. This seems like an issue because the callsite is just taking the first returned element and assuming that's the right one.

This may not be a problem in practice, since I'm assuming that the `pub_id` corresponds to a UI-created publication that wouldn't actually have multiple _task_ specs (though it would have multiple collection specs). But I'd recommend changing this function to accept a `catalog_name` parameter instead of `spec_type`, if possible.

src/api/publicationSpecsExt.ts
```diff
@@ -27,3 +27,16 @@ export const getPublicationHistoryByCatalogName = (catalogName: string) => {
             ascending: false,
         });
 };
+
+export const getLiveSpecIdByPublication = (
+    pubId: string | null, // Do not actually pass null... just making typing easiser
```
psFried, line 32: 🙃 lol this brings up many complicated emotions

psFried (commented):
One comment on the `publication_specs_ext` query, but otherwise looks good to me.

src/api/publicationSpecsExt.ts
travjenkins: Cool... that was a worry I just had this morning. I checked the DB and it looks like it would be fine but still could break.

That's a good call on the `catalog_name` I'll switch to that. We have everything available because we're calling from the create/edit page.

Issue #808, opened 2023-10-30
Better UX for "See Fields" buttons


travjenkins:
You should not have to click the next button before clicking "See Fields" and the "See Fields" button should just "click next" for the user.

PR #810, opened 2023-10-31
Make see fields generate draft

travjenkins:
## Issues

Fixes https://github.com/estuary/ui/issues/808
Fixes https://github.com/estuary/ui/issues/812

## Changes

808
- Create hooks to share functionality
- Create a context for the draft updater so we don't have to pass around
- Make the `See Fields` button a unique button

812
- Add a `scrollIntoView` hook that is used in the errors for entity create/edit

## Tests

Manually tested

- [x] Capture create
- [x] capture edit
- [x] materialization create
- [x] ... clicking `See Fields` before `Next`
- [x] ... clicking `See Fields` after `Next`
- [x] materialization edit
- [x] ... clicking `See Fields` before `Next`
- [x] ... clicking `See Fields` after `Next`

## Screenshots

There is no real screenshots here as most of the functionality should be the same. The biggest visual change for a user is that the page will scroll up to the error if there is one for both technical issues (network is down) and for when there are validation issues.


src/components/capture/Create/index.tsx
```diff
@@ -86,46 +87,48 @@ function CaptureCreate() {
...
-                        }
-                        RediscoverButton={
-                            <RediscoverButton
-                                entityType={entityType}
-                                disabled={!hasConnectors}
-                                postGenerateMutate={updateDraftSpecs}
-                            />
-                        }
-                    />
+                    <MutateDraftSpecProvider value={updateDraftSpecs}>
```
travjenkins, line 90: This kind of change is in several files. The main change is a new wrapper for the context and no longer passing this in props.

src/components/capture/GenerateButton.tsx
```diff
@@ -26,14 +25,11 @@ interface Props {
...
     entityType,
     disabled,
-    postGenerateMutate,
     createWorkflowMetadata,
 }: Props) {
     const rediscoveryRequired = useResourceConfig_rediscoveryRequired();
-
     const { generateCatalog, isSaving, formActive } = useDiscoverCapture(
         entityType,
-        postGenerateMutate,
```
travjenkins, line 36: Now that this function is in a context the hook can fetch it manually and no longer needs passed.

src/components/capture/GenerateButton.tsx
travjenkins, line 1: This kind of change is in a lot of spaces... since there is context the hook can fetch it manually and no longer needs passed.

src/components/materialization/GenerateButton.tsx
travjenkins, line 1: Moved most of the functionality into a hook.

src/components/shared/Entity/Actions/Save.tsx
travjenkins, line 1: Just like generate - moved into a hook.

Issue #505, opened 2023-03-02
Expand handling of possibly ephemeral errors to help users recover from issues


dyaffe:
**Problem**
We have a relatively common error mode currently that says something to the effect of "this isn't something you did wrong, please contact support".  When that happens, the error is usually ephemeral and they could just try it again to get it to work, but the error message makes them think that the product is broken and they should abandon it.  [Example session](https://app.logrocket.com/tue20y/estuarytest/s/5-7a42f775-b059-4729-a338-0dfd76d8c82e/0?t=1677719844484.2603)

**Proposal**
Ideally, we should auto retry when that happens.

Either way, we should change the error message to say something like "We've encountered an issue (with whatever external system it is here), please try again."

travjenkins:
This is a good example of where we should try to handle errors better. Instead of just giving up we should get some logic to retry the calls a couple times. I think while trying we might want to add some kind of messaging to the user so they aren't sitting there wondering what is happening.

https://app.logrocket.com/tue20y/estuarytest/s/5-7a9479fd-6528-47c9-acd9-848862ec6d77/0?network_id=1-fetch-0&network_root_tab_id=d50aa728-343f-40b9-8bc2-24ee86acd07a&network_tab_id=d50aa728-343f-40b9-8bc2-24ee86acd07a&t=1681815223447.994

travjenkins:
All the `swr` calls will retry on their own. However, this is not communicated to a user at all. We'll need to handle showing some kind of messaging when we are retrying. This should probably only be shown when we have retried at least once... that way we don't show issues during a random network blip.

Also, we'll need to make sure that the `swr` retry messages are the same as any other promise that is getting retried. More than likely we'll want to use the normal `notistack` snackbar as it is available everywhere in the app.

In addition to showing when retrying we need to make sure the user knows when we have **stopped** retrying.

PR #814, opened 2023-11-02
Retry async calls if they fail 

travjenkins:
## Issues

https://github.com/estuary/ui/issues/505

#### 505
- Added a promise retry for when calls using the supabase API fail
- Log when a call to supabase fails

#### Misc
- We want to log when calls fail so needed to break up the LogRocket files a bit

## Tests

Manually tested

- Used network tools to mess with the network and force failures

## Screenshots

Calls failed and then succeed (success logging is no longer there)
image


kiahna-tucker (approved):

src/lang/en-US.ts
```diff
@@ -141,6 +143,7 @@ const CTAs: ResolvedIntlConfig['messages'] = {
     'cta.showAll': `Show All`,
     'cta.reload': `Reload`,
     'cta.evolve': `Apply`,
+    'cta.support': `contact support`,
```
kiahna-tucker, line 146: This CTA appears in six locations outside of `'error.instructions'`. Were you thinking of replacing them separately?

src/api/tokens.ts
```diff
@@ -1,16 +1,22 @@
...
-        })
-        .throwOnError()
-        .single();
+    return supabaseRetry<
+        PostgrestSingleResponse<{ id: string; secret: string }>
+    >(
+        () =>
+            supabaseClient
+                .rpc(RPCS.CREATE_REFRESH_TOKEN, {
+                    multi_use,
```
kiahna-tucker, line 16: Where do you stand on using snake case for the input parameters of API-related functions?

src/components/shared/Error/Message.tsx
```diff
@@ -13,21 +14,24 @@ function Message({ error }: Props) {
...
-        typeof error === 'object' && error.hasOwnProperty('code');
+    const failedAfterRetry = retryAfterFailure(error?.message ?? error);
+
+    // Check if we have retried making the call or if we think the message object
+    //      is from Supabase
+    //  Initially was going to write a RegEx that detects if the mesaage
+    //      needs translated. However, Our shortest keys (cta.foo)
+    //      can also look a lot like how Supabase returns
+    //      errors when fetching wrong keys from a table (tableName.col).
+    const displayOnlyError =
```
kiahna-tucker, line 25: This is the smallest of nits, but I believe `displayErrorOnly` is a better phrasing of this variable name, aligning it with flags of a similar demarcation (e.g., `readOnly`, `nameOnly`, etc.).

src/services/shared.ts
```diff
@@ -0,0 +1,19 @@
+import LogRocket from 'logrocket';
+import { CustomEvents } from './types';
+
+export const logRocketEvent = (
```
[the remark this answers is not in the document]
travjenkins: It causes a circular dep issue

src/lang/en-US.ts
travjenkins: Yeah - gonna switch 'em over later on.

src/services/supabase.ts
```diff
@@ -220,35 +214,66 @@ export interface CallSupabaseResponse<T> {
...
-    };
+export const handleFailure = (error: any) => ({
+    data: null,
+    error,
+});
+
+// Retry calls
+const RETRY_ATTEMPTS = 2;
+
+export const supabaseRetry = <T>(makeCall: Function, action: string) => {
```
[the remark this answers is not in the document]
travjenkins: No - I'd rather make it clear that the calling function is not passing an action.

src/api/tokens.ts
travjenkins: I don't really prefer snake case unless it is for something that is "from the back end". So here it kinda makes sense so that we don't have to translate the data.

src/components/shared/Error/Message.tsx
travjenkins: fixed

PR #816, opened 2023-11-06
Create the foundation for prefix-based notifications

kiahna-tucker:
## Issues

## Changes

-   {issue_number}
    -   _list your changes_

## Tests

Manually tested

## Screenshots


src/components/admin/Settings/PrefixAlerts/Dialog/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+        ) {
+            setUpdatedEmails({
+                ...updatedEmails,
+                [prefix]: existingEmails[prefix],
+            });
+        }
+    }, [open, prefix, existingEmails, setUpdatedEmails, updatedEmails]);
+
+    const updatePrefix = (value: string, errors: string | null) => {
+        // if (serverError) {
```
kiahna-tucker, line 71: Will remove this commented code fragment.

src/components/shared/Entity/Details/Overview/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const [subscriptionError, setSubscriptionError] =
+        useState<ErrorDetails>(null);
+
+    const [updateSettingsError, setUpdateSettingsError] =
+        useState<ErrorDetails>(null);
+
+    useEffect(() => {
+        getNotificationSubscription().then(
+            (response) => {
+                if (!response.data) {
```
kiahna-tucker, line 35: This `if` condition and the `else if` condition below can be combined. The two scenarios required unique handling in a previous iteration, but that is no longer the case.

src/components/admin/Settings/PrefixAlerts/Dialog/SaveButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+        Object.entries(updatedEmails).forEach(([key, value]) => {
+            if (Object.hasOwn(existingEmails, key)) {
+                existingEmails[key]
+                    .filter((email) => !value.includes(email))
+                    .map((email): [string, string] => [key, email])
+                    .forEach((subscriptionMetadata) => {
+                        subscriptionsToCancel.push(subscriptionMetadata);
+                    });
+            }
+        });
```
travjenkins, line 75: `todo`
Looking at this... I do not think we need to have two loops here. Can we try to combine these two loops to reduce how often we have to loop. Do not need to get handled with this PR so either it can get fixed or mark it with a `todo`

src/components/admin/Settings/PrefixAlerts/EditButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+            <Button variant="text" onClick={openGenerateAlertDialog}>
+                <FormattedMessage id="cta.edit" />
+            </Button>
+
+            {subscriptions === null ? null : (
+                <AlertSubscriptionDialog
+                    headerId="admin.alerts.dialog.update.header"
+                    open={open}
+                    setOpen={setOpen}
+                    subscriptions={subscriptions}
```
travjenkins, line 25: Since we have a hook I do not think we need to be passing this stuff as the Dialog itself could access this through the hook.

src/components/admin/Settings/PrefixAlerts/EmailSelector.tsx
```diff
@@ -0,0 +1,227 @@
...
+    );
+
+    const { data: userInfo } = useUserInformationByPrefix(
+        [prefix, ...adminPrefixes],
+        minCapability
+    );
+
+    const emails = useMemo(
+        () =>
+            Object.hasOwn(emailsByPrefix, prefix) ? emailsByPrefix[prefix] : [],
```
travjenkins, line 48: This could be made a bit simpler with `emailsByPrefix[prefix] ?? []`

src/components/admin/Settings/PrefixAlerts/Dialog/SaveButton.tsx
kiahna-tucker: Ha, I cannot believe I missed this on my first pass at refactoring. The logic used to be unique here. There is no need for the to loops over the `updatedEmails` object. Will update.

src/components/admin/Settings/PrefixAlerts/EmailSelector.tsx
kiahna-tucker: That should not be possible since the compiler _should_ complain that the conditional is unnecessary since `emailByPrefix[prefix]` cannot be evaluated to `null` or `undefined` at compile time.

src/components/inputs/PrefixedName/useValidatePrefix.ts
```diff
@@ -0,0 +1,157 @@
...
+                    })
+                );
+            });
+        };
+
+        // If there are any errors then populate the list
+        if (prefixErrors) {
+            generateErrorList('prefix', prefixErrors);
+        }
+        if (!prefixOnly && nameErrors) {
```
travjenkins, line 75: Thinking out loud... and I am not sure if I deeply believe this

I wonder if having the settings be more exact like `ignoreName` or something might be a good approach. My main thinking is that is makes it more clear what you're doing and allows for a third option to be added and still work.

However... if we do make this kind of thing `ignoreName` then when adding a third if we wanted to keep this `prefixOnly` we would need to pass `ignoreThird`.

src/components/shared/Entity/Details/Overview/NotificationSettings/DataProcessingSetting.tsx
```diff
@@ -0,0 +1,202 @@
...
+                            setUpdateSettingsError(
+                                response.error ?? defaultUpdateSettingsError
+                            );
+                        }
+                    },
+                    (error) => {
+                        setUpdateSettingsError(error);
+                    }
+                );
+            }
```
travjenkins, line 118: Please share these success/error handlers.

src/components/tables/PrefixAlerts/Rows.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+            <ChipListCell
+                values={data.userSubscriptions.map(({ email }) => email)}
+                stripPath={false}
+                maxChips={3}
+            />
+
+            <TableCell>
+                <AlertEditButton prefix={prefix} />
+            </TableCell>
```
travjenkins, line 33: nit pick / thinking

The component `AlertEditButton` is only ever consumer here... this would probably make sense to make a `cell` for it to follow the pattern.

src/hooks/useUserInformationByPrefix.ts
```diff
@@ -0,0 +1,25 @@
+import { getUserInformationByPrefix } from 'api/combinedGrantsExt';
+import { useSelectNew } from 'hooks/supabase-swr/hooks/useSelect';
+import { Capability, Grant_UserExt } from 'types';
+
+// TODO (optimization): The combined_grants_ext view limits the function of this hook. The original intent of the hook
+//   is to return information (email, full name, avatar URL, and user ID) for all administrators of a tenant, including users
+//   who have administrative access via a prefix privilege (e.g., user A is an admin of tenantB/ because tenantA/ has admin access
+//   to that prefix).
```
travjenkins, line 8: This does not feel like a `todo` to me and just really commenting on limitations which can be handled with just a comment.

src/utils/notification-utils.ts
```diff
@@ -8,3 +10,75 @@ export const snackbarSettings: OptionsObject = {
     preventDuplicate: true,
     variant: 'info',
 };
+
```
travjenkins, line 13: no change needed

But I feel like this is not the right spot for this code. This util is really for notifications displaying in the UI specifically... and the new stuff feels more aligned to the subscriptions and backend implementation

Issue #844, opened 2023-12-05
Not all inferred schema errors show warning


travjenkins:
**Bug**
Some users are seeing inferred schema errors as normal errors and not warnings

**Expected**
They should see as a warning

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Additional context**
N/A

PR #845, opened 2023-12-05
Updating logic so if a schema issue is on the root we catch it

travjenkins:
## Issues

https://github.com/estuary/ui/issues/844

### 844
- Updated logic to catch when the schema has an issue at root. Since this will catch both I just updated the string and did not check for multiple strings

## Tests

Manually tested
This is a prod issue so not testing locally. I checked the string would catch the value by doing a "find" in LogRocket


jgraettinger (approved):
LGTM

Issue #737, opened 2023-09-11
Make Closing Docs Pages "Sticky"


dyaffe:
**Problem**
On small screens, the docs pages compress your view enough that they make the UI less usable.  It would be nice if when a user closed them, they stayed closed for a session.  

Ex.  go to postgres to create a capture and close the docs screen then go to snowflake to create a materialization, it shouldn't re-open.

travjenkins:
This was weirdly harder than I thought. I tried picking this up quickly and how the current solution we have gets a little janky because it is architected expecting to always force the docs open.

This is still totally doable but not like a "30 minute" change as I first assumed.

PR #861, opened 2023-12-11
Preserve side panel docs display state

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/737

## Changes

-   737
    * Remove the `show` state property and `setShow` action from the `SidePanelDocsStore`.
    * Store a `boolean` flag in local storage that indicates whether the side panel docs should be displayed.
    * Create a context provider to access and set the aforementioned side panel docs display flag in local storage.

## Tests

Manually tested

-   737
    * Confirm that the side panel docs open by default when a create/edit workflow is entered (i.e., when they have not been explicitly closed).

    * Close the side panel docs in a create/edit workflow for one entity type, reopen the side panel docs, and validate that the correct documentation is displayed.

    * Close the side panel docs in a create/edit workflow for one entity type, switch connectors at least twice, and validate that the side panel docs remain closed.

    * Close the side panel docs in a create/edit workflow for one entity type, switch connectors at least once, reopen the side panel docs, and validate that the correct documentation is displayed.

    * Close the side panel docs in a create/edit workflow for one entity type, complete the workflow, enter the create/edit workflow for another entity type, and validate that the side panel docs remain closed.

    * Close the side panel docs in a create/edit workflow for one entity type, reopen the side panel docs, complete the workflow, enter the create/edit workflow for another entity type, and validate that the side panel docs remain open and the correct documentation is displayed.

## Screenshots

N/A


src/components/sidePanelDocs/Iframe.tsx
```diff
@@ -67,12 +72,11 @@ function SidePanelIframe() {
...
             }, 100);
         };
-        iframeCurrent?.addEventListener('load', hideNavBar);
 
-        return () => {
-            iframeCurrent?.removeEventListener('load', hideNavBar);
-        };
-    }, [iframeCurrent, setAnimateOpening]);
+        if (iframeCurrent && show && hasLength(docsURL)) {
+            hideNavBar();
```
travjenkins, line 77: If we only call this within this block I think we should just move the declaration inside as well.

src/stores/SidePanelDocs/Store.ts
```diff
@@ -34,23 +33,13 @@ const getInitialState = (
...
-    },
-
     setUrl: (val) => {
         set(
             produce((state: SidePanelDocsState) => {
                 const urlIsFromEstuary = val.includes(iframeStringInclude);
 
                 state.url = val;
-                state.show = true;
+                // state.show = true;
```
travjenkins, line 42: We should be good to just remove this totally.

travjenkins (approved):
Some small changes needed but overall LGTM.

src/stores/SidePanelDocs/Store.ts
kiahna-tucker: This is obviously a commented line of code that was simply missed in a cleanup effort.

src/components/sidePanelDocs/Iframe.tsx
kiahna-tucker: Definitely. I wanted to make sure the code was progressing in the right direction before removing the function definition outright.

PR #871, opened 2023-12-12
Better handling of the user object for LR and more logging

travjenkins:
## Issues

https://github.com/estuary/ui/issues/872

### 872
- Stop used `Auth.useUser` and switch to fetching user from session. When someone is logged out their User object hangs around but their session is cleared right away.
- Stop using `user` at all inside of some directives as the RLS should handle that

### Misc
- Small update on Supabase JS library
- Marking if we have identifies a user so we do not call it multiple times

## Tests

Manually tested

- Logging in and out a lot
- Messing with the auth heads and user object in local storage
- Moving between tabs

## Screenshots

N/A


kiahna-tucker (approved):

kiahna-tucker (changes requested):
Tenant invitations are broken.

kiahna-tucker (approved):

Issue #775, opened 2023-09-28
Ordering of Bindings in Field Selection Doesn't Match Schema Ordering


dyaffe:
**Bug**
When we created the tabular schema viewing functionality, we specifically chose to place `_meta` fields last.  This isn't the case in field selection and it would be good to ensure that field selection mirrors the same design choices we've made elsewhere.

The rationale behind this is the same as for schema viewing -- users don't generally care about _meta fields and have to scroll past them.

Please feel free to re-label this as appropriate if you deem it not to be a defect

kiahna-tucker:
The rows of the field selection table are initially sorted by the constraint type severity as identified and requested in the PRD. Is this request to have the table rows sorted by field name **only** by default, with `_meta` fields appearing last in an ascending sort?

travjenkins:
@dyaffe - what are you expecting here? Would you **always** want the `_meta` fields to be at the bottom? 

Ex:
Sorting by name
```
foo/1
foo/2
foo/3
_meta/1
_meta/2
_meta/3
```

Sorting by constraint
```
name    | constraint

foo/2   | recommended
foo/1   | recommended
foo/3   | required
_meta/2 | recommended
_meta/1 | recommended
_meta/3 | required
```

PR #873, opened 2023-12-13
Adjust the constraint type sort behavior on the field selection table

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/775

## Changes

-   775
    * Update the constraint type sort behavior on the field selection table so that fields prefixed by `_meta` always appear last. See the issue for an example of the desired sort behavior.
    * Correct a comment in the `basicSort_string` function definition.

## Tests

-   775
    * Validate the constraint type ascending sort behavior on the field selection table.
    * Validate the constraint type descending sort behavior on the field selection table.
    * Validate the field ascending sort behavior on the field selection table.
    * Validate the field descending sort behavior on the field selection table.

## Screenshots

**Field Selection Table | Constraint Type Sort | Ascending**

pr_screenshot-873-field_selection_sort-constraint_type_asc

<br />
<br />

**Field Selection Table | Constraint Type Sort | Descending**

pr_screenshot-873-field_selection_sort-constraint_type_desc

<br />
<br />

**Field Selection Table | Field Sort | Ascending**

pr_screenshot-873-field_selection_sort-field_asc

<br />
<br />

**Field Selection Table | Field Sort | Descending**

pr_screenshot-873-field_selection_sort-field_desc

### DIFF
--- a/src/components/tables/FieldSelection/Rows.tsx
+++ b/src/components/tables/FieldSelection/Rows.tsx
@@ -24,7 +24,7 @@ interface RowsProps {
     columnToSort: string;
 }
 
-const evaluateConstraintTypes = (
+const compareConstraintTypes = (
     a: CompositeProjection,
     b: CompositeProjection,
     ascendingSort: boolean
@@ -71,7 +71,7 @@ const constraintTypeSort = (
 
     const ascendingSort = sortDirection === 'asc';
 
-    return evaluateConstraintTypes(a, b, ascendingSort);
+    return compareConstraintTypes(a, b, ascendingSort);
 };
 
 function Row({ row }: RowProps) {
--- a/src/components/tables/FieldSelection/Rows.tsx
+++ b/src/components/tables/FieldSelection/Rows.tsx
@@ -21,6 +21,78 @@ interface RowsProps {
     columnToSort: string;
 }
 
+const constraintTypeSort_ascending = (
+    a: CompositeProjection,
+    b: CompositeProjection
+): number => {
+    // If a and b have constraint types but they are not equal, compare their severity.
+    // The projection with the greater severity should appear first.
+    if (
+        a.constraint &&
+        b.constraint &&
+        a.constraint.type !== b.constraint.type
+    ) {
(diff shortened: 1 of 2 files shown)


src/components/tables/FieldSelection/Rows.tsx
```diff
@@ -21,6 +21,78 @@ interface RowsProps {
...
+
+    // If a is not alpha and b is then return >0 to put b first
+    if (!aIsAlphabetical && bIsAlphabetical) {
+        return 1;
+    }
+
+    // If a is alpha and b isn't then return <0 to put a first
+    if (aIsAlphabetical && !bIsAlphabetical) {
+        return -1;
+    }
```
travjenkins, line 79: These duplicated lines should be shared. I think we could make it quick and just have both this and `basicSort_string` calling a 3rd function that handles the shared stuff.

src/components/tables/FieldSelection/Rows.tsx
```diff
@@ -21,6 +21,78 @@ interface RowsProps {
...
+    a: CompositeProjection,
+    b: CompositeProjection
+): number => {
+    // If a and b have constraint types but they are not equal, compare their severity.
+    // The projection with the greater severity should appear first.
+    if (
+        a.constraint &&
+        b.constraint &&
+        a.constraint.type !== b.constraint.type
+    ) {
```
travjenkins, line 34: This should be shared to make sure it stays in sync.

travjenkins (approved):
Some small code sharing needs handled.
lgtm

src/components/tables/FieldSelection/Rows.tsx
kiahna-tucker: The thought crossed my mind but I could not arrive at an appropriate default return value for this new, shared function. The value should not be a number but `null` and `undefined` seem equally unsuitable. What would you deem acceptable here?

travjenkins: I think we can go with `null`. That way it is easy to test if the function is returning an actual sort value.

Issue #900, opened 2023-12-22
Ensure failed file loading is handled properly


travjenkins:
**Bug**
Sometimes when users are attempting to hit a capture that loads in their files there can be an error that shows up.

**Expected**
We handle this error correctly

**Screenshots**
from slack:
![image](https://github.com/estuary/ui/assets/270078/aa0caed5-292e-474d-8250-0e74cb6ecbf4)

**Desktop (please complete the following information):**
N/A

**Additional context**
We have handled this in the past - but I assume Vite changes how loading fails somehow.

PR #901, opened 2023-12-22
Catch loading errors

travjenkins:
## Issues

https://github.com/estuary/ui/issues/900

### 900

- New function that will handle errors from lazy loading and return the dialog whenever it is needed

### Manually tested

- Ran a build, ran preview, hit the page, made simple change, stopped preview, ran a build, ran a preview, went back to browser and clicked around

### Automated tests

- n/a

## Screenshots

same as before

### DIFF
--- a/src/services/react.ts
+++ b/src/services/react.ts
@@ -1,0 +1,17 @@
+import MustReloadDialog from 'components/shared/MustReloadDialog';
+import { lazy } from 'react';
+import { logRocketConsole, logRocketEvent } from './shared';
+import { CustomEvents } from './types';
+
+const handledLazy = (factory: () => Promise<{ default: any }>) => {
+    return lazy(() =>
+        factory().catch((error) => {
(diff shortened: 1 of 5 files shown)


src/services/react.ts
```diff
@@ -0,0 +1,17 @@
+import MustReloadDialog from 'components/shared/MustReloadDialog';
+import { lazy } from 'react';
+import { logRocketConsole, logRocketEvent } from './shared';
+import { CustomEvents } from './types';
+
+const handledLazy = (factory: () => Promise<{ default: any }>) => {
```
kiahna-tucker, line 6: I am not the _biggest_ fan of this function name, but it gets the general idea across and I cannot think of a better alternative at the moment.

kiahna-tucker (approved):

src/services/react.ts
travjenkins: yeah... I had no clue what to use without getting way too verbose.

Issue #864, opened 2023-12-12
Data preview section of the collection details page does not have a loading state


kiahna-tucker:
## Bug

The _Data Preview_ section of a collection _Details_ page is nearly empty while its content is fetched.

### Recreation Steps

1. Navigate to the _Details_ page of a collection. For higher-speed networks, throttling may be needed to observe the defect.

1. Scroll down the page until the _Data Preview_ section is in the view.

1. Observe the absence of a loading state for the data preview component.

## Expected

The _Data Preview_ section of a collection _Details_ page should display a loading state while its content is fetched that reflects the shape of the content.

## Screenshots

<img width="564" alt="ui_issue-000-data_preview_missing_loading_state-loading" src="https://github.com/estuary/ui/assets/77648584/19100afa-ca94-450b-a04e-133c5b3fccc1">

## Desktop
 - **OS:** Windows
 - **Browser:** Chrome (Version 119.0.6045.200)

## Additional Context

N/A

PR #911, opened 2024-01-02
Display a loading state for data preview

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/864

### 864

* Display a loading state in the _Data Preview_ section of the _Details_ page when journal data is being fetched.

* Complete the outer border of the data preview list view component.

### Manually tested

* Validate that a loading state appears and hides as expected in the _Data Preview_ section of the _Details_ page when journal data is being fetched.

### Automated tests

N/A

## Screenshots

**Data preview | Loading state**

pr_screenshot-911-data_preview_loading_state-loading

<br />
<br />

**Data preview | Initialized**

pr_screenshot-911-data_preview_loading_state-completed_border


travjenkins (commented):
This is a good step - but there is a bi of wonky-ness still.

If there is no data to show then it seems like there is no loading shown, then a flash of something, and then nothing again.

Would we be able to get the loading state showing a bit sooner?

kiahna-tucker:
I noticed the oddity called out in the comment above while testing as well. Commit `4285cc4` should resolve this issue.

src/components/collection/DataPreview/index.tsx
```diff
@@ -47,7 +48,7 @@ export function DataPreview({ collectionName }: Props) {
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
     const journal = useMemo(() => journalsData?.journals?.[0], [journalsData]);
     const journalData = useJournalData(journal?.name, 20, collectionName);
-    const isLoading = journalsLoading || journalData.loading;
+    const isLoading = journalsLoading || !journalsData || journalData.loading;
```
kiahna-tucker, line 51: There is a brief delay between when the data preview card is rendered and the two journal-related hooks are called. which resulted in `isLoading` being a false negative.

This solution produces the desired results the majority of the time but is a hacky fix in and of itself. If the journal client is `null` or the collection name undefined, `journalsData` will be `null` and this component will be stuck in a loading state. Ideally, this condition would specifically check whether `journalsData` is undefined, however it is possible for the variable to quickly switch from undefined to `null` then back to undefined while loading.

travjenkins (approved):
Lgtm

Issue #866, opened 2023-12-12
Add integrity to LogRocket loading


travjenkins:
We need to include the integrity tag when loading in LogRocket.

travjenkins:
Dug into this and do not believe we can do this today. We have added some scripts to keep an eye on how often the hash would change... but it can change at any time.

I have requested LogRocket add a feature to include the integrity hash when they are importing their JS dynamically.

Also, I have looked into several other 3rd party providers and many of them have flat out refused to add support for SRI.

I think for LogRocket our best approach will be to setup a proxy server.

PR #914, opened 2024-01-03
Add in script to check if LogRocket integrity has changed

travjenkins:
## Issues

https://github.com/estuary/ui/issues/866

### 866

- Added a `logRocketIntegrity` script to check that we are using the latest
- ~~Changed to load in LogRocket manually through injecting a script tag~~ reverted
- Added `.env` properties for LogRocket integrity
- ~~Added section to the read me~~reverted

### Manually tested

- Manually tested for using old sha, latest sha, and failed to load
- Hit the page with Log Rocket enabled and disabled

### Automated tests

- n/a

## Screenshots

If the latest hash is being used
image

If an update is required
image

If the main `catch` is hit
image

### DIFF
--- a/.env
+++ b/.env
@@ -26,11 +26,16 @@ VITE_OSANO_PATH=https://cmp.osano.com/16CPXbTOi1sXx4D3/aaa8ccd6-4dbe-43d2-b79b-9
 VITE_OSANO_HIDE_WIDGET_CLASS=estuary-logged-in
 
 # LogRocket settings
+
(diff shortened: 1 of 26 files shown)


kiahna-tucker:
Ran a small set of manual tests and the command functions as expected thus far. I needed to update the version of the `node` CLI to v20.10.0 since the `--env-file` option was introduced recently.

Let me know if any changes arise past this point that you would like me to review.

kiahna-tucker (approved):

Issue #915, opened 2024-01-05
Page crashing due to translation issue


travjenkins:
**Bug**
Sometimes when a journal fails to load the error is getting displayed and crashing the page due to the message not being there/translatable.

**Expected**
The error should display as normal

**Screenshots**

<img width="1064" alt="forced-journal-error" src="https://github.com/estuary/ui/assets/77648584/13e97049-49e4-4434-8ce9-1d602d29826f">

travjenkins:
This happened again

travjenkins:
Added another round of fixes with a shared component

PR #916, opened 2024-01-05
Better error message and error handling for data plane gateway

travjenkins:
## Issues

https://github.com/estuary/ui/issues/915

### 915
- Got journalData hook using proper error handling
- Mare the error message component a bit more safe by including a fallback message
- More logging to find possible issues

### Misc
- Created shared utils for data plane gateway to share
- Added `FormatJS` linting rules

### Manually tested

- Messed with the URL blocking via Chrome debugger to view shards/journals

### Automated tests

- N/A

## Screenshots

- N/A


.eslintrc.js
```diff
@@ -39,6 +39,9 @@ module.exports = {
         // Just do not agree with this one
         'no-negated-condition': 'off',
 
+        'formatjs/enforce-placeholders': 'error',
+        // 'formatjs/no-literal-string-in-jsx': 'error',
```
travjenkins, line 43: Keeping this off for right now. Will make a stand alone PR to fix all of these while turning it on.

src/hooks/journals/useJournalData.ts
```diff
@@ -35,18 +32,28 @@ const useJournalsForCollection = (collectionName: string | undefined) => {
...
+                if (!Array.isArray(dataPlaneListResponse)) {
+                    return Promise.reject(dataPlaneListResponse);
+                }
+
+                return {
+                    journals:
+                        dataPlaneListResponse.length > 0
+                            ? dataPlaneListResponse
+                            : [],
+                };
```
travjenkins, line 56: Made these use identical variable names to make merging these together later easier.

I was going to just have the array and response in the list call but ran into typescript issues and did not want to keep trying to wrangle those.

src/components/shared/Error/Message.tsx
```diff
@@ -25,10 +35,17 @@ function Message({ error }: Props) {
         (typeof error === 'object' && error.hasOwnProperty('code')) ||
         failedAfterRetry;
 
+    // There is a small chance this can happen so adding custom event to track it
+    // Happened before when journal data hook wasn't catching errors
+    const id = error.message ?? FALLBACK;
+    if (id === FALLBACK) {
+        logRocketEvent(CustomEvents.ERROR_MISSING_MESSAGE);
+    }
```
travjenkins, line 43: Make this super safe. This spot should no longer get hit - but in case making sure we have an event just for it. 

We _could_ figure this out with the event/logging above but I want to be able to search for this and not have to manually review sessions.

src/utils/dataPlane-utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    try {
+        // No error so should be fine to unwrap
+        const unwrappedResponse = result.unwrap();
+        return await Promise.resolve(unwrappedResponse);
+    } catch (error: unknown) {
+        // This is just here to be safe. We'll keep an eye on it and possibly remove
+        logRocketConsole(`${key} : unwrapError : `, error);
+        return Promise.reject(error);
+    }
```
travjenkins, line 66: Just snagged this from the shards hook.

kiahna-tucker (commented):
As aforementioned, the error alert is still pretty aggressive. Does this alert still need a full title, subtitle, and description or can the compact, `AlertBox` styling be used here?

### Reference

**Data Preview section error**

journal-error-handling

src/lang/en-US.ts
```diff
@@ -183,6 +183,8 @@ const Error: ResolvedIntlConfig['messages'] = {
     'error.hintLabel': `Hint:`,
     'error.descriptionLabel': `Description:`,
     'error.tryAgain': `Try again and if the issue persists please contact support.`,
+
+    'error.fallBack': `no error details to display`,
```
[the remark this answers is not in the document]
travjenkins: Yes. The point is to keep it super vague so it is not relied on too much as it is simply a failsafe.

kiahna-tucker (approved):
Approved with manual testing since it seems to function and appear as intended at this point in time.

Issue #917, opened 2024-01-08
When we don't have an internet connection, don't display "no captures found"


dyaffe:
**Bug**
We currently display a message stating that the user has no captures, collections or materializations when the user loads those pages without internet.  This has led to users being scared that somehow everything in their account was deleted.

**Expected**
It would be ideal to instead show the user that we weren't able to load their data in that situation.

PR #919, opened 2024-01-11
Adding in handling when a table fails to have data due to network issue

travjenkins:
Make a helper function to check errors the same way

## Issues

https://github.com/estuary/ui/issues/917

### 917

- Added `networkFailed` to the table store query and table statuses to get handled for empty messaging.
- Added `networkFailed` to the stores with hydration (billing needed this)

### Misc

- Handled where the stats column header was stuck in `loading` after an error with hydration
- Added a 'cols' prop to the table columns

### Manually tested

### Automated tests

-   unit testing covered

### Captures
image

### Collections
image

### Materializations
image

### Admin Access
image

### Admin Settings
image

### Admin Billing
image

### Admin Connectors
image

### Details page graph
image


dyaffe:
LGTM

src/components/tables/cells/stats/Header.tsx
```diff
@@ -32,7 +40,12 @@ const StatsHeader = ({ header, selectableTableStoreName }: Props) => {
         <TableCell colSpan={2}>
             <DateFilter
                 header={header}
-                disabled={!hasStats || isValidating || queryCount === 0}
+                disabled={
+                    !hasStats ||
+                    isValidating ||
+                    networkFailed ||
+                    queryCount === 0
+                }
```
travjenkins, line 48: Adding `networkFailed` so the stats column won't sit in loading status forever.

src/services/shared.ts
```diff
@@ -32,5 +32,10 @@ export const logRocketConsole = (message: string, ...props: any[]) => {
 export const FAILED_TO_FETCH = 'failed to fetch';
 export const RETRY_REASONS = [FAILED_TO_FETCH];
 
+export const checkErrorMessage = (
+    checkingFor: string,
+    message?: string | null | undefined
+) => (message ? message.toLowerCase().includes(checkingFor) : false);
+
```
travjenkins, line 39: Quick helper function to make this check consistent. Eventually we might just make a `includesIgnoreCase` function or something like that. But wanted to have this function start by the variables it will normally be used to test.

kiahna-tucker (changes requested):
The collections table on the _Collections_ page appears to be a column short.

**Notes**

The items below do not need to be addressed in this PR.

* The _Configure Alerts_ CTA (under the _Settings_ tab of the _Admin_ page) is enabled but unable to open the dialog because the service its view depends on experienced the network failure.

* The _Generate Invitation_ CTA (in the _Manage Invitations_ dialog) should either be a `LoadingButton` or disabled and the dialog should indicate that there is an asynchronous call in progress.

* The _Grant Access_ CTA (in the _Share Data_ dialog) should either be a `LoadingButton` or disabled and the dialog should indicate that there is an asynchronous call in progress.

### References

**Collections table loading**

failed_network-collections_table-loading_state

<br />
<br />

**Collections table loaded**

failed_network-collections_table-error_state

kiahna-tucker (approved):

Issue #923, opened 2024-01-17
Prevent automatic updates of the connector version of an edited task


kiahna-tucker:
**Last Updated:** Jan. 18, 2024

<hr />

Below are the base requirements for this feature:

* Preserve the connector version of an edited task.

* Use the latest version of a connector when creating a task.

## Dependents

**Last Updated:** Jan. 18, 2024

* [Snowflake connector release.](https://github.com/estuary/connectors/pull/1117) The backend needs to be able to conduct staged roll-outs of new connector versions without breaking the application in order for this feature to be released.

## Additional Context

From Johnny (1/16/2024):

> The smallest possible scoping, as I understand it:
> 
> - we want to be able to add a new, larger `connector_tags` row for Snowflake
> - specs which are using a current row of `connector_tags` should not change their tag if the user happens to make an edit.
> 
> Secondary goal:
> 
> - A new spec creation workflow would pick the largest of the current `connector_tags` tags.

PR #925, opened 2024-01-17
Update how the connector version is determined in workflows

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/923

### 923

* Prevent automatic updates of the connector version of an edited task.

* Use the latest version of a connector when creating a task.

### Manually tested

* Capture Tests:

    * Enter the capture create workflow, select a source connector which has multiple supported versions, complete the form, publish the capture, and observe that the capture uses the latest version of the source connector.

    * Enter the capture create workflow, select a source connector which has multiple supported versions, change the selected connector in the _Details_ section of the form, and observe that the drafted specification uses the latest version of the whichever connector is selected.

    * Enter the capture create workflow, select a source connector which has multiple supported versions, toggle the selected connector in the _Details_ section of the form (i.e., select a different connector then return to the initial connector selected), and observe that the drafted specification uses the latest version of the whichever connector is selected.

    * Enter the edit workflow of a capture that uses an older version of a source connector, edit the resource configuration of an existing binding, edit the endpoint configuration, publish the altered specification, and observe that the version of the source connector was preserved.

* Materialization Tests:

    * Enter the materialization create workflow, select a destination connector which has multiple supported versions, complete the form, and observe that the drafted specification uses the latest version of the destination connector.

    * Enter the materialization create workflow, select a destination connector which has multiple supported versions, change the selected connector in the _Details_ section of the form, and observe that the drafted specification uses the latest version of the whichever connector is selected.

    * Enter the materialization create workflow, select a destination connector which has multiple supported versions, toggle the selected connector in the _Details_ section of the form (i.e., select a different connector then return to the initial connector selected), and observe that the drafted specification uses the latest version of the whichever connector is selected.

    * Enter the edit workflow of a materialization that uses an older version of a destination connector, edit the resource configuration of an existing binding, add/remove a binding, edit the endpoint configuration, and observe that the version of the destination connector was preserved.

### Automated tests

N/A

## Screenshots

N/A

### DIFF
--- a/src/api/connectors.ts
+++ b/src/api/connectors.ts
@@ -1,16 +1,16 @@
 import {
-    ConnectorWithTagDetailQuery,
     CONNECTOR_WITH_TAG_QUERY,
+    ConnectorWithTagDetailQuery,
 } from 'hooks/useConnectorWithTagDetail';
 import {
     CONNECTOR_NAME,
     CONNECTOR_RECOMMENDED,
+    TABLES,
     defaultTableFilter,
     handleFailure,
     handleSuccess,
     supabaseClient,
     supabaseRetry,
-    TABLES,
 } from 'services/supabase';
 import { SortDirection } from 'types';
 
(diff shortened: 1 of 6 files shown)


travjenkins (approved):
lgtm - I was able to test locally and edited a `Hello World v2` and not get an upgrade while also starting create with a `Hello World v3`

For history: had a direct conversation with Kiahna about the solution. I think we could make this a bit simpler by not evaluating the list of tags on change. However, to keep the scope of this change small we are not doing that.

Also, we should eventually push to have the connector_tags sorted via a view.

Issue #954, opened 2024-01-31
Use `update_only` when running background discovers


travjenkins:
```
when doing a background discover due to re-enabling a binding, we should set update_only to true.
```

Conversation from Slack:
**Johnny Graettinger**
Currently we silently add new collections to the draft, even if “auto add new collections” is off, and there’s literally no indication this is happening

**Johnny Graettinger**
Makes it extremely easy to inadvertently add collections new enabled bindings to a capture that you did not intend. (edited) 

**Phil**
Just to clarify, even with update_only = true, we do still add the collections to the draft. We just assume that they'll get pruned

**Johnny Graettinger**
Yes, edited above. The issue is that bindings are silently added and enabled.

Issue #953, opened 2024-01-31
Change capture editing label for collections


travjenkins:
Content change

```
RFC: when editing a capture, should we call them Target Collections instead of Source Collections?
```

kiahna-tucker:
Noting that the _Create a basic Data Flow_ [user guide](https://docs.estuary.dev/guides/create-dataflow/#create-a-capture) in the Estuary documentation still references the former _Output Collections_ section.

Moreover, the aforementioned user guide is considerably outdated. The team recently discussed removing the web application [documentation](https://docs.estuary.dev/concepts/web-app/) from the _Concepts_ section, but I would imagine we would like to keep the _Create a basic Data Flow_ [user guide](https://docs.estuary.dev/guides/create-dataflow/#create-a-capture). Consequently, I would recommend we update this guide as a fast-follow.

PR #955, opened 2024-01-31
Using `update_only` when doing background discover and content change

travjenkins:
## Issues

https://github.com/estuary/ui/issues/953
https://github.com/estuary/ui/issues/954

### 953

- Updated content

### 954

- Set the `update_only` column based on if we are rediscovering
- Allow passing in settings for `update_only` when discovering

### Misc

- Resetting draft id when discover fails to prevent incorrect test/saves

### Manually tested

- 953: viewed edit and create for capture

### Automated tests

- N/A

## Screenshots

Edit content updated
image

Create Content not changed
image

Clicking the `refresh` button still wants to do a discover so the user can get new items
image

### DIFF
--- a/src/components/capture/useDiscoverStartDiscovery.ts
+++ b/src/components/capture/useDiscoverStartDiscovery.ts
@@ -34,7 +34,8 @@ function useDiscoverStartDiscovery(entityType: Entity) {
         async (
             processedEntityName: string,
             encryptedEndpointConfigResponse: any,
-            rediscover?: boolean
+            rediscover?: boolean,
+            updateOnly?: boolean
         ) => {
             // If we are doing a rediscovery and we have a draft then go ahead and use that draft
             //  that way the most recent changes to bindings and endpoints will get added to the draft before rediscovery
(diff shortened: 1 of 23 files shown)


kiahna-tucker (approved):

src/components/capture/useDiscoverStartDiscovery.ts
```diff
@@ -64,9 +67,16 @@ function useDiscoverStartDiscovery(entityType: Entity) {
                 processedEntityName,
                 encryptedEndpointConfigResponse,
                 imageConnectorTagId,
-                newDraftId
+                newDraftId,
+                updateOnly
             );
+
             if (discoverResponse.error) {
+                // If we failed at discovery we need to clear draft ID like we do
```
kiahna-tucker, line 75: This comment is a little muddled to me. What do you think of the following:

```ts
// If discovery fails, clear the draft ID to preserve the _Next_ CTA and allow the user to attempt to discover another time.
```

src/components/tables/cells/logs/MessageCell.tsx
```diff
@@ -21,7 +21,7 @@ function MessageCell({ fields, message }: Props) {
             </Typography>
 
             {fields ? (
-                <Typography component="div" sx={BaseTypographySx}>
+                <Typography component="div" sx={{ ...BaseTypographySx }}>
```
kiahna-tucker, line 24: The spread operator should not be needed here.

Issue #860, opened 2023-12-11
Add support for the UI to allow the user to use `backfill` counter instead of renaming


travjenkins:
### Request
We no longer need to always rename collections and instead tick up the backfill counter

### Thoughts/Discussions
This acts a little different between entities we need to make sure the verbiage/content ensures users understand what is happening

Captures - Re-fetch everything from the source. This is like a repair action
Materializations - Drop a table and recreate it. This prevents any conflicts.

Since this action has pretty big ramifications we want to make sure the user has to actually take an action (click) to make this happen. It should **not** be automatic. There can be extra charges for this - however that does not seem like a major deal so far. We could make this option "opt out" if we wanted to and that is probably fine.

The verbiage/action should be called `Backfill`/`Drop and Backfill` as many users will know what that means. We do **not** want the user thinking of the counter as a version number.

More than likely we will replace this option on Captures and add a similar one to Materializations
![image](https://github.com/estuary/ui/assets/270078/0ceffdef-1549-428f-9758-000642ccaeb6)

kiahna-tucker:
BLOCKED: This issue cannot be worked on until the backfill counter is used in schema evolution.

kiahna-tucker:
On Jan. 18, 2024, it was decided to use an outlined toggle button to increment the backfill counter manually. The button will use the primary color as the fill and outline color and have the label _Backfill_. It will be located in a section directly beneath the resource configuration form in the _Output Collections_ section of the overarching form.

PR #956, opened 2024-01-31
Enable the backfill counter to be manually incremented in edit workflows

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/860

### 860

-   changes...

### Manually tested

### Automated tests

-   unit testing covered

## Screenshots


src/components/shared/OutlinedToggleButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+        },
+    };
+
+    if (defaultStateSx) {
+        sx = { ...sx, ...defaultStateSx };
+    }
+
+    if (disabledStateSx) {
+        sx = { ...sx, [`&.${toggleButtonClasses.disabled}`]: disabledStateSx };
+    }
```
travjenkins, line 56: This is fine for now - but we need to handle this a bit better by just consuming the `sx` that can be passed in.

src/components/editor/Bindings/ManualBackfill.tsx
```diff
@@ -0,0 +1,149 @@
...
+                </Stack>
+            </Stack>
+
+            <OutlinedToggleButton
+                value={increment}
+                selected={selected}
+                disabled={formActive}
+                onClick={(event, checked: string) => {
+                    event.preventDefault();
+                    event.stopPropagation();
```
travjenkins, line 138: I don't think we need there here as I don't think the container isn't listening for a click.

travjenkins (approved):
small comments that can get handled with a fast follow

lgtm

src/components/shared/OutlinedToggleButton.tsx
```diff
@@ -57,6 +58,7 @@ function OutlinedToggleButton({
 
     return (
         <ToggleButton
+            {...props}
```
travjenkins, line 61: This is fine now but this should be at the end to allow a user to override whatever they want.

src/utils/workflow-utils.ts
```diff
@@ -24,6 +24,10 @@ import { ConnectorConfig } from '../../flow_deps/flow';
 // This is the soft limit we recommend to users
 export const MAX_BINDINGS = 300;
 
+export const getBackfillCounter = (binding: any): number => {
```
travjenkins, line 27: :+1:

src/stores/ResourceConfig/Store.ts
```diff
@@ -712,15 +744,36 @@ const getInitialState = (
...
+                let selectedSpec:
+                    | LiveSpecsExtQuery
+                    | DraftSpecsExtQuery_ByDraftId = liveSpecs[0];
+
+                if (draftId) {
+                    const { data: draftSpecs, error: draftSpecError } =
+                        await getDraftSpecsByDraftId(draftId, entityType);
+
+                    if (draftSpecError) {
+                        setHydrationErrorsExist(true);
```
travjenkins, line 765: How does this work? Cause this is totally different logic now.

Even if there is an error the prefill of resource config will happen.

kiahna-tucker: The live spec is used as a fallback... which is new behavior. I can fix this.

kiahna-tucker: This should be resolved.

travjenkins (approved):
changes look good

Issue #962, opened 2024-02-01
Enable setting the backfill counter of all task bindings with a single action


kiahna-tucker:
**Last Updated:** Feb. 6, 2024

<hr />

Below are the base requirements for this feature:

* Add a _Backfill_ section directly above the bindings selector in edit workflows that contains a header, short description, and an outlined, toggle button with the label _Backfill_. This button should have the same styling as the binding-specific _Backfill_ button.

* Update the backfill counter of applicable, drafted bindings when the new backfill button is clicked. The toggle button state triggers are as follows:
    * **Disabled:** the button will be in a disabled state when an asynchronous task that updates the server is in progress and when the drafted specification does not contain at least one enabled binding.
    
    * **Deselected:** the button will be in a deselected state by **default** and when the backfill counter of any drafted binding **has not** been incremented in the editing session.
    
    * **Selected:** the button will be in a selected state when the backfill counter of all drafted bindings **have** been incremented in the editing session.

**NOTE:** The backfill counter of a drafted binding does **not** need to be _reset_ if the binding is disabled post-incrementation in the editing session.

kiahna-tucker:
## Questions

* What label should be used on the new backfill button (e.g., _Backfill_, _Backfill All_, etc.)?

* The present UX makes it easy to increment the backfill counter of all drafted, bindings, but it does not provide a means to _reset_ the backfill counter of all drafted, bindings. Therefore, I propose that the overarching, _Backfill_ section include a standard, outlined button that will reset the backfill counter of all drafted bindings that **have** been incremented in the editing session. What does everyone think about this proposal? If accepted, what label should be used on the reset button (e.g., _Reset_, _Reset All_, _Undo_, _Undo All_, _Unset_, _Unset All_, etc.)?

dyaffe:
"Backfill" for the button would make sense to me with the subtext "Trigger a backfill of all collections from the source when published."

Just to confirm, when you say "reset the backfill counter", you mean if it was previously 0, someone clicked "backfill" and all bindings went to 1, they could click the button and have them all go back to zero?

Why not make it work like it does at the binding level?  It appears that the button has state and can go back and forth between selected and unselected.  I personally like that UX!

kiahna-tucker:
> Just to confirm, when you say "reset the backfill counter", you mean if it was previously 0, someone clicked "backfill" and all bindings went to 1, they could click the button and have them all go back to zero?

Correct.

[...]

PR #965, opened 2024-02-05
Enable setting the backfill counter of all task bindings with a single action

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/962
https://github.com/estuary/ui/issues/973

### 962

* Add a _Backfill_ section directly above the bindings selector in edit workflows that contains a header, short description, and an outlined, toggle button with the label _Backfill_. This button should have the same styling as the binding-specific _Backfill_ button.

* Update the backfill counter of applicable, drafted bindings when the new backfill button is clicked. The toggle button state triggers are as follows:
    * **Disabled:** the button will be in a disabled state when an asynchronous task that updates the server is in progress and when the drafted specification does not contain at least one enabled binding.
    
    * **Deselected:** the button will be in a deselected state by **default** and when the backfill counter of any drafted binding **has not** been incremented in the editing session.
    
    * **Selected:** the button will be in a selected state when the backfill counter of all drafted bindings **have** been incremented in the editing session.

**NOTE:** The backfill counter of a drafted binding does **not** need to be _reset_ if the binding is disabled post-incrementation in the editing session.

### 973

* Add disabled state styling to the `OutlinedToggleButton`.

### Manually tested

### Automated tests

N/A

## Screenshots

**Capture edit | Bindings enabled | Backfill unset**

pr_screenshot-965-manual_backfill_all-capture-edit-unselected

<br />
<br />

**Capture edit | Bindings enabled | Backfill set**

pr_screenshot-965-manual_backfill_all-capture-edit-selected

<br />
<br />

**Capture edit | Bindings disabled**

pr_screenshot-965-manual_backfill_all-capture-edit-all_bindings_disabled

<br />
<br />

**Materialization edit | Bindings enabled | Backfill unset**

pr_screenshot-965-manual_backfill_all-materialization-edit-unselected

<br />
<br />

**Materialization edit | Bindings enabled | Backfill set**

pr_screenshot-965-manual_backfill_all-materialization-edit-selected

<br />
<br />

**Materialization edit | Bindings disabled**

pr_screenshot-965-manual_backfill_all-materialization-edit-all_bindings_disabled

<br />
<br />

**Disabled primary colored outlined toggle button**

pr_screenshot-965-manual_backfill_all-disabled_button_styling


src/components/editor/Bindings/Backfill/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    return (
+        <Box sx={{ mt: 3 }}>
+            <Stack direction="row" spacing={1} sx={{ mb: 2 }}>
+                <Stack spacing={1}>
+                    <Typography
+                        variant={
+                            bindingIndex === -1 ? 'formSectionHeader' : 'h6'
+                        }
+                        sx={{ mr: 0.5 }}
```
travjenkins, line 162: I do not think this `sx` is needed as it seems to not do anything.

src/components/editor/Bindings/Backfill/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+            setBackfilledCollections,
+            setFormState,
+            updateBackfillCounter,
+        ]
+    );
+
+    return (
+        <Box sx={{ mt: 3 }}>
+            <Stack direction="row" spacing={1} sx={{ mb: 2 }}>
+                <Stack spacing={1}>
```
travjenkins, line 157: What is driving this outer `stack`?

src/components/editor/Bindings/Backfill/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+                    handleClick(checked === 'true' ? 'false' : 'true');
+                }}
+            >
+                <FormattedMessage id="workflows.collectionSelector.manualBackfill.cta.backfill" />
+
+                {selected ? (
+                    <Check style={{ marginLeft: 8, fontSize: 13 }} />
+                ) : null}
+            </OutlinedToggleButton>
```
travjenkins, line 189: Since no content changes when this is enabled/disabled we should probably include _some_ aria/accessibility handling here. 

I think adding some aria content for the check to something like `enabled` would work.

src/components/editor/Bindings/Backfill/useUpdateBackfillCounter.ts
```diff
@@ -0,0 +1,135 @@
...
+
+            const spec: Schema = draftSpec.spec;
+
+            if (bindingMetadataExists) {
+                bindingMetadata.forEach(({ bindingIndex }) => {
+                    if (bindingIndex > -1) {
+                        spec.bindings[bindingIndex] = evaluateBackfillCounter(
+                            spec.bindings[bindingIndex],
+                            increment
+                        );
```
travjenkins, line 93: This and the call below should not be setting the entire binding. This should just update the specific property we're controlling.

src/components/editor/Bindings/Backfill/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+                    increment,
+                    bindingMetadata
+                ).then(
+                    () => {
+                        const targetCollection = singleBindingUpdate
+                            ? currentCollection
+                            : undefined;
+
+                        setBackfilledCollections(increment, targetCollection);
+                        setFormState({ status: FormStatus.UPDATED });
```
travjenkins, line 129: This seems to not be resetting the state properly. I think it needs to pass an empty `error` maybe? 

Once you get a network error if you then click again and it works the error just hangs around. We should probably reset the state of failed/error when they click?
image

travjenkins: Playing with this for a second... I think we should probably reset the error up above to make the UX a bit more clear if another error happens.

src/stores/ResourceConfig/Store.ts
```diff
@@ -443,7 +446,7 @@ const getInitialState = (
     },
 
     prefillBackfilledCollections: (liveBindings, draftedBindings) => {
-        const { addBackfilledCollection } = get();
+        const { addBackfilledCollections } = get();
```
travjenkins, line 449: Why are we having this stand alone function that loops through all the bindings?

We are already calling `prefillResourceConfig` right before calling this which means we loop through all the bindings twice in a row.

src/stores/ResourceConfig/Store.ts
```diff
@@ -460,33 +463,59 @@ const getInitialState = (
...
+            produce((state: ResourceConfigState) => {
+                const collections: string[] = targetCollection
+                    ? [targetCollection]
+                    : state.collections
+                    ? state.collections
+                    : [];
+
+                state.backfilledCollections =
+                    increment === 'true'
+                        ? union(state.backfilledCollections, collections)
```
travjenkins, line 508: Unless we need the sorting - I don't think we need to use `union` here. I think `concat` would be okay.

src/components/editor/Bindings/Backfill/useUpdateBackfillCounter.ts
```diff
@@ -0,0 +1,135 @@
...
+                spec.bindings.forEach((binding: Schema, index: number) => {
+                    const collection = getCollectionName(binding);
+
+                    const shouldIncrement =
+                        !backfilledCollections.includes(collection) &&
+                        increment === 'true';
+
+                    const shouldDecrement =
+                        backfilledCollections.includes(collection) &&
+                        increment === 'false';
```
travjenkins, line 106: We should just call the `includes` check once and save it off in a variable.

travjenkins (changes requested):
Some call outs. All of `em do not need handled right now. I think the ones that need handled I marked with a :heart:

src/stores/ResourceConfig/Store.ts
kiahna-tucker: Would you like to conflate the scope of `prefillResourceConfig`?

src/components/editor/Bindings/Backfill/index.tsx
kiahna-tucker: This property was set in an early permutation of the first, manual backfill PR. The button was located elsewhere and required a margin for alignment. Will remove.

src/components/editor/Bindings/Backfill/index.tsx
kiahna-tucker: Likely a C&P error from where this code was originally sourced in the first, manual backfill PR.

src/components/editor/Bindings/Backfill/index.tsx
kiahna-tucker: Not sure why I assumed that the form state store cleared errors when transitioning from an idle to an active state.

src/components/editor/Bindings/Backfill/useUpdateBackfillCounter.ts
kiahna-tucker: This is mildly a case of logical semantics, but agreed.

src/stores/ResourceConfig/Store.ts
travjenkins: Yeah - as best we can I would like that big initial looping to only be done once if possible.

travjenkins (approved):
lgtm

tested a bit and seemed alright with 1k collections

Issue #939, opened 2024-01-22
Allow user to delete all collections related to Capture


travjenkins:
When a user is deleting a Capture we should allow them to also delete all collections.

travjenkins:
Present an option in the modal. I don't think it would need to calls. We would want to make the delete all in a single draft/publication. I believe the backend allows deleting collections while also deleting the capture making 'em. But if not - two draft/publishes is probably not too bad.

kiahna-tucker:
Would we like to present an option to delete associated collections **for each** capture listed in the confirmation modal? Or should there be a **single** option that allows the user to opt into collection deletion that will apply to all captures listed in the confirmation modal?

If an option to delete associated collections will be shown for each listed capture, will this option be presented as a nested list item? If so, should the option be able to be hidden? If so, should it be hidden by default and presented when the capture name and/or an icon button (e.g., settings icon, etc.) inline with the name is clicked?

travjenkins:
Good question... my gut reaction would be to have it be a nested checkbox or something. But when doing this we'd probably want to then handle "select all" and "select none" and that always feels odd to me in a nested list.
![image](https://github.com/estuary/ui/assets/270078/4ff54d8f-318d-4edf-9364-da2218f81b24)

Maybe we just try to follow a normal pattern of opting in on groups with a more table like display? Obviously we'd need to handle width and still the "all" or "none" problem but I feel that sometimes is well suited in a table.
![image](https://github.com/estuary/ui/assets/270078/49c31420-9ca2-40bf-996c-114b5c30ef7d)

I could very well see us only supporting the single option path. Like a new third button or something.

kiahna-tucker:
>my gut reaction would be to have it be a nested checkbox or something. But when doing this we'd probably want to then handle "select all" and "select none" and that always feels odd to me in a nested list.

This idea was the first thing that came to mind and I had the same reservation. Nevertheless, if we plan to provide this option for each capture, presenting it as a nested list item feels like the most appropriate UX given the severity of the action. It feels like an escape hatch for users that are very familiar with their flow system therefore discretion is preferred IMO. The nested list item provides room to communicate a warning in the event the collection is consumed by other tasks as well, if we so choose.

## Design Reference

**Nested list item hidden | No captures opted into collection deletion**

<img width="462" alt="confirmation_modal-icon-default" src="https://github.com/estuary/ui/assets/77648584/6fed3e22-d4ed-49c4-ac03-ff7041a6a551">

<br />
<br />

[...]

PR #978, opened 2024-02-12
Enable the deletion of a capture and associated collections

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/939

### 939

* Add an option to the capture deletion confirmation modal which enables the removal of a capture and its associated collections.

* Update the irreversible action warning (ref: `'common.noUnDo'`, [request](https://github.com/estuary/ui/issues/939#issuecomment-1936057130)).

### Manually tested

-  Enabled, disabled, and deleted captures, collections, and materializations.

### Automated tests

N/A

### Capture Deletion Confirmation Modal

pr_screenshot-965-manual_backfill_all-UNSOLICITED

### Other Deletion Confirmation Modals

**Collections table**

pr_screenshot-978-delete_captured_collections-collection_deletion_modal

<br />
<br />

**Materializations table**

pr_screenshot-978-delete_captured_collections-materialization_deletion_modal

<br />
<br />

**Active invitations table**

pr_screenshot-978-delete_captured_collections-invitation_deletion_modal


src/lang/en-US.ts
```diff
@@ -29,7 +29,7 @@ const CommonMessages: ResolvedIntlConfig['messages'] = {
     'common.loggedOut': `You have been logged out. Please log back in.`,
     'common.missing': `N/A`,
     'common.none': `none`,
-    'common.noUnDo': `This action cannot be undone.`,
+    'common.noUnDo': `All items will be permanently deleted and this action cannot be undone. Please review the list to continue.`,
```
kiahna-tucker, line 32: This piece of content is only passed to the `title` property of the `AlertBox` component (e.g., _src/components/tables/RowActions/Delete/Confirmation.tsx_). This _feels_ like a misuse of the property given it is more body copy than a title/header. Any opposition to displaying this content via a child of the `AlertBox` component?

src/components/tables/RowActions/Shared/NestedListItem.tsx
kiahna-tucker, line 1: This list item nesting logic closely resembles that used in the page-based, derivation create workflow, therefore a shared component should be defined. That said, the page-based, derivation create workflow is broken which would make testing compatibility impossible. Should an attempt be made to share code between the two components while the workflow is out of commission?

Nevertheless, a TODO comment should be left behind addressing the opportunity for code reusability.

src/components/tables/RowActions/Shared/Confirmation.tsx
```diff
@@ -1,29 +1,48 @@
...
     selected,
-}: RowActionConfirmationprops) {
+    selectableTableStoreName,
+    settings,
+}: RowActionConfirmationProps) {
+    const renderListItems =
+        settings && settings.length > 0 && selectableTableStoreName
+            ? (value: string) => (
+                  <NestedListItem
+                      key={`list-item-${value}`}
```
kiahna-tucker, line 27: The legacy list item key can be used here.

src/components/tables/RowActions/Shared/NestedListItem.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+    const setActionSettings = useZustandStore<
+        SelectableTableStore,
+        SelectableTableStore['setActionSettings']
+    >(
+        selectableTableStoreName,
+        selectableTableStoreSelectors.actionSettings.set
+    );
+
+    const [open, setOpen] = useState<boolean>(false);
```
travjenkins, line 63: Should use `material-ui-popup-state` for states like this.

src/lang/en-US.ts
travjenkins: Why did we replace this title? As stated here - seems like this is much more of a body message and not a title. Seems we should keep the title simple and then use this as a child.

kiahna-tucker: Dave requested a change the content of this warning (see link to issue comment in PR description). Moving this to a child of the `AlertBox` component in the two instances where it is used.

src/lang/en-US.ts
```diff
@@ -623,7 +623,8 @@ const Captures: ResolvedIntlConfig['messages'] = {
     'capturesTable.cta.new': `New Capture`,
     'capturesTable.filterLabel': `Filter captures`,
     'capturesTable.disableEnable.confirm': `All items listed below will be {setting}.`,
-    'capturesTable.delete.confirm': `All items listed below will be deleted forever. Please review before continuing.`,
+    'capturesTable.delete.confirm': `Collections used by active materializations will be skipped.`,
```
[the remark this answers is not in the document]
kiahna-tucker: > Collecting artifacts for the description of PR 978, but the code is stable with the exception of the four callouts.

If you are asking why the message ID was not updated, the commit containing this change has not been pushed to the remote branch. The quoted DM states that the four comments left on this PR by me are pending changes.

src/lang/en-US.ts
travjenkins: I feel like the requested change can be done while leaving a title. But whatever y'all decided is fine with me.

kiahna-tucker: If you are fine with unused content being left behind, then yes, this can be updated now. Will do.

src/components/tables/RowActions/Delete/Button.tsx
```diff
@@ -14,12 +15,32 @@ interface Props {
...
+                    messageId={
+                        isCapture ? 'capturesTable.delete.confirm' : null
+                    }
+                />
+            }
+            messageID="cta.delete"
             renderProgress={(item, index, onFinish) => (
                 <UpdateEntity
                     key={`Item-delete-${index}`}
+                    deleteCollections={isCapture}
```
travjenkins, line 43: I think we're okay not passing this in. Since we are already passing in the store name I think we could do this check inside the component. Are we checking it here and passing it down since we are using it to set `settings`?

kiahna-tucker: No. Will remove.

src/components/tables/RowActions/Shared/Confirmation.tsx
```diff
@@ -1,29 +1,48 @@
...
-                        <ListItem key={`confirmation-selected-items-${value}`}>
-                            <Typography variant="h6" component="span">
-                                {value}
-                            </Typography>
-                        </ListItem>
-                    );
-                })}
-            </List>
+
+            <List>{selected.map(renderListItems)}</List>
```
travjenkins, line 45: Since the `NestedListItem` is rendering divs - we should mark this `List` with a `div` component and then update the `ListItem` up above.

src/components/tables/RowActions/Delete/Button.tsx
```diff
@@ -14,9 +15,28 @@ interface Props {
 function DeleteButton({ selectableTableStoreName }: Props) {
     const generator = () => null;
 
+    const isCapture =
+        selectableTableStoreName === SelectTableStoreNames.CAPTURE;
```
travjenkins, line 19: This is okay for now... but I think we should eventually move some setting into the store that controls all this. It could be very direct like `enableRemoveCollectionsOnDelete` or even take it further and just allow any action of some kind of secondary action... however not really sure making it that flexible would really get us anything.

src/components/tables/RowActions/Shared/NestedListItem.tsx
travjenkins: Yeah - I would make it shared now and we can add ensuring this works with it to the tasks we need to get derivation wired back up.

src/stores/Tables/Store.ts
```diff
@@ -340,6 +354,35 @@ export const getInitialState = (
...
+                    const existingEl = state.actionSettings[setting] ?? [];
+
+                    if (addOperation) {
+                        state.actionSettings[setting] =
+                            existingEl.concat(values);
+                    } else if (existingEl.length > 0) {
+                        state.actionSettings[setting] = existingEl.filter(
+                            (el) => !values.includes(el)
+                        );
+                    }
```
travjenkins, line 369: Just thinking... I think we'd probably want to "invert" this and have a loop go over all the `values` passed in. This does not matter right now because it is always a single element array... but in the future if we allow a "Select All" functionality I think looping over just the array that includes changes should be best.

src/components/tables/RowActions/Shared/Confirmation.tsx
```diff
@@ -1,29 +1,51 @@
...
+            : (value: string) => (
+                  <ListItem
+                      component="div"
+                      key={`confirmation-selected-items-${value}`}
+                  >
+                      <Typography component="span" sx={{ fontWeight: 500 }}>
+                          {value}
+                      </Typography>
+                  </ListItem>
+              );
```
travjenkins, line 42: Given how different these are... I wonder if we should just not even try to make them run through the same code? Would it be better to just pass the list renderer component as a prop?

travjenkins (approved):
For what is done here it looks good. Any of the callouts I think could be handles later if we wanted to.

src/lang/en-US.ts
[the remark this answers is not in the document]
kiahna-tucker: It was decided to append the checkbox label content (i.e., `'capturesTable.delete.removeCollectionsOption'`) with the collection deletion warning.

dyaffe:
My only suggestion is to remove the gear and always have the checkboxes show if possible.

Issue #935, opened 2024-01-19
Materialization edit: bindings selector breaks when using partition selectors


psFried:
**Bug**

This was observed when editing an existing materialization in the UI that has multiple bindings for the same collection, just with different partition selectors. For example:

```
bindings:
  - source:
      name: acmeCo/foo
      partitions: { include: { part: [ 'A' ] } }
    resource: { table: foo_a }
  - source:
      name: acmeCo/foo
      partitions: { include: { part: [ 'B' ] } }
    resource: { table: foo_b }
```

**Expected**

Expected to be able to edit the resource config for each binding, and for the spec to not get mutated without me making any changes.

What actually happens is that the resource config editor does not change when clicking between the bindings. It always shows the same one. Also, the draft spec seems have been changed such that the same resource config is duplicated on both bindings.

**Screenshots**

Sorry, but the only example I have is on a production materialization, and I don't want to post an image of their materialization publicly.

**Desktop (please complete the following information):**
 - OS: mac
 - Browser: firefox

travjenkins:
Makes sense. We key off the name today. We'll need to update so we also key off the index. Possibly just switch over and only key off that.

travjenkins:
This also causes the list to get really crazy when taking actions and showing more versions of this than are bound.

Ex: only 2 `whatever` bindings exist
![image](https://github.com/estuary/ui/assets/270078/4e2115a1-742f-46f2-a829-e2281f996327)

PR #986, opened 2024-02-16
Update binding modelling in client state

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/935

### 935

* Enable the editing of multiple bindings that map to the same collection.

* Create a binding store that houses binding-related state.

    * Notable store state:

        * **`backfilledBindings`** -- the analog of resource config store state property, `backfilledBindings`, which is an array of binding UUIDs.

        * **`bindings`** -- the analog of resource config store state property, `collection`, which is dictionary keyed by the name of the collection associated with a binding with the array of related binding UUIDs as a value.
        
        * **`currentBinding`** -- the analog of resource config store state property, `currentCollection`, which is an object containing a reference to the binding UUID and name of the collection associated with the _selected_ binding.
        
        * **`resourceConfigs`** -- the analog of resource config store state property, `resourceConfig`, which is dictionary keyed by binding UUID with a modified `ResourceConfig` object.

    * Notable store actions:

        * **`addEmptyBindings`** -- the analog of resource config store action, `preFillEmptyCollections`.

        * **`evaluateDiscoveredBindings`** -- the analog of resource config store action, `evaluateDiscoveredBindings`, which encapsulates the functionality of `setDiscoveredCollections` and `resetConfigAndCollections`.

        * **`prefillBindingDependentState`** -- the combination of resource config store actions, `prefillResourceConfig` and `prefillBackfilledCollections`, with expanded scope that initializes the time travel-related state in materialization workflows. Formerly, the latter was done in `useInitializeTaskDraft`.

        * **`prefillResourceConfigs`** -- the partition of the second half of resource config store action, `setResourceConfig`, which updated the resource config dictionary when a capture is linked to a materialization (i.e., _src/components/materialization/SourceCapture/AddSourceCaptureToSpecButton.tsx_) and bindings are added to the specification via the collection selector (i.e., _src/components/editor/Bindings/UpdateResourceConfigButton.tsx_).
        
        * **`removeDiscoveredBindings`** -- the analog of resource config store action, `resetResourceConfigAndCollections`.
  
        * **`updateResourceConfig`** -- the combination of resource config store actions, `updateResourceConfig` and the partition of the first half of `setResourceConfig`, which updated the resource config dictionary when editing a binding's resource config form (i.e., _src/components/collection/ResourceConfigForm.tsx_).

* Remove the resource configuration store.

* Move field selection and time travel-related state from the bindings editor store into the new bindings store.

* Update the logic of `getBindingIndex`, `generateTaskSpec`, `useServerUpdateRequiredMonitor` to account for the existence of multiple bindings that map to the same collection.

* Use the binding UUID as the `path` of the Monaco editor used to display/edit the collection specification in the right-hand pane of the binding selector.

* Preserve the value of the `active` hydration-related, binding state property by default when the state is reset. An optional parameter, `resetActive`, was added to the definition of the `resetState` binding store action to enable resetting the aforementioned property when necessary. Presently, `active` should only be reset when a given workflow is exited.

### Manually tested

A myriad of binding-related scenarios in the create and workflows were tested; listing them entirely would be tedious and futile. Consequently, this section will document any scenarios that require retesting after the following commit hash: `fbaaaa6151bbf3ffaf0c0f8e44b9316429e8d6fa`.

### Automated tests

* Updated the binding editor store initialization snapshot after removing field selection and time travel state slices.

* Lightly expanded the test coverage of the `getBindingIndex` and `generateTaskSpec` workflow utils.

## Screenshots

N/A


src/utils/workflow-utils.ts
```diff
@@ -132,10 +147,14 @@ export const updateFullSource = () => {};
...
+    resourceConfigs: ResourceConfigDictionary,
+    resourceConfigServerUpdateRequired: boolean,
+    bindings: Bindings,
     existingTaskData: DraftSpecsExtQuery_ByCatalogName | null,
-    sourceCapture: string | null,
-    fullSource: FullSourceDictionary | null
+    options: {
+        fullSource: FullSourceDictionary | null;
+        sourceCapture: string | null;
+    }
```
travjenkins, line 157: These should stay params and not nested in an `options` object.

kiahna-tucker: This was a _quick_ way to handle the fact that we have exceeded the maximum number of parameters. Ultimately, I wanted to move this into a hook but felt that may be too large of a change for this PR.

src/utils/workflow-utils.ts
```diff
@@ -146,60 +165,76 @@ export const generateTaskSpec = (
...
+                // Check if disable is a boolean otherwise default to false
+                const bindingDisabled = isBoolean(disable) ? disable : false;
+
+                // See which binding we need to update
+                const existingBindingIndex = resourceConfigServerUpdateRequired
+                    ? getBindingIndex(
+                          draftSpec.bindings,
+                          collectionName,
+                          iteratedIndex
+                      )
```
travjenkins, line 188: This is okay for now. But this approach needs to change somewhat quickly.

We should not be calling this inside of a loop. Instead we should handle fetching what collections are at what index before the loop. That way we are not running through a mapping and filtering of all the bindings every time for every binding.

kiahna-tucker: Sure. This is worth a discussion around expectations; primarily, whether `getBindingIndex` should be used in this function at all. I am inclined to say is should not be used here.

As aforementioned, I aspired to make as minimal changes as possible with the binding model refactoring. The binding index was evaluated in a similar fashion previously, so I lightly modified it in this PR.

travjenkins: Yeah - we really should strive to not use it here at all. Can you throw in a `TODO` here and explain what the future work could be.

src/hooks/useServerUpdateRequiredMonitor.ts
```diff
@@ -1,69 +1,79 @@
...
 
-                    // Snag the name so we know which resource config to check
-                    const collectionName = getCollectionName(
-                        binding[collectionNameProp]
-                    );
+                            const existingBindingIndex = getBindingIndex(
+                                draftSpecs[0]?.spec.bindings,
+                                collection,
+                                targetBindingIndex
+                            );
```
travjenkins, line 39: This obviously can have the same issue as the other loop in the generate call. However, since this call happens anytime something like `disable` is changed it is not behind a loading state.

When there are a lot of bindings the page locks up way too much to leave this getting called every loop.

src/components/tables/cells/fieldSelection/FieldActions.tsx
```diff
@@ -4,44 +4,50 @@ import {
...
     const formActive = useFormStateStore_isActive();
 
-    const selectedValue = useMemo(() => selections[field], [field, selections]);
+    const selectedValue = useMemo(
+        () =>
+            Object.hasOwn(selections, bindingUUID)
+                ? selections[bindingUUID][field]
+                : null,
+        [bindingUUID, field, selections]
+    );
```
travjenkins, line 50: Eventually we should pass `bindingUUID` into the store's hook when fetching `bindingUUID` since we never reference it any other way.

travjenkins (changes requested):
We need to do performance work on this before we can relase it.

Against `main` a capture with 18k collections takes ~10 seconds to be interactive. On this branch that takes ~40-50 seconds.

src/components/editor/Store/hooks.ts
```diff
@@ -533,6 +533,7 @@ export const useEditorStore_queryResponse_mutate = (
 
 export const useEditorStore_queryResponse_draftedBindingIndex = (
     collection: string | null,
+    targetBindingIndex: number,
     params?: SelectorParams | undefined
 ) => {
     const localScope = params?.localScope;
```
travjenkins, line 539: With where this is getting called I do not think we need to support the `localScope` stuff anymore.

src/components/editor/Store/hooks.ts
```diff
@@ -533,6 +533,7 @@ export const useEditorStore_queryResponse_mutate = (
 
 export const useEditorStore_queryResponse_draftedBindingIndex = (
```
travjenkins, line 534: I _think_ we should also try getting this to be a shallow hook.

package.json
```diff
@@ -65,6 +65,7 @@
         "immer": "^9.0.21",
         "jose": "^4.15.5",
         "json-ptr": "^3.1.1",
+        "json-refs": "^3.0.15",
```
travjenkins, line 68: Do you know why this is showing up here? This is already in main.

travjenkins: I see what I did. Ignore this LOL

src/hooks/useServerUpdateRequiredMonitor.ts
```diff
@@ -6,14 +6,15 @@ import {
...
-import { getBindingIndex, getDisableProps } from 'utils/workflow-utils';
+import { getCollectionName, getDisableProps } from 'utils/workflow-utils';
 
 const useServerUpdateRequiredMonitor = (draftSpecs: DraftSpecQuery[]) => {
     const bindings = useBinding_bindings();
     const resourceConfigs = useBinding_resourceConfigs();
     const setServerUpdateRequired = useBinding_setServerUpdateRequired();
 
     const resourceConfigUpdated = useMemo(() => {
+        console.log('here');
```
travjenkins, line 17: logging

src/hooks/useServerUpdateRequiredMonitor.ts
```diff
@@ -23,27 +24,17 @@ const useServerUpdateRequiredMonitor = (draftSpecs: DraftSpecQuery[]) => {
...
-                            const targetBindingIndex = Object.hasOwn(
-                                bindings,
-                                collection
-                            )
-                                ? bindings[collection].findIndex(
-                                      (uuid) => uuid === bindingUUID
-                                  )
-                                : -1;
+                            const expectedBindingIndex =
+                                resourceConfigs[bindingUUID].meta.bindingIndex;
```
travjenkins, line 28: If `resourceConfigs[bindingUUID]` is guaranteed to exist then down below we should remove the optional chaining down below.

src/utils/workflow-utils.ts
```diff
@@ -54,16 +54,19 @@ export const getCollectionName = (binding: any) => {
...
-            .filter((bindingIndex) => bindingIndex > -1) ?? [];
+    const matchedCollectionIndices = existingBindings
+        .map((binding: any, index) => {
+            if (getCollectionName(binding) === collectionName) {
+                return index;
+            } else {
+                return -1;
+            }
+        })
+        .filter((bindingIndex) => bindingIndex > -1);
```
travjenkins, line 69: Can you declare this function and then use it. Trying to add all the little tweaks.

travjenkins (commented):
A couple questions. Also - there is a quality check failure.

src/hooks/useServerUpdateRequiredMonitor.ts
```diff
@@ -54,6 +45,11 @@ const useServerUpdateRequiredMonitor = (draftSpecs: DraftSpecQuery[]) => {
                                     return true;
                                 }
 
+                                // Ensure the associated collection matches before comparing binding resources.
+                                if (collection !== getCollectionName(binding)) {
+                                    return true;
+                                }
```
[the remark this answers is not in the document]
kiahna-tucker: In the absence of `getBindingIndex`, it is not guaranteed that the existing binding extracted using the `expectedBindingIndex` is _associated with_ the same collection as the target resource configuration. This condition serves as a logical short circuit for the function and, arguably, should be the first evaluation performed.

travjenkins (approved):
lgtm - please fix things we discussed in person. Marking as approved so you can merge whenever you want.

Issue #1012, opened 2024-03-05
Having multiple versions of connector can render wrong resource config form


travjenkins:
**Bug**
When there are multiple connector versions fetched in the UI the EndPoint Config renders the latest one. However, the resource config will sometimes render the improper one.

**Expected**
The resource config renders properly

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Additional context**
N/A

PR #1013, opened 2024-03-07
Have resource config read the connector tag id from details

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1012

### 1012

- Update the Resource Schema hydration to fetch based on the `connector_tags` `id` column
- We read this `id` from the Details Form so that we know we're using the same settings that endpoint config is using

### Manually tested

- Capture create/edit
- Materialization create/edit

### Automated tests

- n/a

## Screenshots

image

image


kiahna-tucker (approved):

Issue #1011, opened 2024-03-05
UI should be more robust to inconsistent `connector_tags` rows


travjenkins:
The UI trusts the content in the connector and connector_tag tables a lot. We need to make the UI a bit more resilient and make sure the columns we are fetching are usable.

Stuff like...
ensuring the schemas are parasable for JSONForms/AJV
the id columns are valid uuids
etc.

PR #1014, opened 2024-03-07
Only fetch connector/tags that have endpoint/resource/tag

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1011

### 1011

- Check for the existence of the important columns we want
- Since the main search is filtered we do not add extra checks on the queries for schemas

### Misc
- make the check if there are connectors a smaller query in hopes the query is a bit quicker

### Manually tested

- Land on the views that run these queries and check results

### Automated tests

- n/a

## Screenshots

Captures Create List - hubspot does not display
image
image
image
image

Searching still hides invalid connectors
image

Admin page and search hides invalid connector
image

Endpoint Schema Query
image

Resource Schema Query
image

### Comparing `hasConnectors` query size in prod
Old
image

New
image


src/hooks/connectors/useHasConnectors.ts
```diff
@@ -0,0 +1,34 @@
...
+
+// TODO (connectors store) - this is temporary
+// We used to check if connectors exist with a query that returned a bunch
+//      of data and sorting. However, we then go and fetch that again a second
+//      later. So this query is super small to reduce the amount of data and just
+//      making sure there are connectors.
+// We should just make a store that fetches all the needed connectors once
+//      then always pull from that. The store could also cache the schemas, etc.
+//      when the user selects a connector.
+function useHasConnectors(protocol: string | null) {
```
kiahna-tucker, line 15: `useConnectorsExist` and `requiredConnectorColumnsExist` read more easily IMO.

kiahna-tucker (approved):

src/hooks/connectors/useHasConnectors.ts
travjenkins: I love `requiredConnectorColumnsExist`! I could not think of how to state this before and keep the work `Connector` in there haha... I'll update.

PR #1017, opened 2024-03-07
Get endpoint config to read connector tag id from details

travjenkins:
## Issues

issue called out in slack

### misc

- get endpoint config loading similar to resource config
- there is a mild flash when changing connector the first time - but this happens in prod as well. however, removing the excess details setting to help reduce it.
- temporary until I build a new connectors store to manage this stuff

### Manually tested

### Automated tests

-   unit testing covered

## Screenshots

n/a

### DIFF
--- a/src/api/hydration.ts
+++ b/src/api/hydration.ts
@@ -24,15 +24,16 @@ type ConnectorTagEndpointData = Pick<
     'connector_id' | 'endpoint_spec_schema'
 >;
 
-export const getSchema_Endpoint = async (connectorId: string | null) => {
+export const getSchema_Endpoint = async (connectorTagId: string | null) => {
     const endpointSchema = await supabaseRetry(
         () =>
             supabaseClient
                 .from(TABLES.CONNECTOR_TAGS)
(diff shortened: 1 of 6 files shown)


src/api/hydration.ts
```diff
@@ -24,15 +24,16 @@ type ConnectorTagEndpointData = Pick<
...
             supabaseClient
                 .from(TABLES.CONNECTOR_TAGS)
-                .select(`connector_id,endpoint_spec_schema`)
-                .eq('connector_id', connectorId),
+                .select(`endpoint_spec_schema`)
+                .eq('id', connectorTagId)
+                .single(),
         'getSchema_Endpoint'
-    ).then(handleSuccess<ConnectorTagEndpointData[]>, handleFailure);
+    ).then(handleSuccess<ConnectorTagEndpointData>, handleFailure);
```
travjenkins, line 36: Making this work just like resource schema does where we fetch based on the tag id and get a single row back.

kiahna-tucker (approved):

Issue #1023, opened 2024-03-11
Derivations not displaying Ops Logs


travjenkins:
**Bug**
User views the details page for derivation
User hits the logs tab
The logs display an error

**Expected**
The logs display as usual

**Screenshots**
![image](https://github.com/estuary/ui/assets/270078/b28e5253-20b1-4097-90c1-1f007cf6538f)

**Desktop (please complete the following information):**
N/A

**Additional context**
N/A

PR #1024, opened 2024-03-11
Pass correct `kind` for derivations

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1023

### 1023

- Pass the `derivation` `kind` for the ops logs

### Misc

- Do some refactoring to share this code within the details

### Manually tested

- Hitting logs and details page

### Automated tests

- N/A


src/components/shared/Entity/Details/Overview/index.tsx
```diff
@@ -29,21 +28,13 @@ function Overview({ name }: Props) {
...
     );
 
-    const taskTypes: ShardEntityTypes[] = useMemo(() => {
-        if (!isCollection || isDerivation) {
-            return isDerivation ? ['derivation'] : [entityType];
-        }
-
-        return [];
-    }, [entityType, isCollection, isDerivation]);
+    const taskTypes = useDetailsEntityTaskTypes();
```
travjenkins, line 37: Switch over and use the new hook that pulled this login into itself.

src/components/shared/Entity/Details/Ops/index.tsx
```diff
@@ -11,10 +11,15 @@ import { FormattedMessage } from 'react-intl';
...
+import useDetailsEntityTaskTypes from '../useDetailsEntityTaskTypes';
 
 function Ops() {
+    const taskTypes = useDetailsEntityTaskTypes();
     const catalogName = useGlobalSearchParams(GlobalSearchParams.CATALOG_NAME);
-    const [name, collectionName] = useJournalNameForLogs(catalogName);
+    const [name, collectionName] = useJournalNameForLogs(
+        catalogName,
+        taskTypes
+    );
```
travjenkins, line 22: Passing in `taskTypes` so that the journal name hook could be used outside of details in the future.

src/components/shared/Entity/Details/useDetailsEntityTaskTypes.ts
travjenkins, line 1: New hook that pulled in logic from the `Overview` page

src/hooks/journals/shared.ts
```diff
@@ -55,7 +55,7 @@ export async function loadDocuments({
     maxBytes,
     offsets,
 }: LoadDocumentsProps): Promise<LoadDocumentsResponse> {
-    if (!client || !journalName) {
+    if (!client || !journalName || journalName.length === 0) {
```
travjenkins, line 58: We now will return an empty string if we cannot figure out the `kind` of a journal. So making sure there is _some_ string here.

src/hooks/journals/useJournalNameForLogs.ts
```diff
@@ -1,21 +1,25 @@
...
-
         return [
-            `${collectionName}/kind=${entityType}/name=${encodeURIComponent(
-                entityName
-            )}/pivot=00`,
+            !hasLength(entityType)
+                ? ''
+                : `${collectionName}/kind=${
+                      entityType[0] ?? DEFAULT_FILTER
+                  }/name=${encodeURIComponent(entityName)}/pivot=00`,
```
travjenkins, line 22: This is a bit overkill - but wanting to make it safe. Collections do not ever call this but wanted to make it safe.

travjenkins: The defaulting to `DEFAULT_FILTER` is pretty overkill but wanted to make sure this was as safe as possible and make it easier to find calls that are wrong.

src/components/shared/Entity/Details/useDetailsEntityTaskTypes.ts
```diff
@@ -0,0 +1,22 @@
+import { useEntityType } from 'context/EntityContext';
+import { useMemo } from 'react';
+import { ShardEntityTypes } from 'stores/ShardDetail/types';
+import useIsCollectionDerivation from './useIsCollectionDerivation';
+
+function useDetailsEntityTaskTypes() {
```
kiahna-tucker, line 6: Should this hook be named `useEntityTaskTypes` to lessen the likelihood renaming in the future if it is used outside the context of the _Details_ page?

kiahna-tucker (approved):
Approved with a suggestion and without manual testing.

src/components/shared/Entity/Details/useDetailsEntityTaskTypes.ts
travjenkins: I do not think so. I added the `Details` into it to make sure it would not end up outside of details as "easily". If we do move this out I kinda want it to be a "bigger change" that includes renaming the hook.

PR #1032, opened 2024-03-13
point marketplace verify config to cloud run

mdibaiee:
Pointing marketplace verify to the correct endpoint in Cloud Run


travjenkins (approved):
Lgtm

Issue #1043, opened 2024-03-20
Characters entered in the binding filter are forcibly appended to the current input value


kiahna-tucker:
### Reproduction Steps

1. **Paste text into the textbox used to filter bindings.** This input is found in the left-hand pane in the _Target Collections_ (for captures) or _Source Collections_ (for materializations) section. Originally, the problem was observed when a lengthy string was pasted into the input however, upon further investigation, a string of any length will suffice.

1. **Enter characters at the beginning of the pasted text.**

1. **Observe that the first or second character entered may appear at the beginning of the pasted text but additional characters are appended.**

## Expected

Characters in the binding filter should be entered consecutively, starting at the position designated by the cursor.

## Screenshots
Include a screenshot of what you saw when the bug happened.

## Desktop
 - **OS:** Windows
 - **Browser:** Chrome (Version 122.0.6261.129)

PR #1044, opened 2024-03-20
Make the binding filter an uncontrolled component

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1043

### 1043

-   Make the binding filter an uncontrolled component.

### Manually tested

-   Validate that characters entered before pasted text in the binding filter maintains relative positioning and the list of bindings is filtered accordingly.

-   Validate that characters entered after pasted text in the binding filter maintains relative positioning and the list of bindings is filtered accordingly.

### Automated tests

N/A

## Screenshots

N/A


travjenkins (approved):
lgtm

Issue #1047, opened 2024-03-22
Field selection doesn't work after selecting a previously unselected binding


dyaffe:
**Bug**
Field selection hangs indefinitely if you select a binding that was previously de-selected.

Steps:
1. Edit a materilaization that has at least one binding de-selected
2. Select that binding
3. Note that field selection never comes back, even for the bindings that were previously selected.

PR #1048, opened 2024-03-22
prod: mirror specification binding order in client state

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1047

### 1047

* Order binding-related state objects so that the binding order aligns with that of the drafted specification. Previously, the binding-related state objects were being sorted on initialization which caused considerable downstream issues due to the changes in `useServerUpdatedRequiredMonitor` and ripple effects of `useFieldSelectionRefresh`.

* Create a hook to sort and map binding-related information for display in the binding editor.

### Manually tested

-  Validate that the _Next_ CTA appears when the `disable` flag for a given binding has changed in value. The edit workflow for tasks with previously disabled bindings were an area of focus.

* Validate that the _Next_ CTA appears when the resource configuration for the selected binding has changed in value. The edit workflow for tasks with previously disabled bindings were an area of focus.

* Validate that the field selection table actions can be interacted with when the latest built specification contains the selected binding. This includes changing the selected binding in a number of patterns and validating table action interactions.

### Automated tests

-  N/A

## Screenshots

### DIFF
--- a/src/components/collection/Selector/List/index.tsx
+++ b/src/components/collection/Selector/List/index.tsx
@@ -18,7 +18,7 @@ import { useIntl } from 'react-intl';
 import { useUnmount } from 'react-use';
 import {
     useBinding_currentBindingUUID,
-    useBinding_sortedMappedBindings,
+    useBinding_resourceConfigs,
 } from 'stores/Binding/hooks';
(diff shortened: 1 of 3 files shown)


psFried (approved):
LGTM

Issue #1063, opened 2024-04-01
Clicking "x" in the binding editor after searching doesn't clear the form


dyaffe:
**Bug**
Clicking the "x" in the binding editor after searching doesn't clear the form.  

It solely causes the UI to display all entities, but doesn't actually clear the form.  This means that a user still has to remove the thing they previously searched for and search again.

**Expected**
It used to clear the form until the last few days.

PR #1064, opened 2024-04-01
Allow binding search to be clearable and pastable 

travjenkins:
## Issues

Fixes : https://github.com/estuary/ui/issues/1063

### 1063

- Store a local copy so we can still manage the value manually 
    - Fixes the issue with clearing
    - Still allows users to type where ever they want

### Manually tested

- Messed around with the input

### Automated tests

- N/A

## Screenshots

Filtering
image

Filtering and typing at the start of string
image

Filtering and typing at the middle of stirng
image

Filtering and typing at end
image

Clearing out the filter
image
image


kiahna-tucker (approved):

Issue #865, opened 2023-12-12
A flicker of the read-only, Monaco editor can appear on the Spec tab of the collection details page


kiahna-tucker:
## Bug

A flicker of the read-only, Monaco editor can appear on the _Spec_ tab of the collection _Details_ page. This does not happen for all collections.

### Recreation Steps

1. Navigate to the _Details_ page of the following collection: `demo/wikipedia/recentchange`. For higher-speed networks, throttling may be needed to observe the defect.

1. Click on the _Spec_ tab.

1. Observe that a read-only, Monaco editor temporarily appears in the view before it is replaced with a section containing a list of the collection's keys and another containing a table representation of the collection schema.

## Expected

The _Spec_ tab of the collection _Details_ page should not display a read-only, Monaco editor.

## Screenshots

**_Details_ page | _Spec_ tab | Initial load of the tab contents**

<img width="631" alt="ui_issue-865-collection_details_monaco_editor_flicker-editor_appearence" src="https://github.com/estuary/ui/assets/77648584/3fb12f53-4228-44b0-b68c-33ce6029df34">

<br />
<br />

**_Details_ page | _Spec_ tab | Post-editor flicker**

<img width="630" alt="ui_issue-865-collection_details_monaco_editor_flicker-editor_disappeared" src="https://github.com/estuary/ui/assets/77648584/7802322d-dd37-4613-b8ab-2f74c9135f55">

## Desktop

 - **OS:** Windows
 - **Browser:** Chrome (Version 119.0.6045.200)

## Additional Context

N/A

PR #1067, opened 2024-04-02
Enable display of a collection specification on the details page

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1069
https://github.com/estuary/ui/issues/865

### 865

* Add a toggle button to the _Spec_ tab of a collection _Details_ page that allows the user to view a tabular or raw code representation of the specification.

* Add a loading state to the content of the _Spec_ tab of a collection _Details_ page.

### Manually tested

* Validate that the toggle button appears and functions as expected on the _Spec_ tab of a collection _Details_ page.

* Validate that the _Spec_ tab of a non-collection _Details_ page remains unchanged.

* Validate that the loading states of nearly all tables that use the `EntityTableBody` component remain unchanged. The ops logs table was **not** reviewed.

### Automated tests

N/A

## Screenshots

**_Specification_ tab of collection _Details_ page | Table view**

pr_screenshot-1067-collection_spec_presentation-light-table

<br />
<br />

**_Specification_ tab of collection _Details_ page | Table view | Loading**

pr_screenshot-1067-collection_spec_presentation-light-table-loading

<br />
<br />

**_Specification_ tab of collection _Details_ page | Editor view**

pr_screenshot-1067-collection_spec_presentation-light-code

<br />
<br />

**_Specification_ tab of collection _Details_ page | Editor view | Loading**

pr_screenshot-1067-collection_spec_presentation-light-code-loading


src/lang/en-US.ts
```diff
@@ -1156,6 +1156,9 @@ const Details: ResolvedIntlConfig['messages'] = {
     'details.settings.notifications.dataProcessing.noDataProcessedInInterval.intervalOptions.hour': `{interval} {interval, plural, one {Hour} other {Hours}}`,
     'details.settings.notifications.dataProcessing.noDataProcessedInInterval.intervalOptions.day': `{interval} {interval, plural, one {Day} other {Days}}`,
     'details.settings.notifications.dataProcessing.noDataProcessedInInterval.unsetOption': `None`,
+
+    'details.spec.cta.formatted': `Table`,
+    'details.spec.cta.raw': `Code`,
```
kiahna-tucker, line 1161: I contemplated appending both toggle button labels with the word _View_ (i.e., _Table View_ and _Code View_) but that felt a little redundant. Open to any suggestions here.

travjenkins: I think the keys are better names honestly. `Formatted` and `Raw` make more sense... feel much more strong about `Code` -> `Raw`.

src/components/shared/Entity/Details/Spec/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+                                <OutlinedToggleButton
+                                    size="small"
+                                    value="code"
+                                    selected={presentation === 'code'}
+                                    onClick={(_event, value) =>
+                                        evaluatePresentation(value, 'table')
+                                    }
+                                >
+                                    <FormattedMessage id="details.spec.cta.raw" />
+                                </OutlinedToggleButton>
```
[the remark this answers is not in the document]
kiahna-tucker: What would you expect to happen here then? Given there are two presentations to choose from, the only other option I could see would be to make selecting a selected toggle button not do anything -- which doesn't feel great.

Originally, I had these options presented in a select menu which feels like a better fit for the function given these cannot be icon toggle buttons.

src/lang/en-US.ts
kiahna-tucker: Those were the labels I started with but I was not sure if _raw_ was too strong of a word.

src/components/shared/Entity/Details/Spec/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const docsLink = useMemo(() => {
+        switch (entityType) {
+            case 'capture':
+                return 'https://docs.estuary.dev/concepts/captures/#specification';
+            case 'materialization':
+                return 'https://docs.estuary.dev/concepts/materialization/#specification';
+            default:
+                return 'https://docs.estuary.dev/concepts/collections/#specification';
+        }
+    }, [entityType]);
```
travjenkins, line 37: One day we really should move some more stuff into a context for all the little differences between entity types

travjenkins (changes requested):
Only blocking thing is the `Header` being in a file.

Other stuff does not totally have to change. I would review the content with Dave to see his thoughts.

src/components/shared/Entity/Details/Spec/index.tsx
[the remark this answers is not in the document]
travjenkins: I would say doing nothing is better as it is currently designed. If you want it to be a toggle then I think the interaction should probably be on the entire group. As it stands when I interact with it I feel that I am interacting with only one of the buttons and thus the disconnect.

I don't know if we could easily make the interaction feel like it is over the entire group though.

I think a select is not as good as it is just more steps and I doubt (hope) that we never have more versions of the view to add.

src/components/shared/Entity/Details/Spec/index.tsx
kiahna-tucker: Agreed, a similar thought has crossed my mind.

src/lang/en-US.ts
kiahna-tucker: Dave signed off on _table_ and _code_ for the time being.

> our docs do not really have a solid way of expressing this. The only other word I see a lot before a block of spec is `file` but that also _kinda_ feels weird.

I noticed this too and believe it is worth keeping an eye on how our terminology evolves.

Issue #1020, opened 2024-03-08
Break out "Data Read" and "Data Written" in UI Collections View


dyaffe:
**Please propose features in the Discussions area as an `idea`**
We very regularly get asked whether data flows are up to date -- when we do, we usually pull how much data is read and data is written for a particular collection and compare them.  It would be awesome to break those out in the UI.

cc @travjenkins

travjenkins:
So the columns will use these stats:
- `bytes_written_to_me`
- `docs_written_to_me`
- `bytes_read_from_me`
- `docs_read_from_me`

I was thinking "left to right" works so written on the left and read on the right.

So something like this?
![image](https://github.com/estuary/ui/assets/270078/596d6596-121d-4962-9465-ab38ce980389)

PR #1073, opened 2024-04-03
Breaking up the collection columns into two chunks for read and write

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1020

### 1020

- Split up the read and write columns for collections
- Remove the label from the filter button as it was adding too much space
- Moving `docs` up into the column header

### Manually tested

- Viewed all the tables

### Automated tests

- n/a

## Screenshots

image

image

image

image


src/components/filters/Date/index.tsx
```diff
@@ -94,24 +103,18 @@ function DateFilter({ disabled, header, selectableTableStoreName }: Props) {
...
+                        <FormattedMessage id="filter.time.label" />
+                    </Typography>
+                </ListItem>
+                <Divider />
+                <DateFilterOption {...optionProps} option="today" />
+                <DateFilterOption {...optionProps} option="yesterday" />
+                <DateFilterOption {...optionProps} option="thisWeek" />
+                <DateFilterOption {...optionProps} option="lastWeek" />
+                <DateFilterOption {...optionProps} option="thisMonth" />
+                <DateFilterOption {...optionProps} option="lastMonth" />
```
travjenkins, line 117: Made this reusable to prevent copy/paste issues

src/components/filters/Date/index.tsx
```diff
@@ -49,40 +55,43 @@ function DateFilter({ disabled, header, selectableTableStoreName }: Props) {
         },
         setFilter: (option: StatsFilter) => {
             setStatsFilter(option);
-            setCurrentOption(option);
             handlers.closeMenu();
         },
     };
 
+    const optionProps = {
+        currentOption: statsFilter,
```
travjenkins, line 63: Eventually we really need to move the store name into a context so we don't have to pass stuff from a store.

src/components/filters/Date/index.tsx
```diff
@@ -94,24 +103,18 @@ function DateFilter({ disabled, header, selectableTableStoreName }: Props) {
...
-                    <FormattedMessage id="filter.time.thisMonth" />
-                </MenuItem>
-                <MenuItem onClick={() => handlers.setFilter('lastMonth')}>
-                    <FormattedMessage id="filter.time.lastMonth" />
-                </MenuItem>
+                <ListItem>
+                    <Typography>
+                        <FormattedMessage id="filter.time.label" />
+                    </Typography>
+                </ListItem>
```
kiahna-tucker, line 110: Ah, this replaced the tooltip in the intermediate set of screenshots. While the divider helps separate this component (functioning as a header) from the actionable contents of the menu, did you consider increasing the font weight?

Title case is my go-to for menu headers.

src/lang/en-US.ts
```diff
@@ -91,6 +91,7 @@ const CommonMessages: ResolvedIntlConfig['messages'] = {
     'directives.grant.skipped.message': `The access grant link you submitted was not applied as you already have the requested access.`,
 
     // User in filters for tables
+    'filter.time.label': `Show stats for`,
```
kiahna-tucker, line 94: If this content has not been approved yet, I am curious if a concise, informative header would be a better fit (e.g., `Stats Range`).

src/lang/en-US.ts
kiahna-tucker, line 1: `'entityTable.stats.filterMenu'` is unused and can be removed.

kiahna-tucker (approved):
Approved with suggestions and moderate testing.

src/components/filters/Date/index.tsx
travjenkins: I can try that. I didn't like the styling but didn't really try any tweaks. I'll up the weight. You thinking like just making it bold?

src/components/filters/Date/index.tsx
travjenkins: yeah - eventually this'll all live in an array of some sort. Mainly after we move more stuff into store/context

The extra component is kind of a feature and not a bug. The goal will be to move more stuff into stores/context and then all the options don't need to re-render when something changes and are only involved with their own "state".

src/components/tables/Collections/useCollectionColumns.tsx
```diff
@@ -4,7 +4,7 @@ import { hasLength } from 'utils/misc-utils';
 import StatsHeader from '../cells/stats/Header';
 import { ColumnProps } from '../EntityTable/types';
 
-const defaultColumns: ColumnProps[] = [
+const defaultColumnsStart: ColumnProps[] = [
```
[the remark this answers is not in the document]
travjenkins: Yeah - it is used for grouping. I can throw in the underscore - that's an already established pattern.

src/lang/en-US.ts
travjenkins: That sounds good to me.

src/components/filters/Date/index.tsx
kiahna-tucker: Yes. As far as precedents are concerned, the field selection table settings menu is the only other menu that has a header in the application so I would prefer the general look and feel to be _similar_.

travjenkins: I made it `500` 

image

Issue #999, opened 2024-02-26
Somehow users can get stuck waiting for directives even though they loaded


travjenkins:
**Bug**
No clue how this happens but randomly a user might be left hanging on the initial loading of the application waiting for `applied_directives` forever... even though they have already loaded.

**Expected**
The page loads just fine. Maybe we start adding a max time to the initial loading and auto reload the page.

**Screenshots**
Imagine the spinner page

**Desktop (please complete the following information):**
Have seen it once on Edge... maybe browser related?

**Additional context**
N/A

travjenkins:
I got this while working locally. I sadly have no clue. it did happen after the "Entities" error that happens during hot reload. it might be related to that.

![image](https://github.com/estuary/ui/assets/270078/472be7c7-6345-4201-9621-4f8ff760a987)
...
![image](https://github.com/estuary/ui/assets/270078/492e3e74-6489-4903-aa4a-2109c283f823)

travjenkins:
Got this and noticed the "loading slow" comes up. We can detect when that happens for the URL and cause a page reload as a quick fix for the issue.

![Image](https://github.com/estuary/ui/assets/270078/33613c0d-4b79-4c71-8a6c-94ae8e4d2395)

travjenkins:
Saw this locally - no clue if related but looking into it 

![image](https://github.com/estuary/ui/assets/270078/c1a65ed4-8b25-4275-8a2f-88b77abda661)

PR #1076, opened 2024-04-05
Trying to help stuck directives

travjenkins:
## Issues

https://github.com/estuary/ui/issues/999

### 999

- Changing some logging into actual events to make them searchable
- Updating how we manage the state of the directive

### Manually tested

- Really just made sure everything still worked as I am still not sure what exactly is causing the prod issue

### Automated tests

- n/a

## Screenshots

- n/a


src/app/guards/hooks.ts
```diff
@@ -17,15 +17,18 @@ const useDirectiveGuard = (
 ) => {
     logRocketConsole('useDirectiveGuard', selectedDirective);
 
+    const { appliedDirective, isValidating, mutate, error } =
+        useAppliedDirectives(selectedDirective, options?.token);
```
travjenkins, line 21: Want to make this call ASAP.

src/hooks/useAppliedDirectives.ts
```diff
@@ -1,20 +1,24 @@
...
+        () => ({
+            appliedDirective: data
+                ? (data.data[0] as AppliedDirective<any> | undefined)
+                : null,
+            error,
+            mutate,
+            isValidating,
+        }),
+        [data, error, isValidating, mutate]
+    );
```
travjenkins, line 21: This hook was causing some extra renders.

src/app/guards/hooks.ts
```diff
@@ -55,13 +58,16 @@ const useDirectiveGuard = (
...
-    const [freshDirective, setFreshDirective] =
-        useState<AppliedDirective<UserClaims> | null>(null);
+    // The memo above was getting called more often than planned and I think it might have
+    //  been part of https://github.com/estuary/ui/issues/999. So the thinking here is that
+    //  we only really care to update if the value actually changed and that might prevent
+    //  extra hook calls that end up getting "confused"
+    useEffect(() => {
+        logRocketEvent(CustomEvents.DIRECTIVE_GUARD_STATE, {
+            state: directiveState,
+        });
```
travjenkins, line 68: Making this logging an event now so we can easily find sessions where they get a `null` state.

src/types/index.ts
```diff
@@ -155,7 +155,7 @@ export interface StorageMappings {
 export type TenantPaymentProviders = 'external' | 'stripe';
 
 export interface Tenants {
-    collections_quota: number;
+    // collections_quota: number;
```
kiahna-tucker, line 158: Any reason not to remove this now?

kiahna-tucker (approved):
Approved without manual testing/

src/types/index.ts
travjenkins: Since it is in the "root type` I thought it would be best to just comment out so people know it is there.

Issue #770, opened 2023-09-22
Select collections to transform directly from the table on the collections page


dyaffe:
**Bug**
If you select collections and then select the "New Transformation +" button in the UI, your collections won't make their way to the next screen.

dyaffe:
Makes sense that it's a feature request @kiahna-tucker -- in that case, I would suggest that for consistency all grids should work this way.  Multi-selecting checkboxes and then clicking an action button above the grids should start with those entities selected.

kiahna-tucker:
For clarity, all of the entity tables _do_ work this way. The toolbar directly above each table contains available interactions a user can have with its contents; the secondary (i.e., outlined) buttons identify actions you can take on selected rows and the search bar allows you to filter the contents of the table.

The collections table is an interesting case because the derivation create workflow (entered via the primary, _New Transformation_ button) can make use of its data. Nevertheless, there is value to having a primary (i.e., solid, color-filled) button on each page that allows you to enter the relevant creation workflow completely anew that is separate from the table toolbar. I want to state this distinction for the record.

But yes, we can add a _Derive_ or _Transform_ button to the collections table toolbar on the _Collections_ page.

PR #1077, opened 2024-04-05
Transform selected rows of the collections table

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/770

### 770

* Add a _Transform_ CTA to the collections table on the _Collections_ page which allows one to transform selected rows. This is accomplished in a similar fashion as the _Materialize_ CTA.

* Create a shared component for `DataGrid` rows that are in a loading state. `ListItemSkeleton`, the component used to render rows of the data preview list view component in a loading state, was used as a base.

* Extend the `hydrateState` binding store action to prefill state in the derivation create workflow.

* Hydrate the binding store using the existing hydrator component in the derivation create workflow.

* Reset the select table store `hydrated` flag when the filter, sort, or pagination-related properties change to prompt the `EntityTable` loading state.

### Manually tested

* scenarios you manually tested

### Automated tests

N/A

## Screenshots

**NOTE:** Borders in the screenshots below can appear to be too light or missing in the compressed images. Please open the image to see an accurate depiction of the changes in this PR.

**Selected collections table row on the _Collections_ page**

pr_screenshot-1077-transform_collections-table_rows_selected

<br />
<br />

**Derivation create dialog | Collection selector initializing**

pr_screenshot-1077-transform_collections-binding_selector_loading

<br />
<br />

**Derivation create dialog | Collection selector initialized**

pr_screenshot-1077-transform_collections-binding_selector_initialized

<br />
<br />

**Collection _Details_ page | Data preview loading**

pr_screenshot-1077-transform_collections-data_preview_loading


src/components/tables/RowActions/RowSelector.tsx
kiahna-tucker, line 1: I assumed we would like to keep the _Materialize_ and _Transform_ CTAs separate. Nevertheless, a case _could_ be made for displaying them in a button group of their own, akin to the _Disable_, _Enable_, and _Delete_ CTAs. Any strong feelings one way or the other?

src/hooks/useLiveSpecsExt.ts
```diff
@@ -13,7 +13,7 @@ export interface LiveSpecsExtQuery {
     last_pub_id: string;
 }
 
-export interface LiveSpecsExt_MaterializeCapture {
+export interface LiveSpecsExt_MaterializeOrTransform {
```
travjenkins, line 16: I wonder if we should start moving some of the naming conventions away from the entities to more descriptive adjectives. I just have no clue what adjective would make it clear you can `This live spec can be used in something else from the collections table`.

Also, this could easily end up bad if we make a naming convention that does not match other teams's vernacular.

src/components/tables/RowActions/Transform.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+function Transform({ selectableTableStoreName }: Props) {
+    const navigate = useNavigate();
+
+    const selectedRows = useZustandStore<
+        SelectableTableStore,
+        SelectableTableStore['selected']
+    >(selectableTableStoreName, selectableTableStoreSelectors.selected.get);
+
+    const hasSelections = selectedRows.size > 0;
```
travjenkins, line 26: I will leave it up to you

We have duplicated this so many times - should we make a hook? I am like split on this cause the hook will be **tiny** and possibly not worth the effort.

travjenkins: Maybe the better approach is to make a hook for the `RowActions` to all share in general so we have a "headless component" kind of thing.

src/components/tables/RowActions/RowSelector.tsx
```diff
@@ -75,6 +76,12 @@ function RowSelector({
                     selectableTableStoreName={selectableTableStoreName}
                 />
             ) : null}
+
+            {selectableTableStoreName === SelectTableStoreNames.COLLECTION ? (
```
travjenkins, line 80: Can you mark with a `todo` as this can be made a property so all settings are controlled directly by the parent.

I don't think we need to change now and can just handle it later.... mainly because I really want us to move more of the drilled prop settings we have into a context provider for the tables.

src/components/derivation/Create/index.tsx
```diff
@@ -31,6 +52,10 @@ function DerivationCreate() {
         setNewCollectionKey((k) => k + 1);
         resetTransformationCreateState();
         resetBindingState(undefined, true);
+
+        if (collectionsPrefilled) {
+            resetTableSelections();
+        }
```
kiahna-tucker, line 58: Unsure if preserving the selected collection table rows when the derivation create dialog is closes after the _New Transformation_ CTA initiated the workflow is a feature -- as it very well _could_ be -- so I wanted to keep that behavior intact.

src/hooks/useLiveSpecsExt.ts
kiahna-tucker: For the record, I am not a fan of the iterations of this interface name. What is here is the most logical adaptation of the existing name because I could not think of a better refactoring given the narrow scope.

src/components/tables/RowActions/Transform.tsx
kiahna-tucker: Funnily enough, it started as a hook which was fairly small. I meant to leave a comment in this PR calling out the duplication and getting a temperature check for how broadly you were comfortable sharing code at this time.

src/components/derivation/Create/index.tsx
kiahna-tucker: That is what I was thinking too. It is a stretch but it is possible.

travjenkins (approved):
lgtm - made some small mentions in our chat. Approving so you are not blocked when you're ready.

travjenkins (approved):
lgtm - did some quick testing with network failures and they looked alright.

PR #1078, opened 2024-04-08
Force the new edge function url setting

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1079

### 1079

- patched in setting the URL the way it does in the V2 library

### Manually tested

- ran through create while local was pointed to production

### Automated tests

-N/A

## Screenshots

image

### DIFF
--- a/patches/@supabase+supabase-js+1.35.7.patch
+++ b/patches/@supabase+supabase-js+1.35.7.patch
@@ -1,8 +1,36 @@
 diff --git a/node_modules/@supabase/supabase-js/dist/main/SupabaseClient.js b/node_modules/@supabase/supabase-js/dist/main/SupabaseClient.js
-index 8d78eae..5025637 100644
+index 8d78eae..f189e3c 100644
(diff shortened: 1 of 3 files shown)


patches/@supabase+supabase-js+1.35.7.patch
```diff
@@ -0,0 +1,28 @@
...
+         this.storageUrl = `${_supabaseUrl}/storage/v1`;
+-        const isPlatform = _supabaseUrl.match(/(supabase\.co)|(supabase\.in)/);
+-        if (isPlatform) {
+-            const urlParts = _supabaseUrl.split('.');
+-            this.functionsUrl = `${urlParts[0]}.functions.${urlParts[1]}.${urlParts[2]}`;
+-        }
+-        else {
+-            this.functionsUrl = `${_supabaseUrl}/functions/v1`;
+-        }
++        this.functionsUrl = `${_supabaseUrl}/functions/v1`
```
travjenkins, line 17: We need to use this URl as it seems to work better for users. However, this path is only set this way in V2 of the library.

kiahna-tucker (approved):

Issue #1075, opened 2024-04-05
Split out the Collections stats graph for `read` and `written`


travjenkins:
When we start splitting the stats columns for Collections we might need to also update the graph in details. There is a simple approach: we could just have two graphs - one for `read` and one for `written`. However, we might want to put some effort into designing it so `read` and `written` are in the same graph

PR #1080, opened 2024-04-10
Travjenkins/feature/break stats up in collections

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1075

### 1075

- Add a `data`|`docs` toggle to the stats graphs
- Render both `read` and `write` data for collections in graphs

### Manually tested

- Hit the graphs a lot

### Automated tests

- n/a

## Screenshots

Capture
image
image

Collection
image
image

Materialization
image
image


src/components/shared/Entity/Details/Usage/useDetailsUsageStore.ts
kiahna-tucker, line 1: The intention behind preserving the value of `range` and `statType` unless explicitly changed is to make it easy to compare entities to one another, correct?

src/components/filters/HourRange.tsx
```diff
@@ -29,7 +30,9 @@ function HourlyRangeFilter({ range, setRange }: Props) {
 
     return (
         <Stack direction="row" spacing={1} sx={{ alignItems: 'center' }}>
-            <FormattedMessage id="detailsPanel.recentUsage.title.prefix" />
+            <Typography id="hourly-filter-selector__label">
```
kiahna-tucker, line 33: The styling of all card headers should be uniform. This header is missing the following: `{ fontSize: 16, fontWeight: 300 }`.

It should be noted that the heading of the data preview card is similarly divergent.

kiahna-tucker (changes requested):
The review comments identify the items that need to be addressed before this PR can be merged.

### Additional Questions

* Did you intentionally want data/docs written to always use the light color variant on the usage by hour graph, regardless of the entity being viewed? Or should an instance of this graph that does not have overlapping bars use the medium color variant?

src/components/graphs/DataByHourGraph/DataTypePicker.tsx
```diff
@@ -0,0 +1,40 @@
+import { ToggleButtonGroup } from '@mui/material';
+import useDetailsUsageStore from 'components/shared/Entity/Details/Usage/useDetailsUsageStore';
+import OutlinedToggleButton from 'components/shared/OutlinedToggleButton';
+import { outlinedToggleButtonGroupStyling } from 'context/Theme';
+import { FormattedMessage } from 'react-intl';
+
+function StatTypePicker() {
```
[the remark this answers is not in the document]
travjenkins: I'm not concerned about calling it a picker.

src/context/Theme.tsx
```diff
@@ -288,6 +288,12 @@ export const jsonObjectPreview_value = {
 };
 
 // Styles
+// TODO (styling) - we use this styling on non-cards we should probably rename this
+//  and use in other places
+export const cardHeaderSx: SxProps<Theme> = {
+    fontSize: 16,
+    fontWeight: 500,
```
kiahna-tucker, line 295: The font weight should be 300.

kiahna-tucker (changes requested):
See comment below.

src/context/Theme.tsx
travjenkins: I'll update the weight - I am not moving around the component cause that is a major pain and that is why I skipped it when these were originally made. I'll mark it with a `todo`

kiahna-tucker (approved):
Approved without manually testing the changes pushed on April 19th.

Issue #1082, opened 2024-04-16
Google Sheets OAuth sometimes fails but kind of works


travjenkins:
**Bug**
Not totally sure how to always recreate this but pretty sure it is related to the new OAuth design that Google is pushing. 

Sometimes when someone is adding OAuth the pop up can sit awhile on a loading page and that sometimes seems to trigger an issue with the UI thinking it never got the OAuth URL to open.

**Expected**
If you get the `Authenticated` tag the error should go away and the endpoint config should work

**Screenshots**
![image](https://github.com/estuary/ui/assets/270078/f4ed305f-cb97-4e91-b33e-f9affce3f3a9)

![image](https://github.com/estuary/ui/assets/270078/9f5aa803-7713-4a9d-a60f-b4baa3c39df8)

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**

dyaffe:
@travjenkins, is this a UI bug in this case?  I have people running into it in production so wondering if it would be an easy fix.

travjenkins:
The edge function is not returning all the properties inside of the `Credentials` object. Not sure why yet.
![image](https://github.com/estuary/ui/assets/270078/8e3e3aac-c928-4840-9962-8c7e762c132d)

travjenkins:
So what I think is happening is the store has a potential to lose changes due to some old approaches to coding we did. Since this is not always happening (and the issue would not always happen) I am thinking that is the cause.

travjenkins:
Seems to be working no. Still need to find the true root cause.

PR #1084, opened 2024-04-17
Update endpointconfig store to be used directly

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1082

### 1082

- Remove all uses of `get` while inside of a `produce` as I think this is causing changes to get lost
- While we are testing this already - just using hooks for the store directly
- Adding events for LR to make this issue easier to track

### Manually tested

- Capture create/edit with Oauth
- Materialization create/edit with Oauth

### Automated tests

- N/A

## Screenshots

...tbd

### DIFF
--- a/src/stores/EndpointConfig/Store.ts
+++ b/src/stores/EndpointConfig/Store.ts
@@ -132,6 +132,10 @@ const getInitialState = (
                     ? createJSONFormDefaults(state.endpointSchema)
                     : encryptedEndpointConfig;
 
+                // TODO (endpoint config)
+                // I have NO clue why this is passing the encryptedConfig that is becing passed in and
+                //  not what is set to the state. Especially since it could end up with setting defaults to
+                //  the state value
                 populateErrors(
                     encryptedEndpointConfig,
                     state.customErrors,
--- a/src/stores/EndpointConfig/Store.ts
+++ b/src/stores/EndpointConfig/Store.ts
@@ -15,7 +15,6 @@ import {
     getInitialHydrationData,
     getStoreWithHydrationSettings,
 } from 'stores/extensions/Hydration';
-import { EndpointConfigStoreNames } from 'stores/names';
 import { JsonFormsData, Schema } from 'types';
 import {
     configCanBeEmpty,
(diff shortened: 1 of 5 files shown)


src/stores/EndpointConfig/Store.ts
```diff
@@ -87,11 +86,9 @@ const getInitialState = (
                 state.customErrors = val;
                 state.customErrorsExist = hasLength(val);
 
-                const { endpointConfig } = get();
```
travjenkins, line 90: I am hoping that these `get`s are what is causing potential issues. I almost cleaned these up last time I was in here but decided to leave them because they "weren't hurting anything".

src/stores/EndpointConfig/Store.ts
```diff
@@ -135,13 +128,19 @@ const getInitialState = (
...
-                populateErrors(encryptedEndpointConfig, customErrors, state);
+                // TODO (endpoint config)
+                // I have NO clue why this is passing the encryptedConfig that is becing passed in and
+                //  not what is set to the state. Especially since it could end up with setting defaults to
+                //  the state value
+                populateErrors(
+                    encryptedEndpointConfig,
+                    state.customErrors,
+                    state
+                );
```
travjenkins, line 143: @kiahna-tucker - do you have _any_ insight into this? I am totally not sure why this is passing the value it is.

src/stores/EndpointConfig/hooks.ts
```diff
@@ -1,229 +1,98 @@
...
 };
 
 export const useEndpointConfigStore_changed = () => {
-    const entityType = useEntityType();
-
-    return useZustandStore<
-        EndpointConfigState,
-        EndpointConfigState['stateChanged']
-    >(getStoreName(entityType), (state) => state.stateChanged);
+    return useEnpointConfigStore(useShallow((state) => state.stateChanged));
```
travjenkins, line 17: Threw this tweak in while cleaning up. Since this is a computed value I think we're better off using shallow here.

src/stores/EndpointConfig/Store.ts
kiahna-tucker: I cannot recall off the top of my head and would want to look at the commit history for this file.

kiahna-tucker (approved):

src/stores/EndpointConfig/Store.ts
travjenkins: I looked through the history of the file and it has always been this way.

Issue #1088, opened 2024-04-19
Remove `InvariableStores`


travjenkins:
We need to remove `InvariableStores` and use stores directly as hooks.

### DetailsFormStoreNames
- [x] `capture-details-form` : https://github.com/estuary/ui/pull/1123
- [x] `collection-details-form` : https://github.com/estuary/ui/pull/1123
- [x] `materialization-details-form` : https://github.com/estuary/ui/pull/1123

### EditorStoreNames
- [ ] `bindings_editor_store`
- [ ] `capture_editor_store`
- [ ] `derivation_editor_store`
- [ ] `general_editor_store`*
- [ ] `materialization_editor_store`

### FormStateStoreNames
- [ ] `Capture-Create-Form-State`
- [ ] `Capture-Edit-Form-State`
- [ ] `Collection-Create-Form-State`
- [ ] `Materialization-Create-Form-State`
- [ ] `Materialization-Edit-Form-State`

### SelectTableStoreNames
- [ ] `AccessGrants-Links`
- [ ] `AccessGrants-Users`
- [ ] `AccessGrants-Prefixes`
- [ ] `Billing-Table`
- [ ] `Captures-Table`
- [ ] `Collections-Table`
- [ ] `Collections-Selector-Table`
- [ ] `Connectors-Table`
- [ ] `Materializations-Table`
- [ ] `Prefix-Alert-Table`
- [ ] `Refresh-Tokens-Table`
- [ ] `Storage-Mappings-Table`

### ShardDetailStoreNames
- [ ] `Capture-Shard-Detail`
- [ ] `Materialization-Shard-Detail`
- [ ] `Collection-Shard-Detail`

### Misc
- [ ] `queuedTasks` 
- [ ] `Transform-Create`
- [ ] `general_bindings_editor` _K_
- [ ] `Onboarding`
- [ ] `general-schema-evolution`
- [ ] `general-endpoint-config`
- [x] `general-existing-entity` : https://github.com/estuary/ui/pull/1451
- [x] `binding` : https://github.com/estuary/ui/pull/1235
- [x] `source-capture` : https://github.com/estuary/ui/pull/1112
- [x] `general_billing` : https://github.com/estuary/ui/pull/1108
- [x] `Side-Panel-Docs` : https://github.com/estuary/ui/pull/1090
- [x] `Storage-Mappings` : https://github.com/estuary/ui/pull/1109
- [x] `Top-Bar` : https://github.com/estuary/ui/pull/1091
[...]

PR #1089, opened 2024-04-19
Invariable Store Removal : Entities Store

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1088

### 1088

- Moved the 'Entities` store into hook
- Cleaned up old hooks

### Manually tested

- Viewing demo `See Capture` pop up to add `/demo` access
- Added `/demo` access
- Poked around generally

### Automated tests

- n/a

## Screenshots

Read Only User Capture, Transform, Materialization
image
image
image

### DIFF
--- a/src/components/hero/AcceptDemoInvitation.tsx
+++ b/src/components/hero/AcceptDemoInvitation.tsx
@@ -7,7 +7,7 @@ import Error from 'components/shared/Error';
 import { jobStatusQuery, trackEvent } from 'directives/shared';
 import useJobStatusPoller from 'hooks/useJobStatusPoller';
 import { Dispatch, SetStateAction, useState } from 'react';
-import { useEntitiesStore_mutate } from 'stores/Entities/hooks';
+import { useEntitiesStore } from 'stores/Entities/Store';
 
 interface Props {
     tenant: string;
@@ -32,7 +32,7 @@ function AcceptDemoInvitation({
 
     const { jobStatusPoller } = useJobStatusPoller();
 
-    const mutateAuthRoles = useEntitiesStore_mutate();
+    const mutateAuthRoles = useEntitiesStore((state) => state.mutate);
 
     const [serverError, setServerError] = useState<PostgrestError | null>(null);
 
--- a/src/stores/Entities/hooks.ts
+++ b/src/stores/Entities/hooks.ts
@@ -29,23 +29,18 @@ export const useEntitiesStore_capabilities_adminable = () => {
 };
 
 export const useEntitiesStore_capabilities_readable = () => {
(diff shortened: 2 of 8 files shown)


src/components/hero/AcceptDemoInvitation.tsx
```diff
@@ -32,7 +32,7 @@ function AcceptDemoInvitation({
 
     const { jobStatusPoller } = useJobStatusPoller();
 
-    const mutateAuthRoles = useEntitiesStore_mutate();
+    const mutateAuthRoles = useEntitiesStore((state) => state.mutate);
```
travjenkins, line 35: Won't always remove the "mediator hooks" but this store is used in so few spots it seems safe to do it.

src/stores/Entities/hooks.ts
travjenkins, line 1: I think the hooks folder will probably only house things that are computed values or contains some "store logic" that should not be exposed to consumers.

src/stores/Entities/hooks.ts
```diff
@@ -1,131 +1,55 @@
...
+    }));
 };
 
-export const useEntitiesStore_capabilities = (
-    kind: keyof EntitiesState['capabilities']
-) => {
-    return useZustandStore<EntitiesState, Schema>(
-        GlobalStoreNames.ENTITIES,
-        (state) => state.capabilities[kind]
-    );
```
travjenkins, line 24: We should go in this direction (where we pass `admin`,`read`,etc.) to fetch the capabilities we want - but the other ones are used right now so just removing

src/stores/Entities/hooks.ts
```diff
@@ -1,131 +1,72 @@
...
 
-export const useSidePanelDocsStore_resetState = () => {
-    return useZustandStore<EntitiesState, EntitiesState['resetState']>(
-        GlobalStoreNames.ENTITIES,
-        (state) => state.resetState
+export const useEntitiesStore_hasDemoTenantAccess = () => {
+    const readable = useEntitiesStore_capabilities_readable();
+    return useEntitiesStore(
+        useShallow(() => {
+            return Object.keys(readable).includes('demo/');
```
kiahna-tucker, line 47: The `demo/` tenant is referenced in a few places by name. Should it (and the `ops/` tenant) be stored in a constant as well?

kiahna-tucker (approved):

src/stores/Entities/hooks.ts
travjenkins: Yeah - I think we can do that in another PR. I feel like we should probably get some tenant helpers/hooks/whatever setup and store these off in const around that time.

Issue #435, opened 2023-01-04
Admin - Storage Mapping edit 


travjenkins:
**Feature**
Add the ability to add and edit storage mappings via the UI.

**Technical Note**
Look for `// TODO (storage mapping edit)` in the code when implementing this as some files are available.

**Design**
This won't be followed 100% but a general basis of what we want
![image](https://user-images.githubusercontent.com/270078/210590324-364ed466-125e-4e8a-8798-455183a0b889.png)

travjenkins:
Some cross over with https://github.com/estuary/ui/issues/138

PR #1093, opened 2024-04-22
Configure storage mappings

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/435

### 435

* Scope the content of the _Settings_ tab on the _Admin_ page to a specific tenant.

* Create a tenant store used for the tenant selector on the _Settings_ and _Billing_ tabs on the _Admin_ page. The tenant-related state that was found in the billing store was removed.

* Create a shared, tenant selector component for the _Settings_ and _Billing_ tabs on the _Admin_ page.

* Add a CTA to the toolbar of the storage mappings table which enables the user to set up a new storage mapping. The only cloud providers supported at this time are: GCS and S3.

* Add a Log Rocket event to track failures in the republication of a prefix.

### Manually tested

* Validate error handling for the storage mapping directive.

* Validate the success and error handling for the asynchronous tasks involved in republishing a prefix.

### Automated tests

N/A

## Screenshots

**_Settings_ tab**

pr_screenshot-1093-configure_mappings-settings_tab-default

</br>
</br>

**_Configure Storage_ dialog | Opened**

pr_screenshot-1093-configure_mappings-dialog-default

</br>
</br>

**_Configure Storage_ dialog | Provider selector menu**

pr_screenshot-1093-configure_mappings-dialog-provider_selector_opened

</br>
</br>

**_Configure Storage_ dialog | GCS form | Unedited**

pr_screenshot-1093-configure_mappings-dialog-form_gcs-unedited

</br>
</br>

**_Configure Storage_ dialog | S3 form | Unedited**

pr_screenshot-1093-configure_mappings-dialog-form_s3-unedited

</br>
</br>

**_Configure Storage_ dialog | GCS form | Completed**

pr_screenshot-1093-configure_mappings-dialog-form_gcs-completed

</br>
</br>

**_Configure Storage_ dialog | GCS form | Directive errored**

pr_screenshot-1093-configure_mappings-dialog-form_gcs-form_error

</br>
</br>

**_Configure Storage_ dialog | Republishing prefix**

pr_screenshot-1093-configure_mappings-dialog-logs-publishing

</br>
</br>

**_Configure Storage_ dialog | Prefix successfully republished**

pr_screenshot-1093-configure_mappings-dialog-logs-published-success

</br>
</br>

**_Configure Storage_ dialog | Prefix republication failed**

pr_screenshot-1093-configure_mappings-dialog-logs-publishing-failed


src/api/alerts.ts
travjenkins, line 1: No need to change

I wonder if we still need to render the `catalog prefix` in the table now that you can filter them to a single tenant ahead of time.

src/api/publications.ts
```diff
@@ -11,3 +20,25 @@ export const createPublication = (
...
     });
 };
+
+export const getPublicationById = async (pubId: string) => {
+    const data = await supabaseRetry(
+        () =>
+            supabaseClient
+                .from(TABLES.PUBLICATIONS)
+                .select(JOB_STATUS_COLUMNS)
+                .eq('id', pubId),
```
travjenkins, line 30: We are now duplicating this query 3 times. If I remember right it might be a bit of a pain to get all three of these shared. Please look into it though.

src/api/publications.ts
```diff
@@ -11,3 +20,25 @@ export const createPublication = (
...
+        'getPublicationById'
+    ).then(
+        handleSuccess<
+            {
+                job_status: { type: string };
+                logs_token: string;
+                id: string;
+            }[]
```
travjenkins, line 38: Can you toss this into an `interface`... probably in a shared spot near the columns... but not super opinionated on where it should live.

src/api/storageMappings.ts
```diff
@@ -21,14 +25,15 @@ const getStorageMappings = (
             catalog_prefix,
             updated_at
         `
-            // TODO (storage mappins) including count will make pagination work but
+            // TODO (storage mappings) including count will make pagination work but
```
travjenkins, line 28: Did you test if adding the `catalog_prefix` helps at all with this performance? I am assuming not but wanted to check.

src/components/admin/Settings/StorageMappings/Dialog/Form.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+function StorageMappingForm() {
+    const formValue = useStorageMappingStore((state) => state.formValue);
+    const updateFormValue = useStorageMappingStore(
+        (state) => state.updateFormValue
+    );
+
+    const { schema, uischema } = useConfigurationSchema();
+
+    return Boolean(!isEmpty(schema) && !isEmpty(uischema)) ? (
```
travjenkins, line 17: I know we're being safe - but I think we can remove this. Then in the `useConfigurationSchema` could probably work like a context hook where it just throws an error if `undefined` was going to be returned.

src/components/admin/Settings/StorageMappings/Store/types.ts
```diff
@@ -0,0 +1,24 @@
+import { PostgrestError } from '@supabase/postgrest-js';
+import { JsonFormsData } from 'types';
+
+export interface StorageMappingState {
+    provider: string;
```
travjenkins, line 5: I feel like this should probably just be typed based on the mapping enum since we know we do not want to support anything that isn't explicitly coded for.

src/components/admin/Settings/index.tsx
travjenkins, line 1: No change requied

I don't know a better solution... but the tenant selector looks lonely since it does not have content to the left like billing.
image

I think we could probably add something similar for settings explaining that these settings apply to the entire tenant or something.

src/components/admin/Settings/StorageMappings/Dialog/useRepublishPrefix.ts
travjenkins, line 1: For the LogRocket events replace `context`s with something simpler. They should be like `failed`, `missingLogs`, etc. This is not a field intended for "commenting" and more for searching.

src/components/admin/Settings/StorageMappings/Dialog/useRepublishPrefix.ts
```diff
@@ -0,0 +1,113 @@
...
+                async (payload: any) => {
+                    setSaving(false);
+                    setServerError(
+                        intl.formatMessage({
+                            id: 'storageMappings.dialog.generate.error.republicationFailed',
+                        })
+                    );
+
+                    logRocketEvent(CustomEvents.REPUBLISH_PREFIX_FAILED, {
+                        response: payload,
```
travjenkins, line 95: You cannot pass an "open ended" object to eventing. LR will complain and against their terms. This should be like the other ones. If you want the more open ended logging you can then write a log rocket log right after this.

src/components/shared/TenantSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const [inputValue, setInputValue] = useState('');
+    const [defaultValue, setDefaultValue] = useState<string | null>(null);
+
+    const tenantNames = useMemo<string[] | null>(
+        () =>
+            tenants && tenants.length > 0
+                ? tenants.map(({ tenant }) => tenant)
+                : null,
+        [tenants]
+    );
```
travjenkins, line 39: No change needed

I get why we're doing this... but I dislike it for some reason. I wonder if we should move this kinda thing into the context as a stand alone hook.

src/components/shared/TenantSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }, [defaultSelectedTenant, defaultValue, tenantNames, updateState]);
+
+    return selectedTenant && tenantNames ? (
+        <AutocompletedField
+            label={intl.formatMessage({
+                id: 'common.tenant',
+            })}
+            options={tenantNames}
+            defaultValue={defaultValue ?? selectedTenant}
```
travjenkins, line 77: I am pretty sure that this will cause issues. Since we are manipulating `defaultValue` I think we're supposed to use `value` instead. But not 100% sure.

src/components/tables/PrefixAlerts/index.tsx
```diff
@@ -44,14 +45,23 @@ function PrefixAlertTable() {
...
+
     const query = useMemo(() => {
-        return getNotificationSubscriptionsForTable(pagination, searchQuery, [
-            {
-                col: columnToSort,
-                direction: sortDirection,
-            },
-        ]);
-    }, [columnToSort, pagination, searchQuery, sortDirection]);
+        return selectedTenant
```
travjenkins, line 51: Please add a quick comment about why we're returning null since the `null` getting handled is a little "deep" in the store.

src/directives/shared.ts
```diff
@@ -219,6 +219,40 @@ export const DIRECTIVES: Directives = {
...
+                return 'unfulfilled';
+            }
+
+            // If directive already queued and no claim is there we can just use that directive again
+            if (
+                appliedDirective.job_status.type === 'queued' &&
+                !appliedDirective.user_claims
+            ) {
+                return 'in progress';
+            }
```
travjenkins, line 249: Can this actually happen for this directive?

src/lang/en-US.ts
```diff
@@ -600,6 +601,18 @@ const StorageMappings: ResolvedIntlConfig['messages'] = {
...
+    'storageMappings.dialog.generate.description': `Choose where you'd like {tenant} data to be stored. This location will be used for all future write operations.`,
+    'storageMappings.dialog.generate.alert.keyPrefix': `Your tenant, {tenant}, will be used to prefix the keys written to the specified storage location.`,
+    'storageMappings.dialog.generate.providerOption.AZURE': `Azure Object Storage Service`,
+    'storageMappings.dialog.generate.providerOption.CUSTOM': `An S3-compatible Endpoint`,
+    'storageMappings.dialog.generate.providerOption.GCS': `Google Cloud Storage`,
+    'storageMappings.dialog.generate.providerOption.S3': `Amazon Simple Storage Service`,
+    'storageMappings.dialog.generate.logsHeader': `Please wait while we save and apply your storage mapping.`,
+    'storageMappings.dialog.generate.error.republicationFailed': `There was an error republishing the entities in your system. Please try again.`,
+    'storageMappings.dialog.generate.error.unableToFetchLogs': `There was an issue fetching the logs when applying the new storage mapping. Please contact support to confirm that your system has been updated accordingly.`,
```
travjenkins, line 615: These are all very dry but seem okay to me.

src/lang/en-US.ts
```diff
@@ -696,8 +709,8 @@ const Ops: ResolvedIntlConfig['messages'] = {
     'ops.logsTable.tailNewLogs': `stay at bottom as new logs load`,
 
     // These keys get generated inside the WaitingForRowBase
-    'ops.logsTable.waitingForLogs.old.failed': `A network error occured. Please reload.`,
-    'ops.logsTable.waitingForLogs.new.failed': `A network error occured. Please reload.`,
+    'ops.logsTable.waitingForLogs.old.failed': `A network error occurred. Please reload.`,
+    'ops.logsTable.waitingForLogs.new.failed': `A network error occurred. Please reload.`,
```
travjenkins, line 713: :laughing:

src/stores/Tenant/Store.ts
```diff
@@ -0,0 +1,30 @@
...
+    selectedTenant: '',
+});
+
+const getInitialState = (
+    set: NamedSet<TenantState>,
+    _get: StoreApi<TenantState>['getState']
+): TenantState => ({
+    ...getInitialStateData(),
+
+    updateSelectedTenant: (value) => {
```
travjenkins, line 17: We should really just keep this consistent and name it with `set` at the start.

src/lang/en-US.ts
```diff
@@ -600,6 +601,18 @@ const StorageMappings: ResolvedIntlConfig['messages'] = {
     'storageMappings.bucket.description': `The name of the bucket you have setup to store data in.`,
     'storageMappings.lastUpdated.label': `Last Updated`,
     'storageMappings.message': `Below are all the ${CommonMessages['terms.storageMapping']} that you have read or admin access to. These are the locations that your data is stored. You currently cannot edit these in the UI. If you need an update please ${CTAs['cta.support']}.`,
+
+    'storageMappings.cta.addStorageMapping': `Configure Storage`,
+    'storageMappings.dialog.generate.header': `Configure Storage`,
```
travjenkins, line 606: If these are going to be the same content please just use a single key

src/components/shared/TenantSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+                ? tenants.map(({ tenant }) => tenant)
+                : null,
+        [tenants]
+    );
+
+    const updateState = useCallback(
+        (value: string) => {
+            if (updateStoreState) {
+                updateStoreState(value);
+            }
```
travjenkins, line 45: My gut feeling is this should _probably_ happen after the local updates. That way the local is updated and rendered correctly before anything else is updated. Very nit picky and not a big deal.

src/components/admin/Settings/StorageMappings/Dialog/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    const closeDialog = (event: React.MouseEvent<HTMLElement>) => {
+        event.preventDefault();
+
+        setOpen(false);
+        resetState();
+        setSaving(false);
+    };
+
+    return (
```
travjenkins, line 58: Very small improvements - but since this is all store based can you break this dialog up a bit so that chunks that do not need to re-render are not re-rendered.

Probably just title, content , and actions

src/components/admin/Settings/StorageMappings/Dialog/SaveButton.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+            if (response.error) {
+                setSaving(false);
+
+                return setServerError(
+                    (response.error as PostgrestError).message
+                );
+            }
+
+            const data = response.data[0];
```
travjenkins, line 94: Since you handled it super safe elsewhere - we should keep that going.

Let's add a check if the `response.data` isn't there to the `response.error` check up above. I don't think we need to make a custom error message but should just be super safe.

travjenkins (commented):
Overall looks good - most of the callouts are pretty nit picky and strict.

Gonna working on testing it tomorrow.

src/lang/en-US.ts
kiahna-tucker: I was waiting for Dave to review the content before consolidating content strings. It is looking like this PR will be merged before the content is _officially_ approved, so I do not mind doing so at this time.

src/directives/shared.ts
kiahna-tucker: Haha no, the second condition should be removed for this directive at the very least.

src/components/admin/Settings/StorageMappings/Dialog/useConfigurationSchema.ts
```diff
@@ -0,0 +1,94 @@
...
+    properties: {
+        bucket: {
+            description: 'Bucket into which Flow will store data.',
+            type: 'string',
+            pattern: '(^[a-z0-9][a-z0-9\\-_\\.]{1,60}[a-z0-9]$)',
+        },
+        prefix: {
+            description: 'Optional prefix of keys written to the bucket.',
+            type: 'string',
+            pattern: '^([\\p{Letter}\\p{Number}\\-_\\.]+/)*$',
```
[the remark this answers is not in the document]
kiahna-tucker: This is hilarious. I would have caught this on my final pass over the PR but agreed, a regex needs to be interpolated in both locations.

src/api/alerts.ts
kiahna-tucker: That is a decision for Dave. Personally, I would advocate for keeping the column around given there was discussion of allowing notifications for deeper namespaces in the future.

src/components/admin/Settings/StorageMappings/Store/types.ts
kiahna-tucker: Funnily enough, the typing became a cumbersome to implement but I can take another stab at it with fresh eyes.

src/components/shared/TenantSelector.tsx
kiahna-tucker: I would argue that it should be moved into the `useTenantDetails` hook defined inside the context provider because almost every component that uses this hook has a similar, memoized check.

src/components/shared/TenantSelector.tsx
kiahna-tucker: This logic existed in the original tenant selector on the _Billing_ page and, yes, it was definitely incorrect. Only the `value` autocomplete component property should be set since the component is controlled.

src/components/shared/TenantSelector.tsx
travjenkins: Up to you if you wanna mark with a `todo` or include in this PR.

src/components/shared/TenantSelector.tsx
travjenkins: Oh that's fair. Let's leave it then to keep things simpler.

src/components/admin/Settings/index.tsx
kiahna-tucker: Agreed. The same thought crossed my mind as well.

src/components/shared/TenantSelector.tsx
kiahna-tucker: Noting this action item in a `TODO` comment in the tenant context provider.

src/api/storageMappings.ts
kiahna-tucker: No, I have not evaluated the performance impact yet.

src/components/admin/Settings/StorageMappings/Store/types.ts
kiahna-tucker: Ah, I remember the friction in changing the type: the `provider` property would need to be defaulted to a member of the  `CloudProviderCodes` enumeration to use it as the sole type annotation. If you are comfortable with `provider` being nullable as well then there is no issue here.

travjenkins (approved):
changes look good - didn't manually test

Issue #1095, opened 2024-04-25
Entity table headers for stats are wrong


travjenkins:
**Bug**
Entity tables for Capture and Materializations are wrong

**Expected**
Columns do not say 'written' and `read` twice

**Screenshots**
![image](https://github.com/estuary/ui/assets/270078/97818dbc-135f-43c1-a8bb-d2d437d7a262)

PR #1096, opened 2024-04-25
Travjenkins/prod/stats header content

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1095

### 1095

- Make the stats header more reusable
- Remove old content to reduce chance that this happens again

### Manually tested

- Hit all 3 tables

### Automated tests

- n/a

### Captures
image

### Collections
image

### Materializations
image

### DIFF
--- a/src/components/tables/cells/stats/Header.tsx
+++ b/src/components/tables/cells/stats/Header.tsx
@@ -2,7 +2,8 @@ import { TableCell, Typography } from '@mui/material';
 import DateFilter from 'components/filters/Date';
 import { useTenantDetails } from 'context/fetcher/Tenant';
 import { useZustandStore } from 'context/Zustand/provider';
-import { FormattedMessage } from 'react-intl';
+import { useMemo } from 'react';
+import { FormattedMessage, useIntl } from 'react-intl';
 import { SelectTableStoreNames } from 'stores/names';
 import {
     SelectableTableStore,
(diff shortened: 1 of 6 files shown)


src/components/tables/cells/stats/Header.tsx
```diff
@@ -37,32 +46,61 @@ const StatsHeader = ({ header, selectableTableStoreName }: Props) => {
...
+                    id: headerSuffix ? headerSuffix : 'data.read',
+                },
+                {
+                    type: intl.formatMessage({
+                        id: header ?? 'data.docs',
+                    }),
+                }
+            ),
+        ];
+    }, [header, headerSuffix, intl]);
```
travjenkins, line 72: Since we need to pass the second column header in to two different things I just made the first column translate the same way to keep things consistent.

src/components/tables/cells/stats/Header.tsx
```diff
@@ -37,32 +46,61 @@ const StatsHeader = ({ header, selectableTableStoreName }: Props) => {
...
-                    selectableTableStoreName={selectableTableStoreName}
-                />
+                {hideFilter ? (
+                    <Typography
+                        component="span"
+                        sx={{ mt: 0.5, fontWeight: 500, whiteSpace: 'nowrap' }}
+                    >
+                        {secondColHeader}
+                    </Typography>
+                ) : (
```
travjenkins, line 92: Biggest changed so that we can use the stats columns in collections for the part of the columns that do not need the filter

PR #1098, opened 2024-04-29
View and manage refresh tokens

kiahna-tucker:
## Issues

### issue_number

-   changes...

### Manually tested

### Automated tests

-   unit testing covered

## Screenshots

### DIFF
--- a/src/components/admin/Api/index.tsx
+++ b/src/components/admin/Api/index.tsx
@@ -1,8 +1,9 @@
-import { Box, SxProps, Theme, Typography } from '@mui/material';
+import { Box, Grid, SxProps, Theme, Typography } from '@mui/material';
 import { Auth } from '@supabase/ui';
 import { authenticatedRoutes } from 'app/routes';
 import AdminTabs from 'components/admin/Tabs';
 import SingleLineCode from 'components/content/SingleLineCode';
+import RefreshTokenTable from 'components/tables/RefreshTokens';
 import usePageTitle from 'hooks/usePageTitle';
 import { FormattedMessage } from 'react-intl';
 
@@ -24,18 +25,36 @@ function AdminApi() {
         <>
             <AdminTabs />
 
+            <Grid container spacing={{ xs: 3, md: 2 }} sx={{ p: 2 }}>
+                <Grid item xs={12} md={9}>
+                    <Typography variant="h6" sx={{ mb: 0.5 }}>
+                        <FormattedMessage id="admin.cli_api.header" />
+                    </Typography>
+
+                    <Typography>
+                        <FormattedMessage id="admin.cli_api.message" />
+                    </Typography>
+                </Grid>
+            </Grid>
+
             <Box sx={boxStyling}>
                 <Typography variant="h6" sx={{ mb: 0.5 }}>
-                    <FormattedMessage id="admin.accessToken" />
+                    <FormattedMessage id="admin.cli_api.accessToken" />
                 </Typography>
 
                 <Typography sx={{ mb: 2 }}>
-                    <FormattedMessage id="admin.accessToken.message" />
+                    <FormattedMessage id="admin.cli_api.accessToken.message" />
                 </Typography>
 
                 {/* TODO (defect): Display an error in the event the access token does not exist. */}
                 <SingleLineCode value={session?.access_token ?? ''} />
             </Box>
+
+            <Typography variant="h6" sx={{ m: 2 }}>
+                <FormattedMessage id="admin.cli_api.refreshToken" />
+            </Typography>
+
+            <RefreshTokenTable />
         </>
     );
 }
--- a/src/components/tables/RefreshTokens/Rows.tsx
+++ b/src/components/tables/RefreshTokens/Rows.tsx
@@ -1,0 +1,49 @@
+import { TableCell, TableRow, useTheme } from '@mui/material';
+import { RefreshTokenQuery } from 'api/tokens';
+import { getEntityTableRowSx } from 'context/Theme';
+import { FormattedMessage } from 'react-intl';
+import TimeStamp from '../cells/TimeStamp';
+import RevokeTokenButton from '../cells/refreshTokens/RevokeToken';
+
+interface RowsProps {
+    data: RefreshTokenQuery[];
+}
+
+interface RowProps {
+    row: RefreshTokenQuery;
+}
+
+function Row({ row }: RowProps) {
+    const theme = useTheme();
+
+    return (
+        <TableRow hover sx={getEntityTableRowSx(theme)}>
+            <TimeStamp time={row.created_at} enableRelative />
+
+            <TableCell>{row.user_id}</TableCell>
+
+            <TableCell>{row.detail}</TableCell>
+
+            <TableCell>
+                <FormattedMessage
+                    id="admin.cli_api.refreshToken.table.label.uses"
+                    values={{ count: row.uses }}
+                />
+            </TableCell>
+
+            <RevokeTokenButton id={row.id} />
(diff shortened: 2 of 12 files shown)


src/components/tables/RefreshTokens/Rows.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function Row({ row }: RowProps) {
+    const theme = useTheme();
+
+    return (
+        <TableRow hover sx={getEntityTableRowSx(theme)}>
+            <TimeStamp time={row.created_at} enableRelative />
+
+            <TableCell>{row.detail}</TableCell>
```
travjenkins, line 23: Please put a max width and wrap text on this column to prevent the _very very very rare_ chance that the string is super long.

image

src/components/tables/RefreshTokens/Rows.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface RowProps {
+    row: RefreshTokenQuery;
+}
+
+function Row({ row }: RowProps) {
+    const theme = useTheme();
+
+    return (
+        <TableRow hover sx={getEntityTableRowSx(theme)}>
+            <TimeStamp time={row.created_at} enableRelative />
```
travjenkins, line 21: What is driving this not following how we handle the `published` column on the entity tables?

image
`vs`
image

src/lang/en-US.ts
```diff
@@ -457,8 +461,21 @@ const AdminPage: ResolvedIntlConfig['messages'] = {
     'admin.roles.message': `These are all the ${CommonMessages['terms.permissions']} that are currently provisioned. An administrator can update them in the {docLink}.`,
     'admin.roles.message.docLink': `authorization settings`,
     'admin.roles.message.docPath': `https://go.estuary.dev/provision`,
-    'admin.accessToken': `Access Token`,
-    'admin.accessToken.message': `Copy the access token below to authenticate the CLI client.`,
+
+    'admin.cli_api.header': `Programmatic Access to Flow`,
+    'admin.cli_api.message': `Use Refresh and Access tokens to connect to Flow programmatically.`,
```
travjenkins, line 466: I think I recall Olivia making the call we should not capitalize words as much as possible. So I think `Refresh` and `Access` should not be capitalized.

src/components/tables/cells/refreshTokens/RevokeToken.tsx
```diff
@@ -0,0 +1,92 @@
...
+        );
+
+        if (response.error) {
+            setError(response.error);
+            setSaving(false);
+
+            return;
+        }
+
+        hydrate();
```
travjenkins, line 53: This should be `awaited`

src/components/admin/Api/RefreshToken/Dialog/GenerateButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+            return;
+        }
+
+        hydrate();
+
+        // The refresh token ID and secret are needed by Flow, therefore it was decided
+        // to base64 encode the data returned in the response and present that as the
+        // one-time secret presented to the user.
+        setToken(Buffer.from(JSON.stringify(response.data)).toString('base64'));
```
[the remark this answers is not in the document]
travjenkins: Also, we should probably make this not a single line to catch any possible errors. Like if _somehow_ the `response.data` causes the stringify call to return `undefined` we'd probably want to catch that.

src/components/admin/Api/RefreshToken/Dialog/GenerateButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+        setSaving(false);
+
+        if (response.error || isEmpty(response.data)) {
+            setServerError(response.error);
+
+            return;
+        }
+
+        hydrate();
```
travjenkins, line 53: This should also be `awaited`

src/components/admin/Api/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+                    <FormattedMessage id="admin.cli_api.accessToken" />
+                </Typography>
+
+                <Typography sx={{ mb: 3 }}>
+                    <FormattedMessage id="admin.cli_api.accessToken.message" />
+                </Typography>
+
+                {/* TODO (defect): Display an error in the event the access token does not exist. */}
+                <SingleLineCode value={session?.access_token ?? ''} />
+            </Box>
```
travjenkins, line 60: This should be broken out to a stand alone component just to keep it in sync with how the `Refresh` stuff is handled

src/components/admin/Api/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const boxStyling: SxProps<Theme> = {
+    padding: 2,
+};
+
+function AdminApi() {
+    usePageTitle({
+        header: authenticatedRoutes.admin.api.title,
+        headerLink:
+            'https://docs.estuary.dev/reference/authentication/#authenticating-flow-using-the-cli',
```
travjenkins, line 18: This should be removed and this link should move down to be next to the `Access Token` header like `Cloud Storage` does it.

src/components/content/MessageWithLink.tsx
```diff
@@ -1,13 +1,14 @@
-import { Box } from '@mui/material';
+import { Box, SxProps } from '@mui/material';
 import ExternalLink from 'components/shared/ExternalLink';
 import { FormattedMessage, useIntl } from 'react-intl';
 
 interface Props {
     messageID: string;
     link?: string;
+    linkSx?: SxProps;
```
travjenkins, line 8: We should replace this with just passing any of the `ExternalLinks` props in and spreading them into the component.

src/lang/en-US.ts
kiahna-tucker: This content came directly from Dave.

src/components/admin/Api/RefreshToken/Dialog/GenerateButton.tsx
kiahna-tucker: This select table store action is improperly typed -- returning `void` instead of a promise -- therefore it cannot be awaited. Changing the type annotation has ripple effects across all select table stores.

src/components/tables/cells/refreshTokens/RevokeToken.tsx
kiahna-tucker: This select table store action is improperly typed -- returning `void` instead of a promise -- therefore it cannot be awaited. Changing the type annotation has ripple effects across all select table stores.

src/components/tables/RefreshTokens/Rows.tsx
kiahna-tucker: The formatting is closest to what Dave and Johnny specified in their document without being novel.

aa

src/components/admin/Api/RefreshToken/Dialog/GenerateButton.tsx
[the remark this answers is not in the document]
kiahna-tucker: > Was there a specific reason we didn't use btoa here?

`btoa` is depreciated.

src/components/tables/cells/refreshTokens/RevokeToken.tsx
travjenkins: you could just await this and then mark it as an error expected

travjenkins: If you're okay with this getting ignored we're fine - so it is up to you

travjenkins (approved):
lgtm - didn't manually test again

Issue #998, opened 2024-02-26
Stop using `@supabase/ui` for auth


travjenkins:
Need to stop using the `supabase/ui` package for Auth as it has been deprecated. We probably want to switch over to the community package as they recommend: [github.com/supabase-community/auth-ui](https://github.com/supabase-community/auth-ui).

https://github.com/supabase/ui/blob/master/README.md

travjenkins:
Doing some quick research into this

travjenkins:
Some more research

Thought about looking into using `Lucia` but they have pulled Supabase support.
https://github.com/lucia-auth/lucia/discussions/325

PR #1099, opened 2024-04-30
Remove Supabase UI for auth

travjenkins:
## Issues

https://github.com/estuary/ui/issues/998

### 998

- Remove `@supabase/ui` from project
- Bring in a user context based on the new community Supabase project
- Switch over hooks to new context

### Manually tested

- Logged in and out with `oauth` and `magiclink`
- Messed with token cookie to get auto logged out

### Automated tests

- N/A

## Screenshots

N/A


src/app/guards/User.tsx
```diff
@@ -14,7 +14,7 @@ function UserGuard({ children }: BaseComponentProps) {
     //  everytime the user focuses on the tab we could end up spamming calls.
     const identifiedUser = useRef(false);
     const grantToken = useGlobalSearchParams(GlobalSearchParams.GRANT_TOKEN);
-    const { session } = Auth.useUser();
+    const { session } = useUser();
```
travjenkins, line 17: This is the main change in all files. Switching over to using the new hook.

src/components/login/MagicLink.tsx
```diff
@@ -76,45 +76,57 @@ const MagicLink = ({ grantToken, hideCodeInput }: Props) => {
...
+                    email: formData.email,
+                },
+                {
+                    redirectTo: redirectPath,
+                    shouldCreateUser: loginSettings.enableEmailRegister,
+                }
+            );
+        },
+        [redirectPath, supabaseClient.auth]
+    );
```
travjenkins, line 117: Moved some of this stuff into `memo`/`callbacks` to make sure they use the correct variable values. Just being a bit safer.

src/context/SWR.tsx
```diff
@@ -83,6 +83,9 @@ const SwrConfigProvider = ({ children }: BaseComponentProps) => {
                             Boolean(supabaseClient.auth.user());
 
                         if (localUserInvalid || tokenHasIssues(error.message)) {
+                            logRocketEvent(CustomEvents.AUTH_SIGNOUT, {
+                                trigger: 'swr',
+                            });
```
travjenkins, line 88: Added some eventing to keep an eye on when users get signed out. This even does not get trigger on every single call to `auth.signOut` but the main ones that auto log a user out due to an issue.

src/context/UserContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+                return;
+            }
+
+            // Set the latest session into state
+            setSession(authSession);
+
+            // Fetch the user info from the server to ensure we know who the user truly is
+            const { data } = await supabaseClient.auth.api.getUser(
+                authSession.access_token
+            );
```
travjenkins, line 47: This is a bigger change. We now get user data directly from the server instead of the local storage. Even though we don't really ever use this info for security purposes this felt a little safer.

kiahna-tucker (approved):

PR #1103, opened 2024-05-01
Highlight the latest storage mapping in the table

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/435

### 435

* Create a shared table cell component that displays a status in a chip.

* Add a status column to the storage mappings table which displays an _active_ status chip on the first table row (which corresponds to the storage mapping used for write operations).

* Consolidate the first and second rows of the storage mappings table so that the _new_ first row displays data in every column.

* Customize the styling of colored, outlined chips in the theme.

### Manually tested

* Review the appearance of the storage mapping table.

### Automated tests

N/A

## Screenshots

**Storage mappings table | Multiple mappings | Light mode**

pr_screenshot-1103-mappings_table-multi-light

<br />
<br />

**Storage mappings table | Multiple mappings | Dark mode**

sm_table-outlined-2-right

<br />
<br />

**Storage mappings table | Single mapping | Light mode**

pr_screenshot-1103-mappings_table-single-light

<br />
<br />

**Storage mappings table | Single mapping | Dark mode**

pr_screenshot-1103-mappings_table-single-dark

### DIFF
--- a/src/components/tables/StorageMappings/Rows.tsx
+++ b/src/components/tables/StorageMappings/Rows.tsx
@@ -57,42 +57,27 @@ function DataCells({ store }: DataCellProps) {
 
 function Row({ row }: RowProps) {
     const key = `StorageMappings-${row.id}`;
-    const multipleStores = row.spec.stores.length > 1;
-
-    if (multipleStores) {
-        return (
-            <>
-                {row.spec.stores.map((store, index) =>
-                    index === 0 ? (
-                        <TableRow key={`${key}_stores_${index}`}>
-                            <TableCell>{row.catalog_prefix}</TableCell>
-                            <ChipStatus
-                                color="success"
-                                messageId="data.active"
-                            />
-                            <DataCells store={store} />
-                            <TimeStamp time={row.updated_at} enableRelative />
-                        </TableRow>
-                    ) : (
-                        <TableRow key={`${key}_stores_${index}`}>
-                            <TableCell />
-                            <TableCell />
-                            <DataCells store={store} />
-                            <TableCell />
-                        </TableRow>
-                    )
-                )}
-            </>
-        );
-    }
 
     return (
-        <TableRow key={key}>
-            <TableCell>{row.catalog_prefix}</TableCell>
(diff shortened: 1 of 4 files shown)


src/components/tables/cells/ChipStatus.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <TableCell>
+            <Chip
+                component="span"
+                label={intl.formatMessage({ id: messageId })}
+                size="small"
+                variant="outlined"
+                sx={{
+                    borderColor: (theme) => getOutlineColor(theme, color),
+                    backgroundColor: outlinedColoredChipBackground[color],
+                }}
```
travjenkins, line 28: What's driving customized styling?

kiahna-tucker: An outlined chip with a text and background color set by the `color` property is hard to read in dark mode and the color is hard to distinguish as _green_ in light mode since the dark variant would need to be used for readability; a filled chip with a text and background color set by the `color` property is hard to read in both color modes.

src/components/tables/StorageMappings/Rows.tsx
travjenkins, line 1: Didn't catch this before - I don't _think_ we need the `multipleStores` check. All the paths should just run through the map call to keep things simple.

kiahna-tucker: I was going to ask about that. It should be able to be removed without issue.

src/components/tables/cells/ChipStatus.tsx
kiahna-tucker: I can look into it. Are you thinking of adjusting the styling of all outlined chips that have a non-default color?

travjenkins: yeah

src/context/Theme.tsx
travjenkins, line 1: FYI - I think this will be overridden if any 'sx` is passed to a chip

travjenkins (approved):
lgtm

PR #1113, opened 2024-05-10
Add CSV download to entity tables

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1114

### 1114

- Adding in a new `ExportComponent` to pass into the table to render a download button
- Add new hooks to generate the data to download for each entity type
- Allow 100 entities to be viewed

### Related/Misc
- Break up the stats query into part to allow for 100 entities with long names to be rendered
- Share some variables

### Manually tested

### Automated tests

-   unit testing covered

### File Names
image

#### Capture
image

#### Collections
image

#### Materializations
image


src/api/stats.ts
```diff
@@ -130,70 +132,89 @@ export const convertToUTC = (
...
-
-        default:
-            throw new Error('Unsupported filter used in Stats Query');
+const getStatsByName = async (names: string[], filter?: StatsFilter) => {
+    const limiter = pLimit(3);
+    const promises: Array<Promise<PostgrestResponse<CatalogStats>>> = [];
+    let index = 0;
+
+    // TODO (retry) promise generator
+    const promiseGenerator = (idx: number) => {
```
travjenkins, line 141: The bulk of this change is just moving the previous function inside of the `promiseGenerator`

src/api/stats.ts
```diff
@@ -130,70 +132,89 @@ export const convertToUTC = (
...
+        index = index + CHUNK_SIZE;
     }
 
-    return supabaseRetry(() => queryBuilder, '').then(
-        handleSuccess<CatalogStats[]>,
-        handleFailure
-    );
+    const response = await Promise.all(promises);
+    const errors = response.filter((r) => r.error);
+    return errors[0] ?? { data: response.flatMap((r) => r.data) };
```
travjenkins, line 217: This is one of the few places where we need to get all the data back so needing to map the response.

src/components/tables/Captures/useCaptureColumns.tsx
```diff
@@ -1,31 +1,19 @@
...
-        headerIntlKey: 'entityTable.data.writesTo',
-    },
-    {
-        field: 'updated_at',
-        headerIntlKey: 'entityTable.data.lastPublished',
-    },
+    catalogName,
+    connectorType,
+    writesTo,
+    lastPublished,
```
travjenkins, line 16: Shared these things because initially they were all part of the columns. They no longer are but still nice to share some stuff between tables to leaving it in.

src/components/tables/EntityTable/index.tsx
```diff
@@ -98,12 +99,13 @@ function EntityTable({
     hideFilter,
     hideHeaderAndFooter,
     rowsPerPage,
-    rowsPerPageOptions = [10, 25, 50],
+    rowsPerPageOptions = [10, 25, 50, 100],
```
travjenkins, line 102: Bumping up to 100 here... might end up needing to allow more.

Issue #1116, opened 2024-05-15
Show more metadata for rows of rendered schema tables


travjenkins:
There are some extra details about the schema that could be displayed in the UI while viewing. These details are visible only when editing and would be great to have all the time.

Some of the extra details visible only in edit:
![image](https://github.com/estuary/ui/assets/270078/0af740b5-77a6-4b34-bcfe-3a95cdc576be)

Request from Johnny `via Slack`
```
I myself would love to see:
Exists removed, where we perhaps represent "must" with bolding of the row field.
A merged "Detail" column which combines title and/or description
We have a lot of meaningful metadata for what fields are carried by these annotations, and when people ask what various /_meta columns mean, this is really where I'd like to point them.
```

PR #1117, opened 2024-05-15
Adding a details column to the schema viewer

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1116

### 1116

- Display the `string_format` for the `string` type
- Added a new `details` column to display the `title` and `description`
- filtered to must `must` and `may` to reduce confusion
- added a filter for `exist`

### Manually tested

### Automated tests

-   unit testing covered

## Screenshots

We show both the title and description if we have it
image

Filter can be all, must, or may
image
image
image

If the filter is empty but there are fields we show a message just for that
image

If there is an invalid schema the filter will be disabled
image

Filter not shown in edit mode
image

Showing string format when available
image


src/components/schema/ExistFilter.tsx
```diff
@@ -0,0 +1,56 @@
...
+        },
+    ];
+
+    return (
+        <AutocompletedField
+            label={intl.formatMessage({
+                id: 'schemaEditor.table.filter.label',
+            })}
+            options={fieldOptions}
+            defaultValue={fieldOptions.find(
```
kiahna-tucker, line 41: Changing the default value on a controlled `Autocomplete` component will result in an error reported in the console; the `value` property should be set in this case. We discussed this earlier today, but I wanted to leave a comment for record.

src/components/schema/ExistFilter.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useIntl } from 'react-intl';
+import { AutoCompleteOption, FieldFilter } from './types';
+
+interface Props {
+    fieldFilter: FieldFilter;
+    setFieldFilter: (value: FieldFilter) => void;
+    disabled?: boolean;
+}
+
+function ExistFilter({ fieldFilter, setFieldFilter, disabled }: Props) {
```
kiahna-tucker, line 11: `ExistenceFilter` is an alternative component name that came to mind which _may_ increase clarity.

src/components/schema/types.ts
```diff
@@ -1,4 +1,12 @@
 import { autoCompleteDefaults_Virtual_Multiple } from 'components/shared/AutoComplete/DefaultProps';
+import { InferSchemaResponseProperty } from 'types';
 
 export type OnChange =
     (typeof autoCompleteDefaults_Virtual_Multiple)['onChange'];
+
+export type FieldFilter = 'all' | InferSchemaResponseProperty['exists'];
+
+export interface AutoCompleteOption {
+    id: FieldFilter;
```
kiahna-tucker, line 10: Curious if `key` would be a more fitting name for this property.

src/components/shared/toolbar/AutocompletedField.tsx
kiahna-tucker, line 1: This component could use the `Autocomplete` props supplied by MUI directly.

kiahna-tucker (approved):
Approved with moderate testing. All of the stylistic and UX-related concerns were discussed orally and need not be documented here.

src/components/schema/types.ts
travjenkins: I treat `key` as reserved in React

src/components/shared/toolbar/AutocompletedField.tsx
travjenkins: What does this mean?

src/components/schema/ExistFilter.tsx
travjenkins: done

PR #1122, opened 2024-05-21
Remove use of `LoadingButton` due to issues with Google Translate

travjenkins:
## Issues

https://github.com/estuary/ui/issues/915

### 915

- Remove all uses of `LoadingButton` and replace with a new one that wraps the children in `span`
- Added an eslint rule to prevent the user of `LoadingButton` anymore

### Misc
- Cleaned up a Marketplace file that was not needed

### Manually tested

- Hit pages and see buttons working
- Imported `LoadingButton` and saw lint issue in IDE

### Automated tests

- n/a


src/components/admin/Settings/PrefixAlerts/Dialog/SaveButton.tsx
travjenkins, line 1: May have been causing issues since it lacked a span

src/components/tables/cells/refreshTokens/RevokeToken.tsx
travjenkins, line 1: May have been causing issues since it lacked a span

src/components/transformation/create/GitPodButton.tsx
```diff
@@ -228,13 +229,12 @@ function GitPodButton({
...
-                    intl.formatMessage({
-                        id: 'newTransform.button.cta',
-                    })}
-            </span>
-        </LoadingButton>
+            {submitButtonError
+                ? submitButtonError
+                : intl.formatMessage({
+                      id: 'newTransform.button.cta',
+                  })}
```
travjenkins, line 236: Switched to the ternary operator to be safer

kiahna-tucker: If `submitButtonError` had an explicit type annotation, the original operator would be used with a high degree of confidence. Whatever you feel comfortable with is fine with me.

kiahna-tucker (approved):

src/components/transformation/create/GitPodButton.tsx
travjenkins: It is still against a rule though - so sometimes my IDE will complain.

PR #1124, opened 2024-05-22
browser support update

travjenkins:
periodical tech debt


kiahna-tucker (approved):
Approved with minimal testing. Validated that the application could run locally and navigated around the application without issue.

Issue #913, opened 2024-01-03
Sometimes not getting redirected after a successful publication


psFried:
**Bug**

I just had [a session](https://app.logrocket.com/tue20y/estuarytest/s/5-58b2cd5c-2c28-4800-b203-7675697dbb55/13?network_id=3-fetch-95&network_root_tab_id=fa723cf1-a2c4-447f-a52e-430fd4b98f0c&network_tab_id=fa723cf1-a2c4-447f-a52e-430fd4b98f0c&t=1704314131455.037) where the logs window seemed to close automatically after my publication succeeded, but I didn't get redirected away from the edit page. The link should take you to right before the logs window closes. Then you'll see me hit "save and publish" again (I'd stepped away, so wasn't sure if the publication worked or not) and then got a "new row violates RLS policy" error.

**Expected**

That either the logs window would remain open until I close it, or else I'd be redirected to a page with some sort of "success" message.

**Screenshots**
Include a screenshot of what you saw when the bug happened.
<img width="1136" alt="Screen Shot 2024-01-03 at 3 56 41 PM" src="https://github.com/estuary/ui/assets/4495829/5506aeb0-98ef-4646-8959-cdae5b3637d8">

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser: Firefox

travjenkins:
Have a session to dig into 

https://app.logrocket.com/tue20y/estuarytest/s/5-ed055937-08cc-4490-8fbf-5b8bd5680c0b/46?dashboardID=logrocket-overview&filters=%255B%257B%2522id%2522%253A%2522S1TFqEZ7C%2522%252C%2522type%2522%253A%2522email%2522%252C%2522operator%2522%253A%257B%2522name%2522%253A%2522is%2522%252C%2522type%2522%253A%2522IS%2522%252C%2522hasStrings%2522%253Atrue%252C%2522autocompleteEnabled%2522%253Atrue%257D%252C%2522strings%2522%253A%255B%2522chase%2540eyeballdivision.com%2522%255D%257D%255D&fromTab=&persistForm=true&range=Custom&rangeEnd=1715716830349&rangeStart=1715659200000&t=1715714546721.5164

travjenkins:
I am 99% sure that this is because on materializations we run a refresh and that can update the store to hide logs right while you are publishing.

PR #1126, opened 2024-05-22
Try to make create/edit process a bit more robust

travjenkins:
## Issues

https://github.com/estuary/ui/issues/913

### 913
- Remove the Test and Save buttons after a successful publish and replace with a 'View Details` button that takes the user to the details
- Added some error handling for super edge cases where the publish might have a call failed during it.
- Updating the `Test` and `Save & Publish` buttons to show `Testing` and `Saving & Publishing` while there is an action.
- Added `Saved` to the active state so forms cannot be changed after being saved.

### misc
- Update Icon Noir. This was initially part of an experiment but the work is valid so leaving it in
- Adding in a disabled state for remove for when it is removing
- Added a check that the call to the server to remove the binding actually worked before removing it from the store
- Made the removal of bindings a bit more robust to potential errors. This really only happened if a user would click on the remove button of multiple bindings quickly. 
- Fixed some random DOM issues that Chrome called out where `id`s and `name`s were missing on inputs/buttons.
- Fixed issue where time travel was editable during an async action to the server

### Manually tested

### Automated tests

-   unit testing covered

## Screenshots


src/components/capture/useCaptureDraftUpdate.ts
```diff
@@ -115,6 +115,17 @@ function useDiscoverDraftUpdate(options?: {
...
 
+                if (draftSpecsResponse.data.length === 0) {
+                    callFailed({
+                        error: {
+                            title: 'captureCreate.generate.failedErrorTitle',
+                        },
+                    });
+
+                    return false;
+                }
```
travjenkins, line 126: Rare - but wanted to handle this error just in case.

src/utils/publication-utils.ts
```diff
@@ -0,0 +1,23 @@
+import { logRocketConsole } from 'services/shared';
+import { JOB_STATUS_SUCCESS } from 'services/supabase';
+import { hasLength } from './misc-utils';
+
+export const checkIfPublishIsDone = (payload: any): [boolean | null, any] => {
```
travjenkins, line 5: This was broken out for stuff that was reverted - but leaving it stand alone makes sense.

src/components/collection/Selector/List/index.tsx
```diff
@@ -102,9 +102,8 @@ function CollectionSelectorList({
     const formStatus = useFormStateStore_status();
 
     const selectionEnabled =
-        currentBindingUUID &&
-        setCurrentBinding &&
-        formStatus !== FormStatus.UPDATING;
+        // currentBindingUUID &&
```
kiahna-tucker, line 105: This commented out line of code can be removed outright if it is no longer needed.

src/components/shared/Entity/Actions/SaveButton.tsx
```diff
@@ -36,6 +36,8 @@ function EntitySaveButton({ disabled, taskNames, logEvent }: Props) {
     const logToken = useFormStateStore_logToken();
     const formStatus = useFormStateStore_status();
 
+    const isSaving = formStatus === FormStatus.SAVING;
```
kiahna-tucker, line 39: This variable naming convention is _okay_ but, from my vantage point, _is_ is a redundant prefix when the variable name is a verb with an _-ing_ suffix.

src/components/shared/Entity/HeaderLogs.tsx
```diff
@@ -0,0 +1,44 @@
...
+    useFormStateStore_messagePrefix,
+    useFormStateStore_showLogs,
+    useFormStateStore_status,
+} from 'stores/FormState/hooks';
+import { FormStatus } from 'stores/FormState/types';
+import useEntityWorkflowHelpers from './hooks/useEntityWorkflowHelpers';
+import LogDialog from './LogDialog';
+import LogDialogActions from './LogDialogActions';
+
+// TODO (break logs from buttons) - this is starting to display the dialog as a stand alone
```
kiahna-tucker, line 13: Assuming this TODO comment identifies a task out of scope for this PR.

src/stores/FormState/Store.ts
```diff
@@ -14,7 +14,10 @@ const formActive = (status: FormStatus) => {
         status === FormStatus.GENERATING ||
         status === FormStatus.SAVING ||
         status === FormStatus.SCHEMA_EVOLVING ||
-        status === FormStatus.UPDATING
+        status === FormStatus.UPDATING ||
+        // TODO (workflow stores) need to manage form state better
+        // This is crappy - worry. But if we have saved we want to disable everything and this is quickest way
```
kiahna-tucker, line 19: > This is crappy - worry.

I imagine _worry_ is a typo and you meant to say _sorry._

src/stores/Binding/Store.ts
```diff
@@ -1043,13 +1049,27 @@ const getInitialState = (
                     ? state.resourceConfigs[targetBindingUUID]
                     : null;
 
+                const targettedResourceConfig =
```
kiahna-tucker, line 1052: This variable could simply be `targetResourceConfig` since it is the compliment of `targetBindingUUID`.

kiahna-tucker (commented):
Posting the code review comments while I test the PR.

src/components/shared/Entity/hooks/useEntityWorkflowHelpers.ts
```diff
@@ -142,21 +143,30 @@ function useEntityWorkflowHelpers() {
...
         },
         [catalogName, generatePath, navigate, resetState]
     );
 
     // Form Event Handlers
-    const closeLogs = useCallback(() => {
-        setFormState({
-            showLogs: false,
-        });
+    const closeLogs = useCallback(
```
kiahna-tucker, line 153: This function will need to be explicitly called wherever it is used (i.e., `LogDialogActions`, `HeaderLogs` [unused component]) since an input argument has been introduced. Otherwise, `navigateOnly` will be a button event object and, thus, always truthy.

src/components/shared/Entity/HeaderLogs.tsx
travjenkins: Yeah... I didn't want to lose this component as it is really nice to help with extra renders but also didn't want an old file floating around.

kiahna-tucker (changes requested):
The main issue is identified in a comment below.

And did we ultimately decide to no longer allow the _Close_ CTA in the publication dialog to navigate the user to the _Details_ page when clicked after the publication job completed? The _View Details_ CTA was more an alternate means to navigate to the _Details_ page in my recollection, to be used when a user closes the publication dialog before the publication job completes.

src/components/collection/Selector/List/index.tsx
travjenkins: Good catch... this needs added back in. This was a part of testing out not setting a default binding to show.

src/components/shared/Entity/hooks/useEntityWorkflowHelpers.ts
travjenkins: dang it... yeah... I'll just take this out as it was a quick add at the end to try to make things better

travjenkins: also... this was _really_ dumb since we already return `exit`

src/components/shared/Entity/Actions/SaveButton.tsx
travjenkins: I'm fine either way but this followed the `isTesting` in the other file. I could go either way with removing the `is` from booleans denoting a verb-ing.

travjenkins: I forgot to use this variable in the logic right below... switched out the name.

src/stores/Binding/Store.ts
```diff
@@ -658,15 +660,19 @@ const getInitialState = (
...
+                            evaluatedBindingUUID && evaluatedResourceConfig
+                        )
+                            ? {
+                                  uuid: evaluatedBindingUUID,
+                                  collection:
+                                      evaluatedResourceConfig.meta
+                                          .collectionName,
+                              }
+                            : null;
+                    }
```
travjenkins, line 675: We used to keep the current selected during removing... I think that works a bit better. So putting that back in.

src/stores/Binding/Store.ts
```diff
@@ -1043,13 +1049,27 @@ const getInitialState = (
...
+                    logRocketEvent(
+                        CustomEvents.BINDINGS_RESOURCE_CONFIG_MISSING,
+                        {
+                            targetBindingUUID,
+                            targetCollection,
+                        }
+                    );
+
+                    return;
+                }
```
travjenkins, line 1066: This should not really happen and it was really hard to reproduce but putting this here and hoping we can see if this keeps happening.

kiahna-tucker (approved):
Approved with moderate testing. Primarily retested the erring scenario noted in the previous review comment in all affected workflows.

Issue #1137, opened 2024-06-04
Supabase V2: Update Supabase-JS client and cache helper


travjenkins:
Want to get the UI using the newer V2 Supabase client library. Also, we should get off the custom swr handler we have right now and replace with the community one https://supabase-cache-helpers.vercel.app/postgrest/getting-started

The first step of this will **not** consume the type that is generated by Supabase CLI. We'll stick with the manual typing to keep the scope of this a bit smaller.

- [x] Install new versions
- [x] Remove old SWR hook
- [x] Update typing
- [x] Switch to new selector format
- [x] Get insert/delete/etc. returning data
- [x] test everything

PR #1138, opened 2024-06-05
Tech Debt: Upgrade `supabase` client to `V2`  and migrate to community `swr` hook

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1137

### User/Session

- Storing the details in a store so they don't need fetched all the time
- Add new direct hooks to fetch session and users

### Misc

- Move the `flow_deps` under the new `deps` folder to keep all the deps together
- Add in playwright testing support for local testing 
- Replace `lru-cache` with `mnemonist` as we do not need all the features.

### Manually tested

Hitting as many network calls as I could

### Automated tests

Added in some basic tests for Captures, Collections, and Admin. The capture tests are the only ones that are _really_ like tests. But this is a start for using Playwright to run local E2E tests.

## Screenshots


.eslintrc.js
```diff
@@ -63,6 +63,11 @@ module.exports = {
                 name: '@emotion/react',
                 message: 'Do not access emotion directly. Load through MUI',
             },
+            {
+                name: '@supabase/supabase-js',
+                importNames: ['PostgrestResponse'],
+                message: 'Please use @supabase/postgrest-js',
+            },
```
travjenkins, line 70: Want to try to get the deps from their source and not an intermediary package.

deps/supabase/types.d.ts
travjenkins, line 1: This is not being used just yet but adding anyway because we'll need to use it and track it soon.

playwright-tests/package.json
```diff
@@ -0,0 +1,24 @@
...
+        "codegen:setup": "npm run env:setup && NODE_ENV=test npm --prefix ../ start",
+        "codegen": "playwright codegen http://localhost:8082/project/default"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "devDependencies": {
+        "@playwright/test": "^1.41.2",
+        "@types/node": "^20.11.16",
+        "pg": "^8.12.0"
```
travjenkins, line 19: Not being used right now but leaving in - hoping that we can end up writing some `pg` scripts to access the Supabase DB.

scripts/generateSupabaseTypes.js
travjenkins, line 1: New helper script to use the Supabase CLI to generate types

src/api/__tests__/combinedGrantsExt.test.ts
travjenkins, line 1: Some of these old tests just don't make sense to get working since we can more easily write new Playwright tests to replace later on

src/api/alerts.ts
```diff
@@ -132,10 +132,11 @@ const getNotificationSubscriptionForUser = async (
     const data = await supabaseRetry(
         () =>
             supabaseClient
-                .from<AlertSubscriptionQuery>(TABLES.ALERT_SUBSCRIPTIONS)
+                .from(TABLES.ALERT_SUBSCRIPTIONS)
                 .select(`id, catalog_prefix, email`)
                 .eq('catalog_prefix', prefix)
-                .eq('email', email),
+                .eq('email', email)
+                .returns<AlertSubscriptionQuery[]>(),
```
travjenkins, line 139: This will be one of the patterns of changes where we manually set the return of the queries for now since we are not using the typing when creating client.

src/api/draftSpecs.ts
```diff
@@ -169,7 +166,7 @@ export const deleteDraftSpecsByCatalogName = async (
         // In case we get an absolutely massive amount of catalogs to delete,
         // we don't want to spam supabase
         const limiter = pLimit(3);
-        const promises: Array<Promise<PostgrestResponse<any>>> = [];
+        const promises: Array<Promise<PostgrestSingleResponse<any>>> = [];
```
travjenkins, line 169: The typing from the library changed. In the code base it looks like the will further change this typing so there is only a single kind of response.

src/services/gateway-auth-config.ts
```diff
@@ -9,13 +9,13 @@ interface GatewayAuthTokenResponse {
     token: string;
 }
 
+const { gatewayAuthTokenEndpoint } = getGatewayAuthTokenSettings();
+const { supabaseAnonymousKey } = getSupabaseAnonymousKey();
```
travjenkins, line 13: No reason to keep etching these with each call

src/hooks/useUserGrants.ts
```diff
@@ -1,18 +1,18 @@
-import { useUser } from 'context/UserContext';
 import { getUserGrants } from 'api/userGrants';
 import { singleCallSettings } from 'context/SWR';
-import { useSelectNew } from './supabase-swr/hooks/useSelect';
+import { useQuery } from '@supabase-cache-helpers/postgrest-swr';
+import { useUserStore } from 'context/User/useUserContextStore';
 
 function useUserGrants(singleCall?: boolean) {
-    const { session } = useUser();
+    const user = useUserStore((state) => state.user);
```
travjenkins, line 7: New way to fetch user as we can fetch that directly instead of pulling it from the session. This is nice because sometimes the session can change but the user stays the same.

src/hooks/useUserDetails.ts
travjenkins, line 1: No longer needed as we store this off and don't need to generate constantly.

src/hooks/useTenants.ts
```diff
@@ -37,13 +37,12 @@ export function useTenantHidesDataPreview(entityName: string) {
...
-        useSelectSingleNew<TenantHidesDataPreview>(query, {
-            refreshInterval: 15000,
-        });
+    const { data, error, isValidating } = useQuery(query, {
+        refreshInterval: 1500,
+    });
 
     const response = useMemo(
-        () => (isDemo ? false : data ? Boolean(data.data.hide_preview) : null),
+        () => (isDemo ? false : data ? Boolean(data.hide_preview) : null),
```
travjenkins, line 45: Common change as the new fetcher no longer nests data like it used to.

src/pages/Auth.tsx
```diff
@@ -60,7 +59,7 @@ const Auth = () => {
 
             savingUser.current = false;
             await supabaseClient.auth.signOut();
-            navigate(unauthenticatedRoutes.login.path);
+            navigate(unauthenticatedRoutes.login.path, { replace: true });
```
travjenkins, line 62: This should have always been a replace

src/pages/Auth.tsx
```diff
@@ -74,9 +73,7 @@ const Auth = () => {
         if (!user) {
             savingUser.current = true;
             supabaseClient.auth
-                .getSessionFromUrl({
-                    storeSession: true,
-                })
+                .initialize()
```
travjenkins, line 76: This is the new  way that you kick off checking for auth/session

src/pages/Auth.tsx
```diff
@@ -86,8 +83,9 @@ const Auth = () => {
                         success();
                     }
                 })
-                .catch(() => {
+                .catch(async (error) => {
                     trackEvent('exception');
+                    await failed(error.message);
```
travjenkins, line 88: I think this is how some users got stuck on the loading page forever - so making sure we navigate away when there is a huge error.

src/services/supabase.ts
```diff
@@ -250,26 +196,28 @@ export const distributedTableFilter = <Data>(
     return queryBuilder;
 };
 
-export const getUserDetails = (user: User | null | undefined) => {
-    let userName, email, emailVerified, avatar, id;
+export const getUserDetails = (
```
travjenkins, line 199: Made some typing changes so how this handles "empty" users is a bit different.

src/services/supabase.ts
```diff
@@ -312,15 +260,19 @@ export const supabaseRetry = <T>(makeCall: Function, action: string) => {
             const response = await makeCall();
 
             const error = response.error;
+            const {
+                data: { session },
+                error: authError,
+            } = await supabaseClient.auth.getSession();
```
travjenkins, line 266: We now need to fetch user from the client with an awaited function as they removed direct access to session.

src/services/supabase.ts
```diff
@@ -370,7 +323,7 @@ export const updateSupabase = (
     matchData: any
 ): PromiseLike<CallSupabaseResponse<any>> => {
     return supabaseRetry(
-        () => supabaseClient.from(table).update(data).match(matchData),
+        () => supabaseClient.from(table).update(data).match(matchData).select(),
```
travjenkins, line 326: we have a lot of code that expect updates to the db to return what they changed.

src/test/test-utils.tsx
```diff
@@ -128,11 +126,8 @@ const MockProviders = ({ children, username }: any) => {
         mockAuthSession.user.user_metadata = generateMockUserMetadata(username);
     }
 
-    return (
-        <SwrSupabaseContext.Provider value={supabaseClient}>
-            {children}
-        </SwrSupabaseContext.Provider>
-    );
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    return <>{children}</>;
```
travjenkins, line 130: No longer need to provide the hold SWR Client

kiahna-tucker (commented):
Releasing code review comments while I finish manual testing. Any comment prefixed by _PIN_ can be ignored for now.

src/api/stats.ts
```diff
@@ -321,15 +321,14 @@ const getStatsForBillingHistoryTable = (
...
+    return defaultTableFilter<typeof query>(
+        query,
         ['ts'],
         searchQuery,
         sorting
-        // pagination
     );
 
-    return queryBuilder;
+    // return queryBuilder.returns<CatalogStats_Billing[]>();
```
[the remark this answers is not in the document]
travjenkins: Yes - I'll mark these kind of comments as "this is the potential typing to use"

src/api/stats.ts
```diff
@@ -321,15 +321,14 @@ const getStatsForBillingHistoryTable = (
         .lte('ts', convertToUTC(today, monthlyGrain))
         .or(subjectRoleFilters);
 
-    queryBuilder = defaultTableFilter<CatalogStats_Billing>(
-        queryBuilder,
+    return defaultTableFilter<typeof query>(
```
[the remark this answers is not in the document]
travjenkins: Can you show me why you say that? Cause my TS does not say that is the type

src/hooks/useLiveSpecs.ts
```diff
@@ -13,79 +15,60 @@ export interface LiveSpecsQuery extends Schema {
...
         isValidating,
     };
 }
 
-function useLiveSpecs_details(specType: Entity, catalogName: string) {
-    const draftSpecQuery = useQuery<LiveSpecsQuery>(
-        TABLES.LIVE_SPECS_EXT,
-        {
-            columns: `
+const LiveSpecsDetailsQuery = `
```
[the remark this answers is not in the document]
travjenkins: I stick with `Columns` when they are stored in an array and `Query` when stored in a string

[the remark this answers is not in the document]
travjenkins: I forgot to upper case it

src/services/supabase.ts
```diff
@@ -44,7 +35,8 @@ export const ERROR_MESSAGES = {
 export const tokenHasIssues = (errorMessage?: string) => {
     return (
         errorMessage &&
-        (errorMessage === ERROR_MESSAGES.jwtExpired ||
+        (errorMessage.startsWith('JWSError ') ||
```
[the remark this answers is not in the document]
travjenkins: Yes... because otherwise you would not have left this comment and I would not have noticed this being added and huge issues could have happened in prod LOL

I add errors like this only for testing. We should never check for errors that are not stored in the list up above
image

src/context/User/useUserContextStore.ts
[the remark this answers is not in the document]
travjenkins: I don't know what you're asking.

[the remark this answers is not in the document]
travjenkins: This store is like the Details Usage store

kiahna-tucker (approved):
Approved without manually testing changes made after the review on 7/1.

src/context/User/useUserContextStore.ts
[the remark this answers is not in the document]
kiahna-tucker: The details usage store is equally an anomaly. When refactoring existing stores, the function that defines the initial state (e.g., `getInitialState`) has been preserved (ref: [issue 1088](https://github.com/estuary/ui/issues/1088)). Is the removal of the aforementioned intermediate function a new pattern you would like established? If so, should it be used on a go forward basis or selectively? If selectively, what dictates whether the intermediate function should be omitted?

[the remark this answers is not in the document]
travjenkins: Oh I see... if they are small enough states the extra function is just a waste. 

Also, I _think_ Zustand has solved that issue... I remember seeing something where it could reset to the initial state for you now.

PR #1146, opened 2024-06-17
Adding a warning so folks running locals know about missing logs

travjenkins:
Adding a warning for developers

### Dev
image

### Prod
image


kiahna-tucker (approved):

PR #1157, opened 2024-07-09
Reduce how often we check for hiding preview

travjenkins:
Making this ping a little bit less often.
Moved some stuff related to tenant store into the hook and not repeating code.
Started using `shallow` in the tenant store.


kiahna-tucker (approved):

src/components/hero/AcceptDemoInvitation.tsx
```diff
@@ -82,7 +82,7 @@ function AcceptDemoInvitation({
     };
 
     return (
-        <>
+        <Stack>
```
travjenkins, line 85: Make sure the error shows vertically instead of horizontally

image

src/stores/Entities/hooks.ts
```diff
@@ -12,24 +12,33 @@ import { useEntitiesStore } from './Store';
...
-    }));
+    return useEntitiesStore(
+        useShallow((state) =>
+            Object.keys({
+                ...state.capabilities.admin,
+                ...state.capabilities.write,
+                ...state.capabilities.read,
+            })
+        )
+    );
```
travjenkins, line 23: We were always using `Object.keys` so I think we can pull this into the store and stop repeating that code so much.

Also adding shallow.

src/hooks/useTenants.ts
```diff
@@ -38,7 +39,7 @@ export function useTenantHidesDataPreview(entityName: string) {
     }, [isDemo, tenantName]);
 
     const { data, error, isValidating } = useQuery(query, {
-        refreshInterval: 1500,
+        refreshInterval: EXTENDED_POLL_INTERVAL,
```
travjenkins, line 42: Give this query a bit more time to breath.

kiahna-tucker (approved):

PR #1158, opened 2024-07-09
Try to get DOM manipulation crashing better and data plane gateway token cache

travjenkins:
## Issues

`Tech Debt`

### DOM

- No longer conditionally include the stats columns and just shrink/hide them. That way translate messing with the DOM should not matter as much.

### Token Cache

- Noticed we were always calling to get new gateway auth tokens. With some changes in how we fetch them there was no longer a reason to keep them stored in local storage. So added a tiny LRUMap to house the cache and manually cache 'em. This should be an improvement but not perfect.

### Table Hydrating
- Added in a hack to prevent duplicate calls caused by query param stuff.

### Details hydration
- Made it so there is only a single call to live specs for details instead of making one call for details and another to get the spec,name,etc.

### Billing
- Update the stats query to no longer use `ilike` and replace with `like`

### Manually tested

- Hit all the tables in the system
- Searched and kept an eye on when we fetched new tokens

-   [x] Admin
-   [x] Captures
-   [x] Collections
-   [x] HomePage
-   [x] Materialization:
-   [x] Login

### Tables Loading
image

image
image

image
image

These two tables fetch different stuff from the same tables
image
image

image

image


src/components/admin/Tabs.tsx
```diff
@@ -4,13 +4,12 @@ import { useTenantDetails } from 'context/fetcher/Tenant';
...
 import { useIntl } from 'react-intl';
 import { Link, useLocation } from 'react-router-dom';
-import { hasLength } from 'utils/misc-utils';
 
 function AdminTabs() {
     const intl = useIntl();
     const { pathname } = useLocation();
     const [selectedTab, setSelectedTab] = useState(0);
-    const tenantDetails = useTenantDetails();
+    const { hasTenants } = useTenantDetails();
```
travjenkins, line 12: This is repeated... so threw in the length check into the context to reduce how often we were checking that.

src/components/collection/Selector/Table/Hydrator.tsx
```diff
@@ -66,6 +70,7 @@ function Hydrator({ entity, selectedCollections }: TableHydratorProps) {
 
     return (
         <TableHydrator
+            disableQueryParamHack={disableQueryParamHack}
```
travjenkins, line 73: Passing through the hack disable-er so that the Collections Selector pop up works AND the new beta transform create flow works.

src/components/shared/Entity/AddDialog.tsx
```diff
@@ -42,6 +42,7 @@ function AddDialog({
                         <Box>
                             {open ? (
                                 <BindingSelectorTable
+                                    disableQueryParamHack
```
travjenkins, line 45: If a table opens without a URL update first it does not require the hack

src/components/shared/TenantSelector.tsx
```diff
@@ -17,7 +17,7 @@ interface Props {
 function TenantSelector({ updateStoreState }: Props) {
     const intl = useIntl();
 
-    const tenants = useTenantDetails();
+    const { tenants } = useTenantDetails();
```
travjenkins, line 20: Now that there are two props in the context need to deconstruct.

src/components/tables/AccessGrants/AccessLinks/index.tsx
```diff
@@ -68,6 +68,7 @@ function AccessLinksTable() {
     return (
         <Box>
             <TableHydrator
+                disableQueryParamHack
```
travjenkins, line 71: Opens in a modal

src/components/tables/Captures/Rows.tsx
```diff
@@ -63,25 +60,15 @@ function Row({ isSelected, setRow, row, stats, showEntityStatus }: RowProps) {
...
-            {hasLength(tenantDetails) ? (
-                <>
-                    <Bytes
-                        val={
-                            stats
-                                ? stats[row.catalog_name]?.bytes_written_by_me
-                                : null
-                        }
-                    />
+            <Bytes
```
travjenkins, line 63: This is an attempt to both handle issues where a browser translating the app could crash stuff. Also, makes the code a lot simpler to think through as now we will always have the same columns and just hide them when needed.

src/components/tables/Captures/useCaptureColumns.tsx
```diff
@@ -21,34 +32,8 @@ const defaultColumns: ColumnProps[] = [
...
-    return useMemo(() => {
-        if (hasDetails) {
-            const response = [...defaultColumns];
-            response.splice(3, 0, statsHeader);
-            return response;
-        } else {
-            return defaultColumns;
-        }
-    }, [hasDetails]);
+    return useMemo(() => defaultColumns, []);
```
travjenkins, line 36: This could just be shared through a `const` but leaving the hook in to reduce scope a bit and in case we need to end up undoing this in the future.

src/components/tables/PrefixAlerts/index.tsx
```diff
@@ -65,6 +65,7 @@ function PrefixAlertTable() {
 
     return (
         <TableHydrator
+            disableQueryParamHack
```
travjenkins, line 68: This table does open with navigation but still got messed up with the hack so just switching off.

src/components/tables/StorageMappings/index.tsx
```diff
@@ -43,6 +43,7 @@ function StorageMappingsTable({ header }: Props) {
     return (
         <Box>
             <TableHydrator
+                disableQueryParamHack
```
travjenkins, line 46: This table does open with navigation but still got messed up with the hack so just switching off.

src/components/tables/cells/stats/Bytes.tsx
```diff
@@ -15,10 +16,12 @@ const Bytes = ({ read, val }: Props) => {
     const statsLoading = val === null;
     const defaultedVal = val ?? 0;
     const number = useMemo(() => formatBytes(defaultedVal), [defaultedVal]);
+    const hideColumnsSx = useHideStatsColumnsSx();
```
travjenkins, line 19: To hide the columns we need to fetch the styling from the hook.

src/components/tables/hooks/useHideStatsColumnsSx.ts
```diff
@@ -0,0 +1,20 @@
+import { useTenantDetails } from 'context/fetcher/Tenant';
+import { useMemo } from 'react';
+import { SxProps, Theme } from '@mui/material';
+
+function useHideStatsColumnsSxSx() {
```
travjenkins, line 5: Made a hook so that we didn't need to keep duplicating fetching the `hasTenants` flag.

travjenkins: Also this is not around the `theme` stuff cause this is really a one off (for right now)

src/forms/renderers/MissingType.tsx
```diff
@@ -31,7 +31,7 @@ const MissingTypeRenderer = (props: LayoutProps) => {
     return (
         <Hidden xsUp={!visible}>
             <Alert severity="error">
-                Error: Invalid field definition at
+                <span>Error: Invalid field definition at</span>
```
travjenkins, line 34: An eslint rule to handle the translate issues called this out and it seemed fair.

src/hooks/gatewayAuthToken/cache.ts
```diff
@@ -0,0 +1,27 @@
+import { LRUMap } from 'mnemonist';
+import { GatewayAuthTokenResponse } from 'types';
+
+// The limit of 10 is totally arbitrary. Felt like an okay number so all 3 tables could have
+//  something in cache and allow someone some flexibility to search
+const gatewayTokenCache = new LRUMap<string, GatewayAuthTokenResponse>(10);
```
travjenkins, line 6: Moved the cache from local storage into LRU to make it a bit easier to work with. Also, we were always fetching new ones so it isn't like we were even using the local storage values after reloads anymore.

src/hooks/gatewayAuthToken/useGatewayAuthToken.ts
```diff
@@ -12,20 +12,45 @@ import {
...
+    let tokenExpired = true;
+    if (jwt?.exp) {
+        tokenExpired = isBefore(jwt.exp * 1000, Date.now());
+    }
+
+    // If the token is still good and it contains a URL (mainly checking b/c typescript) then just
+    //  used the cached value.
+    if (!tokenExpired && gatewayConfig?.gateway_url) {
+        return Promise.resolve([{ ...gatewayConfig, cached: true }]);
+    }
```
travjenkins, line 52: Down below we pretty much always end up calling the swr call to fetch this. So keeping the caching approach in the fetcher made it easier to do.

src/hooks/gatewayAuthToken/useGatewayAuthToken.ts
```diff
@@ -71,33 +102,22 @@ const useGatewayAuthToken = (prefixes: string[] | null) => {
...
-
-    // Check if the Gateway token is expired and we need to get a new one
-    let tokenExpired = true;
-    if (jwt?.exp) {
-        tokenExpired = isBefore(jwt.exp * 1000, Date.now());
-    }
-
     const { data, mutate } = useSWR(
-        tokenExpired || authorized_prefixes.length > 0
+        hasAuthorizedPrefixes && cacheKey
```
travjenkins, line 106: We can now just call as long as we have prefixes to fetch and a key to save it as. Mainly checking both to just be super safe even though if we have prefixes then we should have a cache key

src/hooks/gatewayAuthToken/useGatewayAuthToken.ts
```diff
@@ -71,33 +102,22 @@ const useGatewayAuthToken = (prefixes: string[] | null) => {
...
+                  cacheKey,
               ]
             : null,
         gatewayFetcher,
         {
             onSuccess: ([config]) => {
-                storeGatewayAuthConfig(config);
+                if (config.cached) {
+                    return;
+                }
```
travjenkins, line 119: If we served the cached value then we don't need to go save it again

src/hooks/login/useLoginStateHandler.ts
```diff
@@ -1,14 +1,13 @@
...
 import { useCallback, useMemo, useState } from 'react';
-import { useEffectOnce, useLocalStorage } from 'react-use';
-import { LocalStorageKeys } from 'utils/localStorage-utils';
+import { useEffectOnce } from 'react-use';
 
 function useLoginStateHandler(showRegistration?: boolean) {
     const grantToken = useGlobalSearchParams(GlobalSearchParams.GRANT_TOKEN);
-    const { 2: clearGatewayConfig } = useLocalStorage(LocalStorageKeys.GATEWAY);
-    useEffectOnce(() => clearGatewayConfig());
+    useEffectOnce(() => clearGatewayAuthTokenCache());
```
travjenkins, line 10: Need to use the new cache and clear it. This is kind of overkill but it probably best.

src/stores/Tables/Hydrator.tsx
```diff
@@ -51,6 +55,17 @@ export const TableHydrator = ({
...
+            // THIS IS GROSS!
+            // Our table queries update the URL and use a library to help manage that. This library
+            //  ties in with the react router which means there will be an extra render. So on most table
+            //  we do not want this and can just skip one of the updates.
+            // TODO (query param rendering)
+            // Read More: https://github.com/pbeshai/use-query-params/issues/160
+            if (skipFirstHydration.current) {
+                skipFirstHydration.current = false;
+                return;
+            }
```
travjenkins, line 67: :-1: :-1: :-1: :-1: :-1: :-1: 

This is annoying but it was the quickest solution. The extra renders caused by the query param stuff ends up causing double calls on both live specs and stats - which is worse than the hack so chose the lesser of two evils.

src/utils/localStorage-utils.ts
travjenkins, line 1: All of this moved to LRU so does not nee to be part of the local storage.

src/components/collection/DataPreview/ListView.tsx
travjenkins, line 1: This'll be all over - switching to the new type name.

src/components/collection/DataPreview/index.tsx
travjenkins, line 1: Getting the spec from the store we're not hydrating with all the details needed.

src/components/editor/Store/LiveSpecsHydrator.tsx
travjenkins, line 1: Updating the hydrator to use a new hook that will fetch everything we need in details all at once.

travjenkins, line 1: Use the new hook to hydrate

src/components/shared/Entity/Details/Overview/index.tsx
travjenkins, line 1: No longer need to fetch live spec and can use the one from hydration.

We could read the name from here but that can be handled later as a lot of these components require the name and don't handle that being async.

src/components/shared/Entity/Details/index.tsx
travjenkins, line 1: Just updating the param

src/hooks/gatewayAuthToken/useGatewayAuthToken.ts
```diff
@@ -40,7 +68,14 @@ export const gatewayFetcher = ({
...
         sessionKey
-    );
+    ).then((response: any) => {
+        return [
+            {
+                ...response[0],
+                cacheKey,
+            },
+        ];
+    });
```
travjenkins, line 78: Need to mess with the response so the key is injected.

src/hooks/useLiveSpecs.ts
```diff
@@ -45,64 +44,49 @@ function useLiveSpecs(specType?: Entity, matchName?: string) {
...
+            required?: string[];
+        };
+        key: string[];
+    };
+    'writes_to': string[] | null;
+    'created_at': string;
+    'connectorName:connector_title->>en-US::text': string;
+    'connector_tag_documentation_url': string;
+    'reads_from': string[] | null;
+}
```
travjenkins, line 61: Created a proper type that includes everything we're querying for instead of just using the `Schema` fallback to allow us to fetch anything off this interface.

src/hooks/useLiveSpecs.ts
```diff
@@ -45,64 +44,49 @@ function useLiveSpecs(specType?: Entity, matchName?: string) {
...
+    .concat([
+        'id',
+        'last_pub_id',
+        'writes_to',
+        'updated_at',
+        'created_at',
+        'connectorName:connector_title->>en-US::text',
+        'connector_tag_documentation_url',
+        'reads_from',
+        'spec',
```
travjenkins, line 72: This list is a combination of the columns from the old `_details` query and `_spec` query.

kiahna-tucker (approved):

PR #1161, opened 2024-07-12
PROD: Reduce how often we get `exact: count` for live spec tables

travjenkins:
Today we are _always_ fetching the exact count and this can cause extra load on the DB. This should help reduce that a little bit. The amount it helps depends on how often users are actually using pagination.


kiahna-tucker (approved):

PR #1162, opened 2024-07-12
Fixing binding collection editing

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1166

### 1166

- Add a initialization done flag so we know to not try to render until ready
- Ensure the draft initializer fires a mutate so that the parent hydrator knows to fetch the recently made draft.
- Ensure the catalog name matches when starting up editing

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [x] Captures
-   [x] Collections
-   [ ] HomePage
-   [ ] Login
-   [x] Materialization

## Screenshots

### DIFF
--- a/src/components/collection/schema/Editor/index.tsx
+++ b/src/components/collection/schema/Editor/index.tsx
@@ -78,21 +78,22 @@ function CollectionSchemaEditor({ entityName, localZustandScope }: Props) {
         setCollectionData,
     ]);
 
+    // If the schema is updated via the scheme inference
+    //  of CLI button we want to fire mutate and make sure we get the latest
     useUpdateEffect(() => {
-        console.log('sup', {
-            schemaUpdated,
-            collectionInitializationDone,
-            mutate,
-        });
-        // If the schema is updated via the scheme inference
-        //  of CLI button we want to fire mutate and make sure we get the latest
-        // If the collection was initialized we need to fire mutate so that the collection
-        //  spec editor can refresh itself and know about the new draft we just created
-        //  for the collection
-        if (mutate && (schemaUpdated || collectionInitializationDone)) {
+        if (mutate && schemaUpdated) {
             void mutate();
         }
-    }, [mutate, schemaUpdated, collectionInitializationDone]);
+    }, [schemaUpdated]);
+
+    // If the collection was initialized we need to fire mutate so that the collection
+    //  spec editor can refresh itself and know about the new draft we just created
+    //  for the collection
+    useUpdateEffect(() => {
+        if (mutate && collectionInitializationDone) {
+            void mutate();
+        }
+    }, [mutate, collectionInitializationDone]);
 
     const onKeyChange = useCallback(
         async (_event, keys) => {
--- a/src/components/collection/schema/Editor/index.tsx
+++ b/src/components/collection/schema/Editor/index.tsx
@@ -1,4 +1,5 @@
 import { Grid, Stack, Typography } from '@mui/material';
+import { useBindingsEditorStore } from 'components/editor/Bindings/Store/create';
 import {
     useBindingsEditorStore_editModeEnabled,
     useBindingsEditorStore_inferSchemaResponseDoneProcessing,
@@ -38,6 +39,9 @@ function CollectionSchemaEditor({ entityName, localZustandScope }: Props) {
 
     // We need to know the schema was updated so we can "reload" this section
     const schemaUpdated = useBindingsEditorStore_schemaUpdated();
+    const collectionInitializationDone = useBindingsEditorStore(
+        (state) => state.collectionInitializationDone
+    );
 
     const setCollectionData = useBindingsEditorStore_setCollectionData();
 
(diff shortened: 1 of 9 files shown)


src/components/collection/schema/Editor/index.tsx
```diff
@@ -74,14 +78,23 @@ function CollectionSchemaEditor({ entityName, localZustandScope }: Props) {
...
     }, [schemaUpdated]);
 
+    // If the collection was initialized we need to fire mutate so that the collection
+    //  spec editor can refresh itself and know about the new draft we just created
+    //  for the collection
+    useUpdateEffect(() => {
+        if (mutate && collectionInitializationDone) {
+            void mutate();
+        }
+    }, [mutate, collectionInitializationDone]);
```
travjenkins, line 96: This allows the proper collection schema to be displayed when the `Collection` tab is open and the binding is changed.

src/components/editor/Bindings/Store/create.ts
```diff
@@ -227,6 +229,16 @@ const getInitialState = (
...
 
+    setCollectionInitializationDone: (value) => {
+        set(
+            produce((state: BindingsEditorState) => {
+                state.collectionInitializationDone = value;
+            }),
+            false,
+            'Collection Initialization Done Set'
+        );
+    },
```
travjenkins, line 240: This helps us know when the draft editor for bindings needs to be refreshed.

src/components/editor/Bindings/Store/hooks.ts
```diff
@@ -81,10 +81,6 @@ export const useBindingsEditorStore_schemaUpdated = () => {
     return useBindingsEditorStore((state) => state.schemaUpdated);
 };
 
-export const useBindingsEditorStore_setSchemaUpdated = () => {
-    return useBindingsEditorStore((state) => state.setSchemaUpdated);
-};
-
```
travjenkins, line 87: This was not being used so removed.

src/components/editor/Store/hooks.ts
travjenkins, line 1: Feels like we should be using shallow here... especially for isValidating since the entire query object can be replaced and if the value does not change it should not matter.

src/hooks/useDraftSpecEditor.ts
```diff
@@ -128,12 +129,6 @@ function useDraftSpecEditor(
         }
     }, [currentCatalog, draftSpecs, entityName, setSpecs]);
 
-    useEffect(() => {
-        if (currentCatalog && !isEqual(draftSpecRef.current, currentCatalog)) {
-            setDraftSpec(currentCatalog);
-        }
-    }, [currentCatalog]);
```
travjenkins, line 135: I am still testing this out. I am 99.99% sure we are good to remove this now that we are firing a mutate. Since this was kind of a janky way to get the draft spec updated anyway.

src/hooks/useInitializeCollectionDraft.ts
```diff
@@ -119,8 +123,14 @@ function useInitializeCollectionDraft() {
                         'workflows.collectionSelector.error.message.draftCreationFailed',
                 });
             }
+
+            setCollectionInitializationDone(true);
```
travjenkins, line 127: This should cover setting to `true` when either we made a new draft or are needing to use a live spec.

travjenkins: Even though I think we don't really _need_ to set to true when using a liveSpec it felt like it was in the spirit of this flag

image

src/hooks/useInitializeCollectionDraft.ts
```diff
@@ -107,20 +111,27 @@ function useInitializeCollectionDraft() {
...
                 });
+
+                setCollectionInitializationDone(true);
             } else {
                 updateBindingsEditorState({
                     spec: liveSpec,
                     belongsToDraft: false,
                 });
 
+                setCollectionInitializationDone(false);
```
travjenkins, line 122: Considered just setting this from within the alert setting but originally I was setting this to `true` even when we got an alert and nothing bad happened.

Gonna leave it here to give more flexibility.

src/hooks/useInitializeCollectionDraft.ts
```diff
@@ -107,20 +111,27 @@ function useInitializeCollectionDraft() {
                     spec: newDraftSpecResponse.data[0].spec,
                     belongsToDraft: true,
                 });
+
+                setCollectionInitializationDone(true);
```
travjenkins, line 115: We created a new draft so need to trigger this.

src/hooks/useDraftSpecEditor.ts
```diff
@@ -128,35 +131,16 @@ function useDraftSpecEditor(
...
     }, [currentCatalog, draftSpecs, entityName, setSpecs]);
 
+    // This for advanced spec editor
     useEffect(() => {
-        if (currentCatalog && !isEqual(draftSpecRef.current, currentCatalog)) {
+        if (
+            monitorCurrentCatalog &&
+            currentCatalog &&
+            !isEqual(draftSpecRef.current, currentCatalog)
+        ) {
```
travjenkins, line 140: Initially I removed this effect and the binding editor worked fine. However, this is how the Advanced Schema editor works. So added it back in with a flag. That way we are not constantly running updates that are just not needed.

kiahna-tucker (approved):

PR #1164, opened 2024-07-16
Sort binding source collections alphabetically

kiahna-tucker:
## Changes

* Sort bindings listed in the binding selector alphabetically by source collection name.

### Manually tested

* Initialization of edit workflow for a materialization with multiple bindings, all of which are enabled.

* Initialization of edit workflow for a materialization with multiple bindings, some of which are disabled.

* A series of binding interactions that would prompt reordering.

Discovery of more than one binding is the only standard scenario that is untested.

### Automated tests

* N/A

All tests were run 36 times with six tests intermittently failing. None of the tests should be impacted by the code adjustment so I am interpreting the failures as false positives.

-   [x] Admin
-   [x] Captures
-   [x] Collections
-   [x] HomePage
-   [x] Login
-   [x] Materialization

### DIFF
--- a/src/stores/Binding/Store.ts
+++ b/src/stores/Binding/Store.ts
@@ -15,7 +15,6 @@ import {
     isEqual,
     omit,
     omitBy,
-    orderBy,
     pick,
     union,
 } from 'lodash';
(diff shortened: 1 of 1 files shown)


travjenkins (approved):
lgtm

PR #1171, opened 2024-07-18
add initial CSP config

skord:
## Issues

https://github.com/estuary/security/issues/100

## Changes

-   This adds an initial CSP configuration to only allow HTTPS connections. We'll narrow it down later.


travjenkins (approved):
lgtm... and annoyed with myself that I did not initially include this :D

Issue #1156, opened 2024-07-09
Reduce impact of `estuary_support` when loading UI


travjenkins:
We are currently loading the tenants way too much in the UI for Estuary folks. The approach was made for people without access to thousands of tenants. We need to change how we load tenants.

Some specific checks we need to know
1. has at least 1 tenant
2. has access to `estuary_support`
3. has access to `demo` tenant

After that I think we can just start loading in tenants when they are actually needed or async in the background or maybe both.

Also - these should be persisted by Zustand in storage to reduce the impact.

travjenkins:
We will be doing this with a new view that is getting created on the backend.

travjenkins:
Phil is current working on a new SQL Function that will return the special cases we needed handled. That way we do not need to query the `tenant` table at all unless we are looking for specific settings about a tenant.

PR #1175, opened 2024-07-19
Use new endpoint to fetch details about user's access

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1156

### 1156

- Wired up new RPC for `user_info_summary`
- No longer fetch payment details for `estuary_support` roles since they never saw payment notifications
- Update all over to use the new flags
- Call mutate when one of the settings could change (like access was removed/added)
- Create a new list that acts more like the old tenant list.
- Cleaned up old calls and fetchers that were no longer needed / not being used anyway

### Misc
- Update the `hidePreview` check to handle when the user does not have access to the tenant and return a `maybeSingle` response.
- 

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots


src/components/admin/Tabs.tsx
```diff
@@ -9,7 +9,7 @@ function AdminTabs() {
     const intl = useIntl();
     const { pathname } = useLocation();
     const [selectedTab, setSelectedTab] = useState(0);
-    const { hasTenants } = useTenantDetails();
+    const hasAnyAccess = useUserInfoSummaryStore((state) => state.hasAnyAccess);
```
travjenkins, line 12: This change will be all over. Anywhere we were checking `hasTenants` should now be `hasAnyAccess`. Please keep an eye and make sure if they are checking positive or negative and that was not lost with changes.

src/components/hero/AcceptDemoInvitation.tsx
```diff
@@ -66,6 +71,7 @@ function AcceptDemoInvitation({
                             void mutateAuthRoles();
                         }
 
+                        setHasDemoAccess(true);
```
travjenkins, line 74: Instead of firing off a full mutate decided just to update the store here. I _might_ like this plan in the future and might remove it. For now I think it is best to reduce calls.

src/components/hero/DemoButton.tsx
```diff
@@ -34,7 +32,9 @@ function DemoButton({ step, type }: Props) {
 
     const objectRoles = useEntitiesStore_capabilities_adminable();
 
-    const demoAccessExists = useEntitiesStore_hasDemoTenantAccess();
+    const demoAccessExists = useUserInfoSummaryStore(
+        (state) => state.hasDemoAccess
+    );
```
travjenkins, line 37: Another repeated change. `demoAccessExists` should now fetch from the user info summary store.

src/components/shared/TenantSelector.tsx
```diff
@@ -17,7 +17,8 @@ interface Props {
 function TenantSelector({ updateStoreState }: Props) {
     const intl = useIntl();
 
-    const { tenants } = useTenantDetails();
+    const tenantNames = useEntitiesStore_tenantsWithAdmin();
```
travjenkins, line 20: This is a new hook that is used in just this spot. Since this for sure should not show nested properties we needed a new list.

src/components/tables/RowActions/AccessGrants/RevokeGrant.tsx
```diff
@@ -78,8 +81,11 @@ function RevokeGrant({
     useEffect(() => {
         if (progress === ProgressStates.SUCCESS) {
             incrementSuccessfulTransformations();
+            if (mutate) {
+                void mutate();
+            }
```
travjenkins, line 86: This can end up making multiple needless calls but the quickest way to handle this right now.

src/context/Authenticated.tsx
```diff
@@ -19,13 +19,13 @@ export function AuthenticatedOnlyContext({ children, hideSpinner }: Props) {
             <OnLoadSpinnerProvider defaultState={!hideSpinner}>
                 <QueryParamProvider>
                     <ZustandProvider>
-                        <AppGuards>
-                            <PreFetchDataProvider>
+                        <UserInfoSummaryStoreProvider>
```
travjenkins, line 22: We want to fetch this **ASAP** so we can use the details in anything else. So this will pretty much always stay super high.

src/context/AuthenticatedHydrators.tsx
```diff
@@ -1,8 +1,15 @@
 import { EntitiesHydrator } from 'stores/Entities/Hydrator';
 import { BaseComponentProps } from 'types';
+import { TenantBillingDetailsContextProvider } from './fetcher/Tenant';
 
 function AuthenticatedHydrators({ children }: BaseComponentProps) {
-    return <EntitiesHydrator>{children}</EntitiesHydrator>;
+    return (
+        <EntitiesHydrator>
+            <TenantBillingDetailsContextProvider>
```
travjenkins, line 8: Now we have something to fetch just the payment details of tenants the user has admin access to.

src/context/fetcher/GrantDetails.tsx
travjenkins, line 1: I think this was just left over. This context was never consumed so it was just making an empty call.

src/context/fetcher/index.tsx
travjenkins, line 1: Since now the tenant details are fetched for payment stuff and that was needed earlier that was pulled out. Then, the grant details stuff wasn't even being used... so this should be fine to be removed.

src/directives/AcceptGrant.tsx
```diff
@@ -61,6 +65,20 @@ function AcceptGrant({
...
+                                trackEvent(
+                                    `${directiveName}:UserInfoSummary:Mutate:Success`
+                                );
+                            })
+                            .catch(() => {
+                                trackEvent(
+                                    `${directiveName}:UserInfoSummary:Mutate:Error`
+                                );
+                            });
+                    }
```
travjenkins, line 80: This is kinda weird and one of those places where I am wondering if we should just throw up a FullPageError... cause if we cannot get this call to work we are in big trouble.

src/hooks/billing/useTenantMissingPaymentMethodWarning.tsx
```diff
@@ -37,8 +37,10 @@ function useTenantMissingPaymentMethodWarning() {
         notificationStoreSelectors.hideNotification
     );
 
-    const { tenants } = useTenantDetails();
-    const hasSupportRole = useEntitiesStore_hasSupportRole();
+    const { tenantBillingDetails } = useTenantBillingDetails();
+    const hasSupportRole = useUserInfoSummaryStore(
+        (state) => state.hasSupportAccess
+    );
```
travjenkins, line 43: This is the only thing that _actually_ needs to get stuff from the tenant table. So it is the main consumer here.

This is a bit overkill to check the support role here again... but to keep things aligning more 1 to 1 I just left it.

src/hooks/useTenants.ts
```diff
@@ -42,10 +41,26 @@ export function useTenantHidesDataPreview(entityName: string) {
...
+        }
+
+        // Null means the call is done but the user does not have access
+        //  to see the tenant details
+        if (data === null) {
+            return true;
+        }
+
+        // We still don't have a good enough response so default to null
+        return null;
```
travjenkins, line 62: The ternary was going to have several nested and decided to break this up a bit so commenting was easier

src/hooks/useUserGrants.ts
travjenkins, line 1: Part of removing the fetcher.

src/context/UserInfoSummary/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    useEffect(() => {
+        if (data) {
+            populateAll(data);
+        }
+    }, [data, populateAll]);
+
+    useEffect(() => {
+        setMutate(mutate);
+    }, [mutate, setMutate]);
```
travjenkins, line 23: Kept these separate for now mainly because I am getting a bit suspicious of storing things like mutate inside the store.

src/types/index.ts
```diff
@@ -167,6 +167,13 @@ export interface Tenants {
     gcm_account_id?: string | null;
 }
 
+export interface TenantPaymentDetails {
+    gcm_account_id: string | null;
+    payment_provider: TenantPaymentProviders;
+    tenant: string;
+    trial_start: string;
+}
```
travjenkins, line 175: We don't need all the details so saving a tiny bit of bits by reducing the query.

src/stores/Entities/hooks.ts
```diff
@@ -41,20 +41,15 @@ export const useEntitiesStore_capabilities_adminable = () => {
...
-        useShallow(() => {
-            return readable.includes(DEMO_TENANT);
-        })
+        useShallow((state) => [
+            ...new Set(
+                Object.keys(state.capabilities.admin).map((tenant) =>
+                    stripPathing(tenant, true)
+                )
+            ),
+        ])
```
travjenkins, line 52: This is still somewhat in progress.... but need a list of unique entities that are just the "tenant" portion.

src/services/supabase.ts
travjenkins, line 1: Added and alpha sorted

src/hooks/useUserInfoSummary.ts
```diff
@@ -0,0 +1,8 @@
+import { useQuery } from '@supabase-cache-helpers/postgrest-swr';
+import { getUserInfoSummary } from 'api/userInfoSummary';
+
+function useUserInfoSummary() {
+    return useQuery(getUserInfoSummary());
+}
```
travjenkins, line 6: This is the "wrapper" hook that made we wonder if it even adds value.

src/context/fetcher/TenantBillingDetails.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    // TODO (payment method notification)
+    // This does work a bit off of a side effect but that is fine for now.
+    //  Eventually I'd like to move this provider logic into the payment method
+    //  warning handled itself. As it is really the thing that care about these values.
+
+    // We suppress payment warnings for support role anyway so don't bother fetching for them
+    const { tenants, error, isValidating } = useTenantsDetailsForPayment(
+        hasSupportRole ? [] : tenantsWithAdmin.slice(0, MAX_TENANTS)
+    );
```
travjenkins, line 34: Initially I thought about moving some of these hook calls into the payment fetch hook but decided against it. I could easily see needing that hook to look up payment details on a specific tenant to open a modal or something. So leaving out here for now.

src/app/guards/TenantGuard.tsx
```diff
@@ -21,29 +19,10 @@ function TenantGuard({ children }: BaseComponentProps) {
         [showTenantCreation]
     );
 
-    const {
-        userGrants,
-        isValidating: checkingGrants,
-        mutate,
-        error,
-    } = useUserGrants(true);
+    const hasAnyAccess = useUserInfoSummaryStore((state) => state.hasAnyAccess);
```
kiahna-tucker, line 22: `hasAccess` or, preferably, `accessExists` would be a better name for this state property IMO.

kiahna-tucker (approved):
Approved without manual testing in the interest of time.

Issue #1186, opened 2024-07-31
Disable include action for forbidden and unsatisfiable fields


travjenkins:
Fields with the constraint type `FIELD_FORBIDDEN` or `UNSATISFIABLE` should have a disabled _Include_ row action in the field selection table.

PR #1187, opened 2024-08-01
Disable include action for forbidden and unsatisfiable fields

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1186

### 1186

* Disable the _Include_ field action for forbidden and unsatisfiable fields.

* Mark forbidden fields as being excluded by default when the draft is evaluated in the UI (i.e., `mapConstraintsToProjections` in _src/component/editor/Bindings/FieldSelections/index.tsx_).

### Manually tested

* Validate that the _Include_ field action is always disabled for forbidden fields. `estuary/stripe/materialize-postgres` is a materialization in production that has at least one forbidden field.

### Automated tests

N/A

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

**Forbidden field row**

pr_screenshot-1169-forbidden_field-rec_flag_true

### DIFF
--- a/src/components/editor/Bindings/FieldSelection/index.tsx
+++ b/src/components/editor/Bindings/FieldSelection/index.tsx
@@ -42,6 +42,7 @@ import { FormStatus } from 'stores/FormState/types';
 import { Schema } from 'types';
 import { BooleanParam, useQueryParam } from 'use-query-params';
 import {
+    evaluateRequiredExcludedFields,
     evaluateRequiredIncludedFields,
     getBindingIndex,
 } from 'utils/workflow-utils';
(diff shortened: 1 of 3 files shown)


src/components/editor/Bindings/FieldSelection/index.tsx
```diff
@@ -89,7 +90,15 @@ const mapConstraintsToProjections = (
                     constraint.type
                 );
 
-                selectionType = includeRequired ? 'include' : null;
+                const excludeRequired = evaluateRequiredExcludedFields(
+                    constraint.type
+                );
```
travjenkins, line 95: Can you make this function call down below in the ternary? That way we can skip if not needed.

kiahna-tucker: No, because it is used to set the `disabled` property of the _Include_ button as well as within the click handler for the _Exclude_ button.

travjenkins (commented):
1 small change

travjenkins (approved):
lgtm

PR #1209, opened 2024-08-09
OPS - enable CSP on dashboard

travjenkins:
Disabled CSP changes just to be safe as it was tested in the middle of the night.

This will enable it with the settings that seemed to work best.


skord (approved):
LGTM!

Issue #1200, opened 2024-08-08
Materializing capture to existing disabled materialization fails


travjenkins:
**Bug**
Create a materialization
Disable it
Select a capture from entity table
Click materialize
Select disabled materialization

See the bindings are temporarily updated but flash back to original

**Expected**
The new bindings of the capture are added

**Screenshots**
On load the bindings appear to only have the original bindings
![image](https://github.com/user-attachments/assets/c7cda49a-9a72-4c72-aa6c-a646ca57c341)

**Notes**
This is totally due to the auto enabling of the entity while we are also trying to add the bindings. Need to only auto enable after the bindings are added. This is not necessarily going to be too easy. We could easily add some new state props to handle this... but also might just want to take a look at moving more calls "up" higher in the order.

PR #1215, opened 2024-08-12
Force enable materializations during initialization to allow prefill bindings to work

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1200

### 1200

- Have the draft init code handle enabling disabled materializations
- Edit now knows that we are force enabling right away instead of setting after all of edit is loaded

### Misc

- Add encoding to search param handler for booleans

### Manually tested

- Editing [`enabled`,`disabled`] materialization
- Materializing to a [`enabled`,`disabled`] materialization

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Editing a disabled materialization
image

Materializing to disabled materialization
image

### DIFF
--- a/src/components/editor/Bindings/FieldSelection/useBackgroundTest.ts
+++ b/src/components/editor/Bindings/FieldSelection/useBackgroundTest.ts
@@ -4,7 +4,6 @@ import {
 } from 'components/editor/Store/hooks';
 import { useEntityType } from 'context/EntityContext';
 import { useEntityWorkflow_Editing } from 'context/Workflow';
-import { GlobalSearchParams } from 'hooks/searchParams/useGlobalSearchParams';
 import { useEffect, useRef, useState } from 'react';
 import { logRocketEvent } from 'services/shared';
 import { CustomEvents } from 'services/types';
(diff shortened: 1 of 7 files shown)


src/components/editor/Bindings/FieldSelection/useBackgroundTest.ts
travjenkins, line 1: The background test does not need to auto enable now and that will be up in the draft initializer.

src/components/materialization/useGenerateCatalog.ts
travjenkins, line 1: No longer need to have generate catalog toggle disable.

src/utils/misc-utils.ts
travjenkins, line 1: This is a simple solution for handling encoding search params. This ends up only ever being used in the `create` and `edit` navigation code.

src/hooks/searchParams/useSearchParamAppend.ts
travjenkins, line 1: Instead of handling this within the options itself - decided to allow this handled to encode booleans.

kiahna-tucker (approved):

Issue #1165, opened 2024-07-16
Add a Dashboard page


travjenkins:
**Last Updated:** Aug. 26, 2024

<hr />

Create the foundation for a dashboard on the _Welcome_ page.

Below are the base requirements for this feature:

* Allow the welcome section on the _Welcome_ page to be hidden.

* Create a simple dashboard which consists of a tenant selector and three cards that surface relevant, entity-specific statistics. Relevant statistics are as follows:
    * Active entity count.
   
    * Data usage for the current month (which is only applicable to captures and materializations).

## Design

**Last Updated:** Aug. 26, 2024

**Welcome section | Open**

<img width="521" alt="ui_issue-1165-dashboard_init-figma-welcome_open" src="https://github.com/user-attachments/assets/566f279b-317b-47d3-853f-bccc474d647e">

<br />
<br />

**Welcome section | Collapsed**

<img width="525" alt="ui_issue-1165-dashboard_init-figma-welcome_collapsed" src="https://github.com/user-attachments/assets/6922161c-8d1f-4634-b68b-8281e6c6ce13">

<br />
<br />

**Entity overview cards**

<img width="587" alt="ui_issue-1165-dashboard_init-figma-entity_overview_cards" src="https://github.com/user-attachments/assets/a46c2264-379d-45ff-9d94-f0a622c6127f">

<br />
<br />

**NOTE:** The page container in the client is styled as a card itself, therefore it was decided to add a border and background to the welcome section to distinguish it from dashboard elements.

## References

**Design**

* [Welcome Page Redesign | Latest.](https://www.figma.com/design/V98PvMk6l0a9CNVrTraB0W/Estuary?node-id=0-1&t=X1Tjm6NF2RCn2uPu-0) A set of mock ups depicting variations of the _Welcome_ page that includes a collapsible welcome section as well as a basic dashboard. All frames highlighting the _Welcome_ page are relevant to this issue.

* [Welcome Page Designs | 2024-08-20.](https://www.figma.com/design/veWI8AVomUD9OypCr86faq/Estuary-app-(Welcome-page)?node-id=2165-3438&t=Ra6ek5HEwyctegrG-0) A set of mock ups depicting variations of the _Welcome_ page that includes options for the content and layout of the welcome section as well as a basic dashboard. The frames relevant to this issue are titled `WElcome page`.

travjenkins:
[...]

PR #1217, opened 2024-08-14
Create the foundation for a dashboard on the welcome page

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1165

### 1165

* Create a foundation for the dashboard which will be added to the existing _Welcome_ page.

* Implement the entity statistic overview cards as shown on the latest set of mockups.

### Manually tested

* Validate that the welcome section is displayed by default.

* Validate that the welcome section can be closed and reopened.

* Validate that the entity statistic cards display data relevant to the selected tenant.

### Automated tests

N/A

-   [x] Admin
-   [x] Captures
-   [x] Collections
-   [x] HomePage
-   [x] Login
-   [ ] Materialization

### Welcome Section

**Section hidden | Light mode**

pr_screenshot-1217-dashboard_init-welcome_collapsed

<br />
<br />

**Section displayed | Light mode**

pr_screenshot-1217-dashboard_init-welcome_open

<br />
<br />

**Section hidden | Dark mode**

pr_screenshot-1217-dashboard_init-dark-welcome_collapsed

<br />
<br />

**Section displayed | Dark mode**

pr_screenshot-1217-dashboard_init-dark-welcome_open

<br />
<br />

### Dashboard

**Loading state**

pr_screenshot-1217-dashboard_init-all_failed

<br />
<br />

**Unable to fetch the monthly data usage**

pr_screenshot-1217-dashboard_init-monthly_usage_failed

<br />
<br />

**Unable to fetch the number of active entities**

pr_screenshot-1217-dashboard_init-dark-active_count_failed

<br />
<br />

**NOTE:** The monthly usage statistic uses the same tooltips as the statistic cells of the entity table. A tooltip reading `${byte_usage} Bytes Written` (for captures) or `${byte_usage} Bytes Read` (for materializations) is shown when the call succeeds; a tooltip reading `Failed to fetch stats.` is shown when the call fails.


src/api/stats.ts
```diff
@@ -269,6 +274,45 @@ const getStatsForDetails = (
...
+            break;
+        case 'materialization':
+            query = MATERIALIZATION_QUERY;
+            break;
+        case 'collection':
+            query = COLLECTION_QUERY;
+            break;
+        default:
+            query = DEFAULT_QUERY;
+    }
```
travjenkins, line 303: Is there a specific future plan you have for supporting these queries? 

If this is just adding to handle it _in case_ I think we can comment out so reduce complexity for queries that aren't actually getting used.

src/components/dashboard/EntityStatOverview/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    if (entityType === 'collection') {
+        return 'terms.collections';
+    }
+
+    if (entityType === 'capture') {
+        return 'terms.sources';
+    }
+
+    return 'terms.destinations';
+};
```
travjenkins, line 40: Don't think we need to change anything... but this initially just _feels_ odd to me but  I do not know why. I'll process it a but more to try to figure out why exactly.

src/components/dashboard/EntityStatOverview/useActiveEntityCount.ts
```diff
@@ -0,0 +1,28 @@
...
+        }
+    );
+
+    const count = useMemo(
+        () =>
+            data ? data.filter(({ disable }) => disable !== true).length : 0,
+        [data]
+    );
+
+    return { count, error, isValidating, loading: !isArray(data) };
```
travjenkins, line 27: adding a manual `loading` is okay... but does give me a bit of pause since the new SWR also handles validating vs loading and might get confusing.

src/components/dashboard/EntityStatOverview/useActiveEntityCount.ts
```diff
@@ -0,0 +1,28 @@
...
+        {
+            refreshInterval: 15000,
+        }
+    );
+
+    const count = useMemo(
+        () =>
+            data ? data.filter(({ disable }) => disable !== true).length : 0,
+        [data]
+    );
```
travjenkins, line 25: I understand why we are calculating `count` but again might end up causing some confusion since it isn't the normal count that get's returned by queries.

src/components/dashboard/EntityStatOverview/useMonthlyUsage.ts
```diff
@@ -0,0 +1,61 @@
...
+import { isArray } from 'lodash';
+import { DateTime } from 'luxon';
+import { useMemo } from 'react';
+import { useTenantStore } from 'stores/Tenant/Store';
+import { CatalogStats_Dashboard } from 'types';
+import { hasLength } from 'utils/misc-utils';
+
+const isDefaultStatistic = (
+    datum: CatalogStats_Dashboard | DefaultStatsWithDocument
+): datum is DefaultStatsWithDocument => 'bytes_read_by_me' in datum;
```
travjenkins, line 12: Can you add some comment on why this approach was taken? It was not immediately clear to me why we would check it this way and not just have the logic be more like "if capture use these variables".

src/components/dashboard/EntityStatOverview/useMonthlyUsage.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        return [dataWritten, dataRead];
+    }, [data]);
+
+    return {
+        captureUsage,
+        error,
+        isValidating,
+        loading: !isArray(data),
```
travjenkins, line 58: Same comment here with the `loading` prop

src/components/dashboard/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+                sx={{ display: 'flex', justifyContent: 'flex-end' }}
+            >
+                <Box style={{ width: 300 }}>
+                    <TenantSelector />
+                </Box>
+            </Grid>
+
+            <Grid item xs={4}>
+                <EntityStatOverview
+                    Icon={<CloudUpload fontSize={ICON_SIZE} />}
```
travjenkins, line 31: Might be over engineering... but should we make some kind of shared "config" for entities where we store off stuff like what icon is shown for them? Might be too soon since it is only used in two spots right now.

src/components/dashboard/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+                    entityType="capture"
+                    monthlyUsage={captureUsage}
+                    monthlyUsageLoading={loading}
+                />
+            </Grid>
+
+            <Grid item xs={4}>
+                <EntityStatOverview
+                    Icon={<DatabaseScript fontSize={ICON_SIZE} />}
+                    background={semiTransparentBackground_blue}
```
travjenkins, line 42: If we're already passing in the `entityType` and figuring out internals what was driving passing in these two and not just making them handled internally?

src/context/Router/index.tsx
```diff
@@ -169,7 +170,9 @@ const router = createBrowserRouter(
                         path={authenticatedRoutes.home.path}
                         element={
                             <Suspense fallback={null}>
-                                <HomePage />
+                                <DashboardWelcomeProvider>
```
travjenkins, line 173: Since the hook `useShowDashboardWelcome` I _think_ we're good to remove this to make sure the surface area of possible issues is as small as possible.

src/lang/en-US.ts
```diff
@@ -61,6 +61,8 @@ const CommonMessages: ResolvedIntlConfig['messages'] = {
     'terms.derivation': `Derivation`,
     'terms.documentation': `Docs`,
     'terms.entity': `Entity`,
+    'terms.sources': `Sources`,
+    'terms.destinations': `Destinations`,
```
travjenkins, line 65: Since these are getting added can you use this down in `routeTitle.captures` and `routeTitle.materializations`

src/lang/en-US.ts
```diff
@@ -155,6 +157,7 @@ const CTAs: ResolvedIntlConfig['messages'] = {
     'cta.back': `Back`,
     'cta.revoke': `Revoke`,
     'cta.goToDetails': `See Details`,
+    'cta.goToEntityPage': `See all`,
```
travjenkins, line 160: I think making this key name more general might be good as I could see us using `See all` in other places.

src/api/stats.ts
kiahna-tucker: The addition of the data trends graph would duplicate a considerable amount of code for the data usage graph on the _Details_ page. Consequently, it was decided that the two graphs will share logic; this opens the door for `getStatsForDashboard` and `getStatsForDetails` to be consolidated into a single function. This branch is cherry picked from one which implements the initial plan for the dashboard.

**Tl;dr:** The highlighted code fragment can be commented out for now but will likely need to be reintroduced when the data trends graph is developed.

travjenkins: Ahhhh - that makes sense. We can leave it in then.

src/components/dashboard/EntityStatOverview/useActiveEntityCount.ts
kiahna-tucker: From what I have read, there is not a way to reference the value of or extract an object property in a query formed using the Supabase JavaScript client. Would you like this variable renamed?

travjenkins: No... we can worry about it if it actually causes a problem during future development.

src/context/DashboardWelcome.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const DashboardWelcomeContext = createContext<DashboardWelcomeState | null>(
+    null
+);
+
+const DashboardWelcomeProvider = ({ children }: BaseComponentProps) => {
+    const [welcomeShown, setWelcomeShown] = useLocalStorage(
+        LocalStorageKeys.DASHBOARD_WELCOME,
+        false
+    );
```
travjenkins, line 19: Why are we making a provider for this?

src/components/dashboard/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { CloudDownload, CloudUpload, DatabaseScript } from 'iconoir-react';
+import EntityStatOverview from './EntityStatOverview';
+import useMonthlyUsage from './EntityStatOverview/useMonthlyUsage';
+
+const ICON_SIZE = 12;
+
+export default function Dashboard() {
+    const { captureUsage, isLoading, materializationUsage } = useMonthlyUsage();
+
+    return (
```
travjenkins, line 17: image

I think this section should probably handle smaller screen sizes a bit better and make the cards stack vertically.

travjenkins (commented):
some small changes needed

src/api/liveSpecsExt.ts
```diff
@@ -368,6 +369,14 @@ const getLiveSpecsByLiveSpecId = async (liveSpecId: string) => {
     return data;
 };
 
+const getLiveSpecShards = (tenant: string, entityType: Entity) => {
+    return supabaseClient
+        .from(TABLES.LIVE_SPECS_EXT)
+        .select(SHARDS_DISABLE)
+        .ilike('catalog_name', `${tenant}%`)
```
travjenkins, line 376: needs to use `escapeReservedCharacters`

src/api/stats.ts
```diff
@@ -269,6 +274,45 @@ const getStatsForDetails = (
...
+    //     case 'collection':
+    //         query = COLLECTION_QUERY;
+    //         break;
+    //     default:
+    //         query = DEFAULT_QUERY;
+    // }
+
+    return supabaseClient
+        .from(TABLES.CATALOG_STATS)
+        .select(`${DEFAULT_QUERY},flow_document`)
```
travjenkins, line 307: It looks like we only ever get taskStats... can we just have that be fetched?

`flow_document:flow_document->>taskStats::json`

src/components/dashboard/index.tsx
kiahna-tucker: This was discussed in Slack and on a VC.

src/context/DashboardWelcome.tsx
kiahna-tucker: I was unsure whether the contents of the provider would be extended. Originally, I anticipated a partial redesign of the welcome section and didn't want to strip it until I saw what was replacing the data trends chart/how Jonathan was planning to extend the dashboard.

src/components/dashboard/EntityStatOverview/StatOverview.tsx
```diff
@@ -0,0 +1,155 @@
...
+    if (entityType === 'collection') {
+        return <DatabaseScript fontSize={iconSize} />;
+    }
+
+    if (entityType === 'capture') {
+        return <CloudUpload fontSize={iconSize} />;
+    }
+
+    return <CloudDownload fontSize={iconSize} />;
+};
```
travjenkins, line 74: I think I figured out why this threw me off initially. These should really just return the actual component classes.

Then down below we use them as components

```
<Icon fontSize={iconSize} />
```

src/services/supabase.ts
```diff
@@ -25,6 +25,9 @@ export const CONNECTOR_IMAGE = `image:connector_logo_url->>en-US`;
 
 export const QUERY_PARAM_CONNECTOR_TITLE = `connector_title->>en-US`;
 
+export const SHARDS_DISABLE = `spec->shards->disable`;
+export const TASK_STATS = `task_stats:flow_document->>taskStats::json`;
```
travjenkins, line 29: Ugh... I don't know why I made this so complex... I think we can just use this since it is already stored as `json`

`task_stats:flow_document->taskStats`

src/pages/Home.tsx
```diff
@@ -1,28 +1,41 @@
...
     GlobalSearchParams,
 } from 'hooks/searchParams/useGlobalSearchParams';
 import usePageTitle from 'hooks/usePageTitle';
+import { Plus, Xmark } from 'iconoir-react';
 import { useMemo } from 'react';
 import { FormattedMessage } from 'react-intl';
 
 const Home = () => {
-    usePageTitle({ header: authenticatedRoutes.home.title });
+    const theme = useTheme();
```
travjenkins, line 25: Since we are expanding the functionality of this page... we should now toss the error message into a stand along component. Since the search param stuff usually causes extra renders.

image

src/components/dashboard/EntityStatOverview/useMonthlyUsage.ts
```diff
@@ -0,0 +1,61 @@
...
+    const endDate = DateTime.utc().startOf('month');
+
+    const query = useMemo(
+        () =>
+            hasLength(selectedTenant)
+                ? getStatsForDashboard(selectedTenant, 'monthly', endDate)
+                : null,
+        [endDate, selectedTenant]
+    );
+
```
travjenkins, line 26: I'm not sure... cause linting isn't complaining... but I should have thought that `endDate` would cause this memo to always rerun.

src/components/dashboard/EntityStatOverview/Statistic.tsx
```diff
@@ -0,0 +1,58 @@
...
+            >
+                <Typography
+                    variant="h6"
+                    sx={{ opacity: loading || error ? 0.4 : 1 }}
+                >
+                    {byteUnit
+                        ? prettyBytes(value, {
+                              minimumFractionDigits: 2,
+                              maximumFractionDigits: 2,
+                          })
```
travjenkins, line 52: Can you import the use the `formatBytes` function here

src/components/dashboard/EntityStatOverview/useMonthlyUsage.ts
travjenkins: Also - noting our convo from Slack. Looks like the values are being cached and I think we should yank that out for now.

src/components/dashboard/EntityStatOverview/Statistic.tsx
travjenkins: We should probably just rename `formatBytes` and leave the other ones. Cause they all aren't using the same min and max. It is more like the `standardByteFormat` rather than one that is to be used everywhere.

src/pages/Home.tsx
```diff
@@ -1,46 +1,76 @@
...
+                                style={{ color: theme.palette.text.primary }}
+                            />
+                        )}
+                    </IconButton>
+                </Toolbar>
+
+                <Collapse in={welcomeShown}>
+                    <HeroImageAndDescription />
+                </Collapse>
+            </Box>
```
travjenkins, line 71: Future work - should make this also a stand alone component since it is the only thing in here that cares about the `welcomeShown` flag.

src/components/home/dashboard/EntityStatOverview/useActiveEntityCount.ts
```diff
@@ -0,0 +1,24 @@
...
+import { hasLength } from 'utils/misc-utils';
+
+export default function useActiveEntityCount(entityType: Entity) {
+    const selectedTenant = useTenantStore((state) => state.selectedTenant);
+
+    const { data, error, isLoading } = useQuery(
+        hasLength(selectedTenant)
+            ? getLiveSpecShards(selectedTenant, entityType)
+            : null
+    );
```
travjenkins, line 15: Forgot... since this is fetching an untold number of live specs we should probably update the UI to have a UX for `1000+` active entities and then later on decide if we ever want to show more than that in the dashboard.

src/services/supabase.ts
```diff
@@ -25,6 +25,9 @@ export const CONNECTOR_IMAGE = `image:connector_logo_url->>en-US`;
 
 export const QUERY_PARAM_CONNECTOR_TITLE = `connector_title->>en-US`;
 
+export const SHARDS_DISABLE = `spec->shards->disable`;
+export const TASK_STATS = `task_stats:flow_document->taskStats::json`;
```
travjenkins, line 29: Since this is already json I think we're safe to remove `::json` as well

src/components/dashboard/EntityStatOverview/index.tsx
kiahna-tucker: The routing-related ambiguous discomfort was discussed on a VC and this was deemed _okay for now_.

travjenkins (approved):
lgtm

PR #1224, opened 2024-08-20
Get LogRocket listening to network calls

travjenkins:
## Issues

tech debt

### tech debt

-  Need to ensure LogRocket starts up before Supabase since the Supabase Client messes with the fetch function. This is kind of annoying because LogRocket will init asynchronously but Supabase client creation is synchronously. 

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

image


src/context/GlobalProviders.tsx
travjenkins, line 1: This is the meat and taters of the change.

src/context/GlobalProviders.tsx
```diff
@@ -0,0 +1,102 @@
...
+// Setup Monaco
+self.MonacoEnvironment = {
+    getWorker(_, label) {
+        if (label === 'json') {
+            return new JsonWorker();
+        }
+        return new EditorWorker();
+    },
+};
+loader.config({ monaco });
```
travjenkins, line 77: Went ahead and pulled everything that was previously a "global setup" that was in the main app into here so they could all be in one place.

kiahna-tucker (approved):

PR #1231, opened 2024-08-26
Break up the content file

travjenkins:
Grouping some messages to reduce number of files

## Issues

tech debt

## Changes

These are the steps I followed to reduce chance of errors

1. Use IDE to refactor every object into a stand alone file
2. Create a new folder for `en_US` and move new files there
3. Create a new `index` file to wrap up all the files
4. Fix typing error due to tons of spreads
5. Went through the files and logically grouped them to reduce the number of files. Whatever the original object name was is added as a human readable comment above where they are added.
6. Updated imports (like `Header` moving to `Navigation`)
7. Alpha sorted the main `index` object
8. Added a LR event to catch when we are missing a key

### Manually tested

- Clicked through all pages and checked for errors

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Broke a key on purpose
image

Falls back to
image

Falls back to
image


src/lang/en-US/AdminPage.ts
```diff
@@ -0,0 +1,179 @@
...
+    'admin.alerts.table.noContent.message': `To begin receiving email notifications for a prefix you admin, click "Configure Notifications" above to create a subscription.`,
+    'admin.alerts.table.label.alertMethod': `Notification Method`,
+
+    'admin.tabs.users': `Account Access`,
+    'admin.tabs.connectors': `Connectors`,
+    'admin.tabs.api': `CLI-API`,
+    'admin.tabs.billing': `Billing`,
+    'admin.tabs.settings': `Settings`,
+
+    // Storage Mappings
```
travjenkins, line 130: This is an example of how stuff was grouped with the object name made human readable as a comment.

src/lang/en-US/Captures.ts
travjenkins, line 1: To make some grouping easier it is now more based on what the things _are_ and no longer _where_ they are being used.

kiahna-tucker (approved):

src/lang/en-US/JsonForms.ts
```diff
@@ -0,0 +1,30 @@
+import { Errors } from './Errors';
+
+export const JsonForms: Record<string, string> = {
+    'jsonForms.clearInput': `Clear input field`,
+
+    // Custom Renderers
+    'dateTimePicker.button.ariaLabel': `Open date time picker for {label}`,
+    'dateTimePicker.picker.footer': `Timezone: UTC`,
+    'datePicker.button.ariaLabel': `Open date picker for {label}`,
+    'informational.sshEndpoint.title': `All Estuary traffic comes from a group of IPs that can be whitelisted:`,
```
travjenkins, line 10: Updated the text to say `group of IPs` instead of `single IP`

PR #1254, opened 2024-09-11
Enable data plane selection in workflows

kiahna-tucker:
## Issues

### issue_number

-   changes...

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots


src/api/dataPlane.ts
travjenkins, line 1: This should probably get renamed to `dataPlanes' to match the table name and also so we can mildly differentiate between the table `datePlanes` and the endpoint `data plane`

src/api/discovers.ts
```diff
@@ -1,17 +1,34 @@
...
         endpoint_config: config,
         connector_tag_id: connectorID,
         draft_id: draftID,
         update_only: updateOnly ?? false,
-    });
+    };
+
+    if (hasLength(dataPlaneName)) {
+        data.data_plane_name = dataPlaneName;
+    }
```
travjenkins, line 31: I think you can just set the value in the object directly and then if it is undefined Supabase should remove the value - not 100% sure though.

src/components/capture/useDiscoverCapture.ts
```diff
@@ -117,7 +120,8 @@ function useDiscoverCapture(
                     processedEntityName,
                     encryptedEndpointConfig.data,
                     options.initiateRediscovery,
-                    options.updateOnly
+                    options.updateOnly,
+                    dataPlaneName?.whole
```
travjenkins, line 124: If we're only using the nested `whole` but fetching and depending on `dataPlaneName` I think there could end up being unintended side effects. We should probably just fetch `whole` directly in the hook up above.

src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
```diff
@@ -0,0 +1,205 @@
...
+            ? {
+                  [DATA_PLANE_SCOPE]: {
+                      description: intl.formatMessage({
+                          id: 'workflows.dataPlane.description',
+                      }),
+                      oneOf: dataPlanesOneOf,
+                      type: 'object',
+                  },
+              }
+            : null;
```
travjenkins, line 142: If we are going to check 

```
dataPlaneOption === 'show_option'
```

at the end we should just check that before running anything and return null right at the start.

src/forms/renderers/DataPlaneSelector/DataPlaneIcon.tsx
```diff
@@ -0,0 +1,94 @@
...
+    if (provider === 'aws') {
+        return colorMode === 'light' ? awsLogo : awsLogoWhite;
+    }
+
+    if (provider === 'gcp') {
+        return gcpLogo;
+    }
+
+    return null;
+};
```
travjenkins, line 31: This function should just replaced with an object with the providers as keys.

src/images/data-plane-providers/google_cloud.png
travjenkins, line 1: Can we resize this - this thing is huge.

travjenkins (commented):
Some small callouts with the code. Going to start testing tomorrow.

travjenkins (approved):
lgtm - we can merge as is

src/components/capture/useDiscoverCapture.ts
kiahna-tucker: `dataPlaneName` was originally the complete string. When the data plane selector was developed and the name needed to be parsed, `dataPlaneName` became an object. The two workflow hooks that utilize the name did not have their dependency arrays properly updated at _that_ time. After the parsing kinks were resolved in the selector components, a commit was created that corrected this issue.

**Tl;dr:** There were commits on my local branch that were not pushed to the remote branch by the time you reviewed this PR.

src/api/discovers.ts
kiahna-tucker: That was not the case when I attempted this early on.

travjenkins (approved):
lgtm

Issue #1220, opened 2024-08-14
"Reset" data flows with single click in the UI


travjenkins:
### Overview

Need a way for the UI to allow users to do a full proper "backfill" 

Before implementing this we need to see if it is technically easier to pull this off by adding it to the details page OR work it into the existing flows.

### Notes from UI/UX
[The batch script](https://github.com/estuary/automatic-backfills) is not 100% correct and incorrect in small ways that are not huge. This is a good basis for making the final script

**The core workflow is:** 

1. Capture - Disable
2. Capture - Publish
    - Runtime must stop 100% and is done writing documents (wait for publication to succeed and then wait… or keep looking for shards)
    - (IMPORTANT) - save current time  
6. Bindings - update backfill property
7. Capture - enable
10. Materialization - update `notBefore` property
11. Capture & Materialization - Publish

Maybe still add a CTA next to backfill all button that tells the user to go to this flow.

### Original Notes
Control plane:
1. Extend automatic backfills to the collection level
2. Change it so that it first disables the capture and then does what it currently does in two publications
3. Turn it into a script that the UI could call to initiate a backfill when the user is in the loop

UI:
New button on the the binding and capture level in the edit ui for captures next to "Backfill" that is called "Reset Dataflow" and calls the script as a cloud function.

PR #1257, opened 2024-09-12
Full entity refresh start (not live)

travjenkins:
## Issues

contributes: https://github.com/estuary/ui/issues/1220
contributes: https://github.com/estuary/ui/issues/1258

### 1220

- disable the advanced editor when data flow reset is enabled
- move away from using the confirmation dialog due to it being too restrictive
- create a new `prompt` concept
- create a new `presave` prompt
- create hooks so the state of the prompt can be shared (needs a lot more work)
- updated hook to fetch related materializations
- move away from chips for selecting materializations
- show which collections a materialization write to in the dropdown

### misc

- Broke out the `generateDisabledSpec` stuff so that we can share the disabling ability in the table with the new flow
- start saving the live spec in a store during the entity existence check
- add a `fitWidth` option to `AlertBox` so we don't have to wrap it needlessly
- create `autoCompleteDefaults_Virtual` so `AutoComplete` can share values for non-`multiple` uses
- replacing more uses of `FormatMessage` with the `intl` hook

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

This is what the modal will end up looking like
image

It also supports when we want to use it for non-backfill (in the future)
image


src/components/shared/Entity/hooks/useDataFlowResetHandler.tsx
travjenkins, line 1: I doubt this will be used but leaving just in case. I think the `step` component will do the actual work to keep it easier... however not 100% sure yet.

src/app/guards/EntityExistenceGuard.tsx
travjenkins, line 1: Don't need to save this stuff off juuuust yet.

src/components/tables/RowActions/DisableEnable/Button.tsx
travjenkins, line 1: This is probably the biggest change that is going live.

src/services/ajv.ts
travjenkins, line 1: Happen to notice these during debugging. So adding them even though the UI does not use them.

src/stores/FormState/types.ts
```diff
@@ -50,6 +50,12 @@ export interface EntityFormState {
     isIdle: boolean;
     isActive: boolean;
 
+    showPreSavePrompt: boolean;
+    setShowPreSavePrompt: (data: EntityFormState['showPreSavePrompt']) => void;
+
+    liveSpec: Schema | null;
+    setLiveSpec: (data: EntityFormState['liveSpec']) => void;
```
travjenkins, line 57: Need a place to store this and I honestly don't care where cause eventually all the create/edit stores need to end up being in one big store anyway so added it here.

src/components/tables/RowActions/DisableEnable/Button.tsx
travjenkins: It also seems low risk as basically just moving a function into a shared spot

src/utils/entity-utils.ts
travjenkins, line 1: starting to make `utils` more specific. Also... we should probably end up dropping the `-utils` as the end of the file name :shrug:

kiahna-tucker (commented):
Releasing review comments while testing is in progress.

src/components/shared/AlertBox.tsx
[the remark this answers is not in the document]
travjenkins: I am mainly focused on new stuff and making sure things are not in the component file. For existing things it really depends on the impact moving the file will have.

src/components/shared/Entity/prompts/PreSave/Actions.tsx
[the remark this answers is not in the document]
travjenkins: Not until the thing is ready to go live.

src/components/shared/Entity/prompts/steps/dataFlowReset/SelectMaterialization/SelectorOption.tsx
```diff
@@ -0,0 +1,25 @@
...
+function SelectorOption({ option }: MaterializationSelectorOptionProps) {
+    const { catalog_name, reads_from } = option;
+
+    return (
+        <Stack component="span" direction="column" spacing={1}>
+            <Typography
+                component="span"
+                sx={{
+                    ...truncateTextSx,
+                }}
```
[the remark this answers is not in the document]
travjenkins: It previously had custom styling under it... I'll remove the spread

kiahna-tucker (approved):
Approved with minimal, happy path testing.

Issue #1260, opened 2024-09-19
Adding seconds during edit of `time travel` can cause issues


travjenkins:
**Bug**
Create materialization
Add binding
Add `notBefore` time with hour and minute
Exit time selector
Open time selector again
Add seconds
Value is - "notBefore": "2024-09-19T18:5:41Z"

**Expected**
The value should format the time correctly

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A (I think)

**Additional context**
N/A

travjenkins:
I see that this happened but not sure exactly how to get into this state. However, going to move the seconds into the flow to reduce the chance someone needs to enter values manually.

PR #1263, opened 2024-09-20
Time Travel UX changes

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1260

### 1260

- Add ability to set `seconds` with the picker
- Update UX to include current step in footer (helps make it clear when switching between minutes and seconds)
- Make the input a lot shorter to keep the picker closer to the value it is updating

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Added label
image

Now the user will be asked for the seconds
image


src/components/shared/pickers/DateTimePickerCTA.tsx
```diff
@@ -32,7 +34,22 @@ const formatDate = (formatValue: Date) => {
...
+        content &&
+        typeof content === 'object' &&
+        // @ts-expect-error How we call this we should get a react node
+        content.props &&
+        // @ts-expect-error We check if the thing exists so we should be okay
+        typeof content.props === 'object'
+    ) {
+        // @ts-expect-error being pretty safe up above
+        currentStep = content.props.focusedView ?? content.props.view ?? '';
+    }
```
kiahna-tucker, line 51: How would you feel about replacing the `if` condition with a custom type guard that evaluates whether a `ReactNode` is a `ReactElement`?

The type guard would take the following shape and stored off as a util:
```ts
export const isReactElement = (value: ReactNode): value is ReactElement =>
    isObject(value) && 'props' in value;
```

kiahna-tucker (commented):
Could you attach screenshots of the portions of the new UI to this PR?

kiahna-tucker (approved):

PR #1264, opened 2024-09-23
Depreciate data-plane gateway authorization for shard and journal information

kiahna-tucker:
## Issues

https://github.com/estuary/flow/issues/1627

### 1627

* Integrate the following control-plane APIs: `/authorize/user/task` and `/authorize/user/collection`.

* Use the dynamic broker address and authorization token generated by the authorize collection, control-plane API to retrieve the journal data for data preview. Previously, the journal client was used to source this information.

* Create a journal store to house journal state independent of the ops logs table.

* Remove the warning that displayed on the _Logs_ tab of the _Details_ page in a local environment.

### Manually tested

* Validate that shard information can be fetched for display on the entity tables and the _Details_ page of a given entity.

* Validate that the authorize task, control-plane API is not called for collections.

* Validate that journal information can be fetched for display in the ops logs table and in the data preview section of the _Overview_ tab.

### Automated tests

N/A

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

N/A


src/api/liveSpecsExt.ts
```diff
@@ -29,6 +29,7 @@ const baseColumns = [
     'spec_type',
     'updated_at',
     'last_pub_id',
+    'shard_template_id',
```
travjenkins, line 32: :tada: :tada: :tada: :tada: :tada: :tada:

src/components/tables/hooks/useRowsWithStatsState.ts
travjenkins, line 1: Future change (cause my branch is not done yet) but I'll probably pull this filter into the `entity-utils` file.

src/hooks/journals/useJournalData.ts
```diff
@@ -21,54 +23,66 @@ const useJournalsForCollection = (collectionName: string | undefined) => {
...
+    }, [brokerAddress, brokerToken]);
 
     const fetcher = useCallback(
         async (_url: string) => {
-            if (journalClient && collectionName) {
+            if (
+                journalClient &&
+                collectionName &&
+                brokerAddress &&
+                brokerToken
```
travjenkins, line 48: I _really_ wish typescript was smart enough to not need this.

src/hooks/journals/useJournalData.ts
```diff
@@ -21,54 +23,66 @@ const useJournalsForCollection = (collectionName: string | undefined) => {
...
                 }
 
                 return {
                     journals:
-                        dataPlaneListResponse.length > 0
-                            ? dataPlaneListResponse
+                        dataPlaneListResponse.result.journals &&
+                        dataPlaneListResponse.result.journals.length > 0
+                            ? dataPlaneListResponse.result.journals
                             : [],
```
travjenkins, line 71: If we know that the `dataPlaneListResponse.result.journals` is always an array then I think we could just always return it. Cause even if it is an empty array then that is basically what we're defaulting here.

src/hooks/journals/useJournalData.ts
```diff
@@ -81,8 +95,16 @@ const useJournalsForCollection = (collectionName: string | undefined) => {
...
-                if (session && shouldRefreshToken(`${error}`)) {
-                    await refreshAuthToken();
+                if (
+                    session &&
+                    shouldRefreshToken(`${error}`) &&
+                    collectionName
+                ) {
+                    await refreshAuthToken(
+                        session.access_token,
+                        collectionName,
```
travjenkins, line 105: This is fine for now... but we probably don't want to use the `collectionName` from the outer scope. Rather I _think_ we can read it from the `swr` config that should be getting passed in to the `onError`.

src/hooks/journals/useJournalData.ts
```diff
@@ -107,28 +129,34 @@ interface UseJournalDataSettings {
...
-        collectionName ? [collectionName] : null
+    const brokerAddress = useJournalStore((state) =>
+        opsJournalTarget
+            ? state.taskBrokerAddress
+            : state.collectionBrokerAddress
+    );
+
+    const brokerToken = useJournalStore((state) =>
+        opsJournalTarget ? state.taskBrokerToken : state.collectionBrokerToken
     );
```
travjenkins, line 149: Since we are using logic - please combine these two selectors into one. That way it reduces the chance of that logic being wrong if another selector is added.

src/hooks/shards/useShardsList.ts
```diff
@@ -1,85 +1,80 @@
...
+        if (!session) {
             return { shards: [] };
         }
 
-        const dataPlaneListResponse = await dataPlaneFetcher_list(
-            shardClient,
-            taskSelector,
-            'ShardsList'
-        );
+        const shardPromises = catalogNames.map(async (name) => {
```
travjenkins, line 21: I think we need to limit this and make sure there is some limits of how many will be called all at once.

Issue #1272, opened 2024-09-25
Default the data-plane selector to the GCP data-plane option


kiahna-tucker:
**Last Updated:** Sept. 25, 2024

<hr />

Presently, there are two, public data-planes to choose from: `gcp-us-central1-c1` and `aws-eu-west-1-c1`. Flow uses the GCP data-plane by default, therefore the data-plane selector should be defaulted accordingly.

Below are the base requirements for this feature:

* Default the data-plane selector to the GCP data-plane option (i.e., `gcp-us-central1-c1`) in production workflows.

## Additional Comments

* Investigate the ways in which the data-plane selector input can be customized. It wasn't obvious that the field could be interacted with and adding an arrow-like, end icon may make this clearer. That said, it _is_ possible that defaulting the field will sufficiently improve the UX.

PR #1278, opened 2024-09-26
Default the data-plane selector to GCP data-plane option

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1272

### 1272

* Fetch data plane options in details form store hydrate action.

* Default the data-plane selector to the GCP data-plane option (i.e., `gcp-us-central1-c1`) in production, create workflows.

* Default the data-plane selector to the single, out-of-the-box data-plane option (i.e., `local-cluster`) in local, create workflows.

* Display an error alert when the details form store fails to hydrate and `hydrationError` has a value.

* Store the suffix of the default data-plane option in an environment variable.

* Add a custom LogRocket event to log when the data-plane option corresponding to the `data_plane_id` in the URL cannot be found.

* Move `getDataPlaneScope` and `parseDataPlaneName` into the data-plane utils file.

* Create a shared constant for the public data-plane prefix (i.e., `ops/dp/public/`).

* Minorly refactor the `parseDataPlaneName` util.

### Manually tested

* Validate that the data-plane selector is defaulted to the GCP data-plane option (i.e., `gcp-us-central1-c1`) in production, create workflows.

* Validate that the data-plane selector is defaulted to the single, out-of-the-box data-plane option (i.e., `local-cluster`) in local, create workflows.

* Validate that the default value of the data-plane selector can be changed and retained when the page is refreshed in production and local, create workflows.

* Validate that the data-plane selector is defaulted to the option corresponding to the data plane ID in the URL in local, edit workflows.

* Validate that a LogRocket event is created when the data-plane option corresponding to the data plane ID in the URL is not found when hydrating local workflows.

* Validate that an error alert is displayed when data-plane options fail to be fetched or none are found in local workflows.

### Automated tests

* `getDataPlaneScope`

* `parseDataPlaneName`

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

### Production

**Capture create workflow**

pr_screenshot-1242-default_data_plane_selector-prod-capture_create

<br />
<br />

**Materialization create workflow**

pr_screenshot-1242-default_data_plane_selector-prod-materialization_create


src/api/dataPlanes.ts
travjenkins, line 1: Just comment out the entire function and don't export.

Then if it isn't needed for awhile we can fully remove later.

src/stores/DetailsForm/Store.ts
```diff
@@ -295,14 +318,56 @@ export const getInitialState = (
...
+                            };
+                        }
+                    );
+
+                    setDataPlaneOptions(dataPlaneOptions);
+                } else {
+                    setHydrationError(
+                        dataPlaneResponse.error?.message ??
+                            'An error was encountered initializing the details form. If the issue persists, please contact support.'
+                    );
```
kiahna-tucker, line 364: Not a fan of how the error is handled when no data-planes options could be found, but this seemed like the lesser of two evils. The alternative approach would be to listen to both `hydrationErrorsExist` and `hydrationError` in the `EntityCreate` and `EntityEdit` components and display (internationalized) content if `hydrationErrorsExist` is `true` but `hydrationError` is undefined.

src/api/dataPlanes.ts
kiahna-tucker: Yep, you started reviewing before the branch was ready. Haha resolved.

src/services/types.ts
```diff
@@ -16,6 +16,7 @@ export enum CustomEvents {
     CAPTURE_TEST = 'Capture_Test',
     COLLECTION_CREATE = 'Collection_Create',
     CONNECTOR_VERSION_UNSUPPORTED = 'Connector_Version:Unsupported',
+    DATA_PLANE_SELECTOR_DEFAULTED = 'Data_Pane_Selector:Unexpectedly_Defaulted',
```
travjenkins, line 19: Sorry - this isn't really the pattern anymore.

You can just have the event be `DATA_PLANE_SELECTOR` and then set properties in the data that is being passed to LR. I noticed that it makes updating queries and segments a lot easier that way.

src/stores/DetailsForm/Store.ts
```diff
@@ -295,14 +318,56 @@ export const getInitialState = (
...
+                            const { cluster, prefix, provider, region } =
+                                parseDataPlaneName(data_plane_name, scope);
+
+                            return {
+                                dataPlaneName: {
+                                    cluster,
+                                    prefix,
+                                    provider,
+                                    region,
+                                    whole: data_plane_name,
```
travjenkins, line 351: I think we should just put that inside the `parseDataPlaneName` function. Even though it is kinda weird that it is just passing back what was provided... but then we wouldn't need the destruct up above and we have a "one stop shop" for name related stuff.

src/utils/dataPlane-utils.ts
```diff
@@ -139,6 +140,85 @@ export const getJournals = async (
...
+
+    const suffix =
+        slashIndex === -1
+            ? truncatedName
+            : truncatedName.substring(slashIndex + 1);
+
+    return [prefix, suffix];
+};
+
+const splitDataPlaneSuffix = (suffix: string, firstHyphenIndex: number) => {
```
travjenkins, line 173: This is good for now as it is super understandable... but this seems like the rare case possibly splitting up the string with a RegEx could be easier.

src/utils/dataPlane-utils.ts
```diff
@@ -139,6 +140,85 @@ export const getJournals = async (
...
+    const region = regionOnly
+        ? suffix.substring(firstHyphenIndex + 1)
+        : suffix.substring(firstHyphenIndex + 1, lastHyphenIndex);
+
+    const cluster = regionOnly ? '' : suffix.substring(lastHyphenIndex + 1);
+
+    return [provider, region, cluster];
+};
+
+export const parseDataPlaneName = (
```
travjenkins, line 190: This is much nice with the broken out functions

src/utils/env-utils.ts
```diff
@@ -20,6 +20,8 @@ declare global {
 const ENABLED = 'true';
 
 export const isProduction = import.meta.env.PROD;
+export const defaultDataPlaneSuffix = import.meta.env
+    .VITE_DEFAULT_DATA_PLANE_SUFFIX;
```
travjenkins, line 24: This is risky... so if this is not set this would fail silently. Unless you know of a way where this could support being an empty string.

I don't like all the `getXYZ` in that we need to keep calling functions that we shouldn't... but here I think we can just write a iife that sets this and then throws an error if it isn't there.

src/stores/DetailsForm/Store.ts
kiahna-tucker: Agreed.

src/utils/dataPlane-utils.ts
kiahna-tucker: Left a comment behind for this.

travjenkins (approved):
lgtm

Issue #1271, opened 2024-09-25
Meter the control-plane, task authorization API call


kiahna-tucker:
**Last Updated:** Sept. 25, 2024

<hr />

To display shard status in an entity table, the control-plane, task authorization API, `/authorize/user/task`, is called for each task row and followed by a call to the shards list API, `/shards/list`. This can result in a considerable number of concurrent API calls, with the task authorization information needlessly being fetched. 

Below are the base requirements for this feature:

* Cache the reactor address and token returned by the control-plane, task authorization API.

PR #1293, opened 2024-10-03
Cache response of the control-plane, task authorization API

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1271

### 1271

* Cache response of the control-plane, task authorization API.

* Move `useReactorToken` into the `hooks/gatewayAuth/` directory.

### Manually tested

### Automated tests

N/A

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

N/A


src/hooks/gatewayAuth/useTaskAuthorization.ts
```diff
@@ -0,0 +1,100 @@
...
+        async (prefix): Promise<TaskAuthorizationFetcherResponse> => {
+            // The endpoint should stay the same so just using the prefix and the access token as the cache key.
+            const cacheKey = `${prefix}__task-auth__${accessToken}`;
+
+            const gatewayConfig = getAuthorizationConfig(cacheKey);
+            const tokensExpired = validateJWTs(gatewayConfig);
+
+            // If the token is valid and the cached configuration is defined, then return the cached value.
+            if (!tokensExpired && gatewayConfig) {
+                return Promise.resolve({ ...gatewayConfig, cached: true });
```
travjenkins, line 38: Since it does not look like we're doing anything with it - let's remove the `cached` flag.

src/hooks/gatewayAuth/useTaskAuthorization.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const { data, mutate } = useSWR(
+        hasLength(authorizedPrefixes)
+            ? [authorizedPrefixes, session?.access_token]
+            : null,
+        gatewayFetcher,
+        {
+            onSuccess: (configs) => {
+                if (!hasLength(configs)) {
+                    return;
```
travjenkins, line 74: Should we add an event here? Trying to think of "real world" scenarios where this would happen.

src/hooks/gatewayAuth/cache.ts
```diff
@@ -0,0 +1,59 @@
+import { isBefore } from 'date-fns';
+import { decodeJwt, JWTPayload } from 'jose';
+import { LRUMap } from 'mnemonist';
+import { TaskAuthorizationResponse } from 'utils/dataPlane-utils';
+
+// The limit of 10 is totally arbitrary. Felt like an okay number so all 3 tables could have
+//  something in cache and allow someone some flexibility to search
```
travjenkins, line 7: We should rewrite this comment now. I think just calling out it is generally arbitrary is enough.

src/hooks/gatewayAuth/cache.ts
```diff
@@ -0,0 +1,59 @@
...
+        brokerJWT = config?.brokerToken
+            ? decodeJwt(config.brokerToken)
+            : undefined;
+        reactorJWT = config?.reactorToken
+            ? decodeJwt(config.reactorToken)
+            : undefined;
+    } catch {
+        brokerJWT = undefined;
+        reactorJWT = undefined;
+    }
```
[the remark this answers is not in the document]
kiahna-tucker: Since these tokens cannot be isolated and the entire task authorization response is cached, it made sense to be sensitive to both tokens here.

src/hooks/gatewayAuth/useTaskAuthorization.ts
```diff
@@ -0,0 +1,100 @@
...
+    );
+
+    if (!hasLength(authorizedPrefixes)) {
+        logRocketEvent(CustomEvents.GATEWAY_TOKEN_INVALID_PREFIX, {
+            prefixes,
+        });
+    }
+
+    const { data, mutate } = useSWR(
+        hasLength(authorizedPrefixes)
```
travjenkins, line 67: If we require session `access_token` then I think we should probably check for it as will.

src/hooks/gatewayAuth/useTaskAuthorization.ts
kiahna-tucker: I went back and forth on this one. To err on the side of caution, I will throw one in.

travjenkins (approved):
lgtm

src/hooks/shards/useShardsList.ts
```diff
@@ -34,10 +52,10 @@ const useShardsList = (catalogNames: string[]) => {
 
     const swrKey = useMemo(
         () =>
-            session && catalogNames.length > 0
-                ? `shards-${catalogNames.join('-')}`
+            session && taskAuthorizationData?.length && catalogNames.length > 0
```
travjenkins, line 55: Asked a question in convo so marking answer here.

SWR will serialize the array passed in when generating the cache key on their end.

PR #1295, opened 2024-10-04
PROD: Clean up the GitPod URL more

travjenkins:
Missed this portion of the path - though it seems to not break anything (yet)


kiahna-tucker (approved):

PR #1296, opened 2024-10-04
PROD: putting the GitPod link back

travjenkins:
The URL was fine - https://github.com/gitpod-io/gitpod/issues/20263 

Still breaking the code up a bit to make it more understandable

image


kiahna-tucker (approved):

Issue #1305, opened 2024-10-10
Previously disabled bindings stay in draft after a refresh


travjenkins:
**Bug**
Create a capture
Disable bindings A,B,C without keys
Edit capture
Refresh bindings
Disable new bindings Z,Y  without keys
Click Save
Publication fails due to collections missing keys

**Expected**
Should be able to save.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Additional context**
he UI is only deleting bindings that are disabled within the UI. We should also delete any bindings in the spec that are marked disabled.

PR #1306, opened 2024-10-10
Removing all disabled bindings based on the draft during save

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1305

### 1305

- No longer keep a list of disabled collections
- Just read the bindings directly from the draft
     - Kinda sucks we have to loop over but this is by far the safest approach

### Manually tested

- Discovered and rediscovered with postgres locally 

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [x] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [x] Materialization

### Edited capture and did a rediscover. Left all the previously disabled disabled and those are now removed.
image


kiahna-tucker (approved):

Issue #1194, opened 2024-08-05
Add `interval` to the `capture` create/edits


travjenkins:
We need to add the `interval` property on the process for creating and editing `Captures`.

![image](https://github.com/user-attachments/assets/377cd168-3214-4ef3-97be-798546755a46)

This should _probably_ end up getting added in a similar space as `backfill' and `auto discover` stuff. However, we could also easily make this _appear_ to be within endpoint config.

The input should end up something like this:
![image](https://github.com/user-attachments/assets/d9a1700a-2972-4af7-b0e1-6ea996b9b20f)
![image](https://github.com/user-attachments/assets/574d5b52-c6b2-493f-82e5-f9d1ec28d5d0)

However, we might just end up having the user type `s`, `m`, or `h` as that is also easy
![image](https://github.com/user-attachments/assets/d8436fbf-4fb7-46c7-912e-1381a530a2fb)
![image](https://github.com/user-attachments/assets/36b7293f-c7b6-44ef-820f-ffb3be25ebf0)

kiahna-tucker:
The column, `default_capture_interval`, on the `connector_tags` table dictates whether the field should appear and what the default value should be.

kiahna-tucker:
The capture interval field has four options in the time unit menu: seconds, minutes, hours, and interval (in that order). The interval option was added to support capture intervals that span more than one time unit (e.g., _four hours and thirty minutes_). Please note the following characteristics:

* A _compound_ interval can take the shape of a Postgres interval (i.e., `HH:MM:SS`) or be in the format the specification expects (e.g., `1h 20m 10s`, `1h 10s`, etc.).

* The selected time unit will be updated to _interval_ accordingly if the field input matches a supported interval format.

* Plain, numeric input is considered invalid when the selected time unit is _interval_.

* When the selected time unit moves from the _interval_ option, the entered interval will be parsed in alignment with the new time unit. If the entered interval does not contain a segment that corresponds to the new time unit, the field will be empty. For example, say the entered interval is `1h 30m`. If the _hours_ option is selected, the field will contain `1`. If the _minutes_ option is selected, the field will contain `30`. If the _seconds_ option is selected, the field will be empty.

dyaffe:
[...]

PR #1319, opened 2024-10-18
Extend capabilities of the hidden, capture `interval` field

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1194

### 1194

* Extend the logic for the component that exposes the `interval` property in capture workflows.

**NOTE:** The capture interval field will **remain hidden** as an investigation into the `Duration` formats supported by the backend is underway. Additionally, it should be noted that `luxon` should be leveraged to handle converting to and from strings in a `Duration` format (ref: `luxon` [documentation](https://moment.github.io/luxon/api-docs/index.html#duration)).

### Manually tested

* Validate that the capture interval section does not appear in any workflows.

### Automated tests

* `formatCaptureInterval`

* `getCaptureIntervalSegment`

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

N/A


src/stores/Binding/Store.ts
```diff
@@ -219,6 +224,76 @@ const initializeAndGenerateUUID = (
...
+    if (!hasLength(connectorTagId)) {
+        return null;
+    }
+
+    const { data, error } = await getSchema_Resource(connectorTagId);
+
+    if (error) {
+        get().setHydrationErrorsExist(true);
+    } else if (data?.resource_spec_schema) {
+        const { setBackfillSupported, setResourceSchema } = get();
```
travjenkins, line 240: Just access directly.

src/stores/Binding/Store.ts
```diff
@@ -433,47 +509,51 @@ const getInitialState = (
...
+                fallbackInterval,
+                get,
+                liveSpecs[0].spec,
+                searchParams
+            );
+
+            if (draftSpecError) {
+                get().setHydrationErrorsExist(true);
+
+                return Promise.reject(draftSpecError.message);
```
travjenkins, line 548: Since we do nothing with this promise rejection we should not pass back a value with it.

src/stores/Binding/Store.ts
```diff
@@ -433,47 +509,51 @@ const getInitialState = (
...
-                    get().prefillBindingDependentState(
-                        entityType,
-                        liveSpecs[0].spec.bindings
-                    );
-                }
+
+                return Promise.reject(
+                    liveSpecError?.message ??
+                        `An issue was encountered fetching the live specification for this ${entityType}`
+                );
```
travjenkins, line 533: Same as below.

src/utils/time-utils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export const formatCaptureInterval = (
+    interval: string | null | undefined,
+    intervalUnitSupported?: boolean
+): string | null => {
+    if (typeof interval !== 'string') {
+        return null;
+    }
+
+    let formattedInterval = '';
```
travjenkins, line 73: Instead of using concat on a string this should just be an array that is joined at the end.

travjenkins: That also helps in making sure the last element does not have the join char.

src/utils/time-utils.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+const parsePostgresInterval = (interval: string): SegmentedInterval => {
+    const [hours, minutes, seconds] = interval.split(':').map((segment) => {
+        const numericSegment = Number(segment);
+
+        return isFinite(numericSegment) ? numericSegment : -1;
+    });
+
+    return { h: hours, m: minutes, s: seconds };
```
travjenkins, line 21: What is driving the renaming to a single letter? The only consumer of this just renames them back to what they were.

src/utils/time-utils.ts
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const parseCaptureInterval = (interval: string): SegmentedInterval => {
+    let hours = -1;
+    let minutes = -1;
+    let seconds = -1;
+
+    interval.split(' ').forEach((segment) => {
+        const numericSegment = Number(segment.substring(0, segment.length - 1));
+        const numericValue = isFinite(numericSegment) ? numericSegment : -1;
```
travjenkins, line 31: We repeat this check so should probably just throw into a function.

src/utils/time-utils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+        if (segment.includes('h')) {
+            hours = numericValue;
+        } else if (segment.includes('m')) {
+            minutes = numericValue;
+        } else if (segment.includes('s')) {
+            seconds = numericValue;
+        }
```
travjenkins, line 39: We might want to check `endsWith` to be really explicit.

src/utils/time-utils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export const getCaptureIntervalSegment = (interval: string, unit: string) => {
+    let truncatedInterval = -1;
+
+    const granularTimeUnit = unit === 'h' || unit === 'm' || unit === 's';
+
+    if (granularTimeUnit && POSTGRES_INTERVAL_RE.test(interval)) {
+        const segments = parsePostgresInterval(interval);
+
+        truncatedInterval = segments[unit];
```
travjenkins, line 53: I think we can just make this and the line below a single line.

src/utils/time-utils.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (CAPTURE_INTERVAL_RE.test(interval)) {
+        formattedInterval = formattedInterval.concat(interval);
+    }
+
+    formattedInterval =
+        hasLength(formattedInterval.trim()) &&
+        intervalUnitSupported &&
+        !DURATION_RE.test(formattedInterval.trim())
+            ? formattedInterval.trim().concat('i')
+            : formattedInterval.trim();
```
travjenkins, line 104: If we have to call `trim` on this variable constantly just call it once before.

src/utils/time-utils.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (seconds > 0) {
+            formattedInterval = formattedInterval.concat(`${seconds}s`);
+        }
+    }
+
+    if (CAPTURE_INTERVAL_RE.test(interval)) {
+        formattedInterval = formattedInterval.concat(interval);
+    }
+
+    formattedInterval =
```
travjenkins, line 99: I would probably just break this into an if

src/utils/time-utils.ts
travjenkins, line 1: There are two places where the regexs are tested with separate `if`s. It looks like these pattern matches must be mutually exclusive. Those `if`s should probably be switched over to include an `else if`

src/utils/time-utils.ts
travjenkins: We should also probably use `isSafeInteger` to make sure these numbers don't overflow or anything.

src/utils/__tests__/time-utils.test.ts
travjenkins, line 1: Please add a new more "insane edgecases". Since today you can parse `999999999999999999999999999999999999999s` today and that should probably be called out.

src/hooks/captureInterval/useCaptureInterval.ts
```diff
@@ -0,0 +1,183 @@
...
+
+            let spec: Schema = draftSpec.spec;
+            const originalInterval = draftSpec.spec?.interval;
+
+            if (!hasLength(interval)) {
+                spec = omit(spec, 'interval');
+            } else {
+                const strippedInterval = interval.endsWith('i')
+                    ? interval.slice(0, interval.length - 1)
+                    : interval;
```
travjenkins, line 75: If we are doing stuff like this then the `i` needs to be a const and this kind of action should go into the utils.

src/stores/Binding/Store.ts
kiahna-tucker: As previously discussed, this message is relayed in the LogRocket console in the hydrator in an effort to aid debugging. The details form store takes a similar approach which can be quite helpful.

src/stores/Binding/Store.ts
kiahna-tucker: Same as below.

src/utils/time-utils.ts
kiahna-tucker: The value of the select menu items and the need for `getCaptureIntervalSegment`, which is the second consumer of these functions.

src/utils/__tests__/time-utils.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        expect(getCaptureIntervalSegment('00:01:30', 'h')).toBe(0);
+
+        expect(getCaptureIntervalSegment(postgresInterval, 'm')).toBe(0);
+
+        expect(getCaptureIntervalSegment(postgresInterval, 's')).toBe(0);
+
+        expect(getCaptureIntervalSegment('2m 0s', 's')).toBe(0);
+    });
+
+    test('returns a number greater than 0 when the time unit corresponds to an interval segment that is non-zero', () => {
```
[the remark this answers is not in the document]
kiahna-tucker: Sure, I can roll it up into the test above.

src/utils/time-utils.ts
kiahna-tucker: Given this function is only consumed after a string is confirmed to take the capture interval shape, this _should_ be fine. There is a minor concern about the regular expression as it isn't as tight as I would prefer (e.g., `"2h  "` is a valid interval), but the change handlers _should_ trim leading and trailing spaces appropriately.

src/utils/time-utils.ts
kiahna-tucker: I'm not the biggest fan of indexing a function's return argument directly but sure, if that is your preference.

src/utils/time-utils.ts
kiahna-tucker: As previously mentioned, this PR was opened before the logic to support capture and Postgres intervals was reviewed and refactored.

src/utils/__tests__/time-utils.test.ts
kiahna-tucker: There are limits the to values that can be used to test but the test coverage can be extended within reason.

src/utils/time-utils.ts
kiahna-tucker: As previously mentioned, this PR was opened before the logic to support capture and Postgres intervals was reviewed and refactored.

src/hooks/captureInterval/useCaptureInterval.ts
```diff
@@ -0,0 +1,183 @@
...
+                    formatCaptureInterval(strippedInterval);
+
+                if (
+                    !formattedInterval ||
+                    !CAPTURE_INTERVAL_RE.test(formattedInterval)
+                ) {
+                    logRocketConsole(CustomEvents.CAPTURE_INTERVAL_ERROR, {
+                        interval,
+                        formattedInterval,
+                    });
```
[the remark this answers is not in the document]
kiahna-tucker: Could you explain your reasoning here? At the very least, the highlighted line should remain a console log given it isn't a hard or fixed error, for lack of better term.

src/stores/Binding/Store.ts
kiahna-tucker: This is an oversight given the PR was opened before it was reviewed. This PR is older than that which changed how the state getter is interacted with.

src/hooks/captureInterval/useCaptureInterval.ts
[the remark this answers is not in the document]
travjenkins: For something like this we mainly care to find when they happen. Events are searchable where logs are not (without paying more money)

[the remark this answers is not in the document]
kiahna-tucker: How do you feel about "false positives?" This particular line could be reached often (depending on the user's actions) and it does not always indicate an error.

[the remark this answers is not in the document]
travjenkins: Those are worse then any error and need to be found so they can be removed.

PR #1320, opened 2024-10-18
Change how `auth_roles` are saved in store

travjenkins:
## Issues

tech debt - This has been an issue awhile and has been more and more annoying as we keep growing and getting new tenants. This is really only going to change things for `estuary_support` folks.

### misc

-  stop storing auth roles in an object and just keep it as an array of strings

### Manually tested

- hit the site pointing locally and to prod. Made sure things loaded.

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

We now use the normal virtual selector and that shows a check next to the selection
image


src/components/shared/AutoComplete/DefaultProps.tsx
travjenkins, line 1: After some other PRs I'm gonna clean these things up a bit more. For now just throwing in a new one.

src/components/shared/guards/AdminCapability.tsx
travjenkins, line 1: Made a new hook to just fetch the length since sometimes that's all we're checking.

src/stores/Entities/Store.ts
travjenkins, line 1: We only ever read this stuff as arrays of strings. This also helps performance a bungh... going from around 2500ms to 250ms.

src/stores/Entities/hooks.ts
travjenkins, line 1: Still need to use `Set`s here so we don't get duplicates.

src/stores/Entities/hooks.ts
```diff
@@ -35,17 +35,21 @@ export const useEntitiesStore_capabilities_readable = () => {
 //     );
 // };
 
-export const useEntitiesStore_capabilities_adminable = () => {
+export const useEntitiesStore_atLeastOneAdminTenant = () => {
```
kiahna-tucker, line 38: What do you think about `adminExists` or `adminTenantExists` as an alternative term? The current name is _fine_ as-is but I know this convention is used elsewhere for existence checks.

travjenkins: With the change in how we store things I think we can also change the approach to naming. So potentially good for future.

I think `adminExists` was one I ruled out because I'd worry we would add other "admin" things (like `adminUser`) that we might also need to check.

PR #1334, opened 2024-10-30
Adding CSP nonce and manually handling GTM

travjenkins:
Cleaning up handing always

## Issues

https://github.com/estuary/security/issues/196

### 196

- Manually load in GTM so we can more easily set the nonce. Also - do not use most of the features in the plugin so it was kind of wasteful
- Adding in nonce support (kinda)


.env.development.local
travjenkins, line 1: Need to set this in dev version so locals don't run gtm

public/nginx.conf
```diff
@@ -128,6 +139,8 @@ http {
             alias   /usr/share/nginx/html/static;
             expires 1y;
             add_header Cache-Control "public";
+            add_header X-Frame-Options SAMEORIGIN;
+            add_header X-Content-Type-Options nosniff;
```
travjenkins, line 143: A config validator called out that these might be missing when serving static content

skord (approved):
per the call and extra review, LGTM

Issue #1353, opened 2024-11-04
Allow `optional` objects in config to be removed


travjenkins:
Once you add an object to the schema is must stay there. This makes it impossible for the user to remove an `optional` object (without using the advanced schema editor).

We should add a `Remove Settings` kind of thing from `optional` objects in the forms.

Issue #1352, opened 2024-11-02
Add informational statement on materialization reporting view 


dyaffe:
**Background**
Reporting in our system is delayed by up to 2x the update delay that's set in the system.  This is confusing for users and makes them think that we aren't syncing.  It would be great to let them know this to avoid support requests.

**Proposal**
In the materialization details histogram reporting view, we should add an informational statement at the bottom right that is based on the actual update delay for the materialization.  It should only appear if the update delay is greater than 30 minutes and say "Reporting can be delayed by up to 2x the set update delay or up to 4 hours for this materialization." < this would be for a materialization with a 2 hour update delay.

PR #1355, opened 2024-11-04
Adding ability for collapsible group to reset data and details graph message

travjenkins:
## Issues

completes https://github.com/estuary/ui/issues/1353
completes https://github.com/estuary/ui/issues/1352

### 1353

- Add ability to clear collapsible sections

### 1352

- Add message that informs users about potential delay
- Added a backup message in case we cannot get duration time

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

### 1353
Show a button with the same label
image

Shows enabled when there is data
image

Tooltip to explain
image

We will show two clears for different objects to keep it simple right now
image

### 1352
Show a delay message
image
image

Show default delay if we cannot parse sync OR it is `0s`
image

### DIFF
--- a/src/components/shared/Entity/Details/Usage/DelayWarning.tsx
+++ b/src/components/shared/Entity/Details/Usage/DelayWarning.tsx
@@ -1,0 +1,29 @@
+import { Stack, Tooltip, Typography } from '@mui/material';
+import { useSyncScheduleDelayWarning } from 'hooks/details/useSyncScheduleDelayWarning';
+import { HelpCircle } from 'iconoir-react';
+import { useIntl } from 'react-intl';
+
+function DelayWarning() {
+    const intl = useIntl();
+
+    const reportingDelayMessage = useSyncScheduleDelayWarning();
+
+    if (!reportingDelayMessage) {
+        return null;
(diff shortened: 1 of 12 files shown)


src/components/shared/Entity/Details/Usage/DelayWarning.tsx
travjenkins, line 1: Moved to a stand alone component as I think we _might_ show this elsewhere. Also, don't need the full graph rendering more times than needed.

src/hooks/details/useSyncScheduleDelayWarning.ts
travjenkins, line 1: This is only for `materialization`s but instead of checking for them specifically I went ahead and just based on the config. That way if we ever add these settings to other entity types then this will just work.

kiahna-tucker (approved):
Approved with minimal testing. The couple UX-related reservations I mentioned last week still stand but, if the current UX has been discussed and approved, that is fine with me.

PR #1358, opened 2024-11-06
Introduce an `interval` field to capture workflows

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1194

### 1194

* Simplify the logic for the component that exposes the `interval` property in capture workflows.

### Manually tested

* Validate that the capture interval section does not appear in materialization workflows.

* Validate that the capture interval section does not appear in capture workflows when the selected connector does not specify a default capture interval.

* Validate that the field is prefilled with the default interval in the capture create workflow when the selected connector specifies one.

* Validate that the field is prefilled with the default interval in the capture edit workflow when the live specification does not specify an interval but a default capture interval exists.

* Validate that the field is prefilled with the interval specified in the live specification in the capture edit workflow when it exists.

* Validate that the field considers input in an **unsupported** interval format invalid.

### Automated tests

Test coverage updated for the following utils:

* `formatCaptureInterval`

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

**Capture interval field | Value selected**

pr_screenshot-1358-capture_interval-standard-single_unit_interval

<br />
<br />

**Capture interval field | Tooltip**

pr_screenshot-1319-capture_interval-standard-tooltip

<br />
<br />

**Capture interval field | Predefined interval options**

pr_screenshot-1358-capture_interval-standard-menu-options

<br />
<br />

**Capture interval field | Empty**

pr_screenshot-1358-capture_interval-standard-empty

<br />
<br />

**Capture interval field-level error**

pr_screenshot-1358-capture_interval-standard-error-input-compound

### DIFF
--- a/src/components/capture/Interval/index.tsx
+++ b/src/components/capture/Interval/index.tsx
@@ -23,8 +23,6 @@ import { hasLength } from 'utils/misc-utils';
 import { CAPTURE_INTERVAL_RE } from 'validation';
 import { CaptureIntervalProps } from './types';
 
-const DESCRIPTION_ID = 'capture-interval-description';
-
 function CaptureInterval({ readOnly }: CaptureIntervalProps) {
     const intl = useIntl();
     const label = intl.formatMessage({
@@ -124,7 +122,7 @@ function CaptureInterval({ readOnly }: CaptureIntervalProps) {
                 value={input}
             />
 
-            <FormHelperText id={DESCRIPTION_ID} style={{ marginLeft: 0 }}>
+            <FormHelperText style={{ marginLeft: 0 }}>
                 {intl.formatMessage(
                     { id: 'captureInterval.input.description' },
                     {
@@ -137,11 +135,7 @@ function CaptureInterval({ readOnly }: CaptureIntervalProps) {
             </FormHelperText>
 
             {errorsExist ? (
-                <FormHelperText
-                    id={DESCRIPTION_ID}
-                    error={errorsExist}
-                    style={{ marginLeft: 0 }}
-                >
+                <FormHelperText error={errorsExist} style={{ marginLeft: 0 }}>
                     {intl.formatMessage({
                         id: 'captureInterval.error.intervalFormat',
                     })}
--- a/src/components/capture/Interval/index.tsx
+++ b/src/components/capture/Interval/index.tsx
@@ -115,7 +115,6 @@ function CaptureInterval({ readOnly }: CaptureIntervalProps) {
                         error={errorsExist}
                         label={label}
                         size="small"
-                        variant="outlined"
                     />
                 )}
                 sx={{ width: 400 }}
--- a/src/components/capture/Interval/index.tsx
+++ b/src/components/capture/Interval/index.tsx
@@ -1,19 +1,12 @@
 import {
-    FormControl,
+    Autocomplete,
+    AutocompleteRenderInputParams,
     FormHelperText,
-    InputAdornment,
-    InputLabel,
-    MenuItem,
-    OutlinedInput,
-    Select,
     Stack,
+    TextField,
     Tooltip,
     Typography,
 } from '@mui/material';
-import {
-    primaryButtonText,
-    primaryColoredBackground_hovered,
-} from 'context/Theme';
 import useCaptureInterval from 'hooks/captureInterval/useCaptureInterval';
 import { HelpCircle } from 'iconoir-react';
 import { isEmpty } from 'lodash';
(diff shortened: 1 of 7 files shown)


src/components/capture/Interval/index.tsx
```diff
@@ -108,162 +78,69 @@ function CaptureInterval({ readOnly }: CaptureIntervalProps) {
...
+                    <TextField
+                        {...params}
+                        InputProps={{
+                            ...InputProps,
+                            sx: { borderRadius: 3 },
+                        }}
+                        error={errorsExist}
+                        label={label}
+                        size="small"
+                        variant="outlined"
```
travjenkins, line 118: This should probably match the other inputs and just use standard

src/components/capture/Interval/index.tsx
```diff
@@ -108,162 +78,69 @@ function CaptureInterval({ readOnly }: CaptureIntervalProps) {
...
+            />
+
+            <FormHelperText style={{ marginLeft: 0 }}>
+                {intl.formatMessage(
+                    { id: 'captureInterval.input.description' },
+                    {
+                        value: Duration.fromObject(defaultInterval).toHuman({
+                            listStyle: 'long',
+                            unitDisplay: 'short',
+                        }),
```
travjenkins, line 132: Since `toHuman' does not filter out `0` let's go ahead and do it ourselves. It is really weird to see 0 hours, n minutes, 0 seconds

src/utils/time-utils.ts
```diff
@@ -1,105 +1,26 @@
...
-        if (minutes > 0) {
-            segments.push(`${minutes}m`);
-        }
-
-        if (seconds > 0) {
-            segments.push(`${seconds}s`);
-        }
+        formattedInterval = Duration.fromObject(
+            parsePostgresInterval(interval)
+        ).toFormat("h'h' m'm' s's'");
```
travjenkins, line 21: Weird - but can you store this in a string template. That way we can put whatever in the format pattern.

src/utils/__tests__/time-utils.test.ts
travjenkins, line 1: Add a few more edge cases - that way we have some documentation that calls out we allow really really really huge numbers in here.

src/components/capture/Interval/index.tsx
```diff
@@ -57,31 +41,17 @@ function CaptureInterval({ readOnly }: CaptureIntervalProps) {
...
-            unit === 'i'
-                ? intervalErrorsExist
-                : intervalErrorsExist && !NUMERIC_RE.test(input)
-        );
-    }, [input, unit]);
+    const errorsExist = useMemo(
+        () =>
+            Boolean(
+                input && hasLength(input) && !CAPTURE_INTERVAL_RE.test(input)
+            ),
```
travjenkins, line 52: We have basically the same test in 3 places - we should just make this a general function in the time utils.

src/components/capture/Interval/index.tsx
kiahna-tucker: I am not a fan of the look of the standard input here. Dave will likely express the sentiment that the field does not appear to be interactive or is easily overlooked, but I will update it since that is your preference.

src/utils/__tests__/time-utils.test.ts
kiahna-tucker: The testing library does restrict the size of the numbers that can be tested here, but I can extend the tests to sit just before the limitation. I can also note this in a comment.

src/components/capture/Interval/index.tsx
kiahna-tucker: Sweeping through the code to see where this condition appears three times. As of right now, I do not see where this is occurring.

src/stores/Binding/Store.ts
```diff
@@ -1024,8 +1021,18 @@ const getInitialState = (
...
+                    const intervalObject =
                         Duration.fromISOTime(defaultInterval).toObject();
+
+                    Object.entries(intervalObject).forEach(
+                        ([timeUnit, unitValue]) => {
+                            if (unitValue === 0) {
+                                intervalObject[timeUnit] = undefined;
+                            }
+                        }
+                    );
```
travjenkins, line 1033: These should be filtered out in the format function and not the store. I know this means we'll run the filter more but it isn't that huge of a piece of logic.

kiahna-tucker: The format function does not alter the default interval, but I can look into rearranging things to make that work.

travjenkins: Just make it a boolean option so we can control it just for the default output.

The goal is to have this logic where it should be eventually when luxon does add this feature : https://github.com/moment/luxon/issues/1134

travjenkins: We could also just make a new function just for the defaulted option... since it is a special case and could be easier to understand than an extra boolean.

travjenkins (approved):
lgtm - didn't manually test the last round of changes

PR #1372, opened 2024-11-22
Expose `onIncompatibleSchemaChange`  properties in materialization workflows

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1375

### 1375

* Enable the specification-level `onIncompatibleSchemaChange` property to be configured in materialization workflows. An autocomplete field located in the top-level _Backfill_ section accomplishes this.

* Create a hook which updates the specification-level `onIncompatibleSchemaChange` property.

* Extend `BindingState` with an `onIncompatibleSchemaChange` property that captures the value of the specification-level `onIncompatibleSchemaChange` property.

* Create a utility function that updates or removes the specification-level `onIncompatibleSchemaChange` property.

* Enable the binding-level `onIncompatibleSchemaChange` property to be configured in materialization workflows. An autocomplete field located in the binding-level _Backfill_ section accomplishes this.

* Rename the hook which updates the binding-level `onIncompatibleSchemaChange` property.

* Extend `BindingState` with a `ResourceConfig['meta']['onIncompatibleSchemaChange']` property that captures the value of the binding-level `onIncompatibleSchemaChange` property.

### Manually tested

* Validate that the incompatible schema change section does not appear in capture workflows.

* Validate that the field initialized properly when editing a materialization that has the specification-level, `onIncompatibleSchemaChange` property set to a supported value.

* Validate that the drafted specification is updated properly when the field is set to a supported value and cleared.

### Automated tests

N/A

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

This field is positioned between the top-level _Backfill_ section and the binding selector.

**Spec-level, `onIncompatibleSchemaChange` field | Empty**

pr_screenshot-1372-incompatible_schema-spec_level-empty

<br />
<br />

**Spec-level, `onIncompatibleSchemaChange` field | Menu options**

pr_screenshot-1372-incompatible_schema-spec_level-menu_options

<br />
<br />

**Spec-level, `onIncompatibleSchemaChange` field | Option selected**

pr_screenshot-1372-incompatible_schema-spec_level-option_selected

<br />
<br />

**Spec-level, `onIncompatibleSchemaChange` field | Unsupported value detected | String**

pr_screenshot-1372-incompatible_schema-spec_level-error-invalid_existing_value

<br />
<br />

**Spec-level, `onIncompatibleSchemaChange` field | Unsupported value detected | Other**

pr_screenshot-1372-incompatible_schema-spec_level-error-unsupported_value-object


dyaffe:
Thanks, confirming that this looks good to me @kiahna-tucker

src/components/editor/Bindings/OnIncompatibleSchemaChange/SelectorOption.tsx
kiahna-tucker, line 1: Let me know if you would like this component to be moved to a shared location at this time. Given the likelihood that the corresponding, binding-level property will be exposed soon, I am fine doing so now or when that work is picked up.

travjenkins: I think we're good leaving it until later.

src/hooks/OnIncompatibleSchemaChange/useSpecificationIncompatibleSchemaSetting.ts
travjenkins, line 1: I struggled with how to name a hook that is for a component... thoughts on keeping the "components are capitalized" rule within the hooks folder?

src/stores/Binding/slices/IncompatibleSchemaChange.ts
```diff
@@ -0,0 +1,43 @@
...
+    ...getInitialIncompatibleSchemaChangeData(),
+
+    setSingleIncompatibleSchemaChange: (bindingUUID, value) => {
+        set(
+            produce((state: BindingState) => {
+                if (value) {
+                    state.incompatibleSchemaChanges[bindingUUID] = value;
+                } else {
+                    omit(state.incompatibleSchemaChanges, bindingUUID);
+                }
```
travjenkins, line 37: Please move this over to `workflow-utils` 

I know it isn't shared but I think we need to start getting all our "schema updates" into a single location so we can start building out a solid reusable util for schema property updates

src/utils/workflow-utils.ts
```diff
@@ -216,6 +218,20 @@ export const generateTaskSpec = (
...
+                        hasLength(incompatibleSchemaChanges[bindingUUID])
+                    ) {
+                        draftSpec.bindings[
+                            existingBindingIndex
+                        ].onIncompatibleSchemaChange =
+                            incompatibleSchemaChanges[bindingUUID];
+                    } else if (entityType === 'materialization') {
+                        delete draftSpec.bindings[existingBindingIndex]
+                            .onIncompatibleSchemaChange;
+                    }
```
travjenkins, line 233: What is driving storing this outside of `meta` ?

src/utils/workflow-utils.ts
```diff
@@ -242,7 +258,18 @@ export const generateTaskSpec = (
...
                             ...resourceConfig,
                         },
-                    });
+                    };
+
+                    if (
+                        entityType === 'materialization' &&
+                        incompatibleSchemaChanges &&
+                        hasLength(incompatibleSchemaChanges[bindingUUID])
+                    ) {
```
travjenkins, line 267: If we're duplicating this check this it should be stored off so they always stay in sync.

I don't think we van get around checking it so putting it up with `bindingDisabled` and `existingBindingIndex` probably makes sense.

src/hooks/OnIncompatibleSchemaChange/useSpecificationIncompatibleSchemaSetting.ts
```diff
@@ -0,0 +1,69 @@
...
+        () =>
+            draftSpecs.length > 0 && draftSpecs[0].spec ? draftSpecs[0] : null,
+        [draftSpecs]
+    );
+
+    const updateOnIncompatibleSchemaChange = useCallback(
+        async (value: AutoCompleteOption['val'] | undefined) => {
+            if (!mutateDraftSpecs || !draftId || !draftSpec) {
+                logRocketEvent(
+                    `${CustomEvents.INCOMPATIBLE_SCHEMA_CHANGE} : missing critical resources to update draft`,
```
travjenkins, line 31: Don't write sentences in the eventing. These should be in between human and machine readable... like `${CustomEvents.INCOMPATIBLE_SCHEMA_CHANGE}:Missing Fields`.

src/hooks/OnIncompatibleSchemaChange/useSpecificationIncompatibleSchemaSetting.ts
```diff
@@ -0,0 +1,69 @@
...
+                return Promise.resolve();
+            }
+
+            let spec: Schema = draftSpec.spec;
+
+            if (!hasLength(value)) {
+                spec = omit(spec, 'onIncompatibleSchemaChange');
+            } else {
+                spec.onIncompatibleSchemaChange = value;
+            }
```
travjenkins, line 48: Please move this over to `workflow-utils`. Trying to get all these kind of schema/property update-ers in one spot to make writing a single "tool" for this easier.

kiahna-tucker:
Responding to the potential defects you have identified:

* **Issue 1** -- the behavior you are observing is intended but it is worth discussing what the _desired_ behavior should be in this unique case. There are three possible ways this can be handled but only two are valid from my vantage point: remove the binding-level `onIncompatibleSchemaChange` property when restoring a binding that exists in the live specification **or** prefill the binding-level `onIncompatibleSchemaChange` property with its live value when restoring a binding that exists in the live specification. The former is the easiest to implement and aligns with how other binding-level properties are handled but it is, technically, destructive.

* **Issue 2** -- this is a defect that is related to the absence of a verdict on how the binding-level property should be restored.

* **Issue 3** -- this is a defect. Will update this comment once it has been investigated.

* **Issue 4** -- this is a defect. Will update this comment once it has been investigated.

src/components/editor/Bindings/OnIncompatibleSchemaChange/SelectorOption.tsx
kiahna-tucker: This has been moved to a shared location since this PR enveloped the binding-level scope.

src/utils/workflow-utils.ts
kiahna-tucker: Honestly? I was on the fence on whether this piece of state should be part of the binding metadata or be stored in a separate slice. Presently, all materialization-specific binding properties are stored in slices and, for ease of implementation, it made sense for `onIncompatibleSchemaChange` to follow suit. Will remove the new slice entirely and fold the state into the binding store since that _sounds_ like the direction you are leaning toward.

src/hooks/OnIncompatibleSchemaChange/useSpecificationIncompatibleSchemaSetting.ts
kiahna-tucker: Since `addOrRemoveSourceCapture` exists in _src/utils/entity-utils.ts_, I defined a similar function for the specification-level `onIncompatibleSchemaChange` property there as well. It is used in this hook and `generateTaskSpec`.

src/hooks/OnIncompatibleSchemaChange/useSpecificationIncompatibleSchemaSetting.ts
kiahna-tucker: This was an oversight on my end as it was a console log originally. Will update.

src/components/incompatibleSchemaChange/Form.tsx
```diff
@@ -87,7 +88,7 @@ export default function IncompatibleSchemaChangeForm({
             ) : null}
 
             {/* The presence of ListboxComponent in autoCompleteDefaultProps prevents
-                the options from rendering  */}
+                the options from rendering. */}
```
travjenkins, line 91: I think this comment needs replaced - this is now just a TODO to get this `Autocomplete` using the defaults since that `Listbox...` issue was handled in another PR.

src/components/materialization/OnIncompatibleSchemaChange/Form.tsx
```diff
@@ -21,6 +26,8 @@ export default function SpecificationIncompatibleSchemaChangeForm() {
         async (option?: AutoCompleteOption | null) => {
             setFormState({ status: FormStatus.UPDATING, error: null });
 
+            setIncompatibleSchemaChange(option?.val);
```
travjenkins, line 29: Not sure what the best UX here is. If the network call fails to update the draft then this new value is still in the store and rendered to the user as if the value changed.

I think we might need to make a quick `ref` and just set the value back if it fails.

src/utils/workflow-utils.ts
```diff
@@ -253,13 +261,9 @@ export const generateTaskSpec = (
                         },
                     };
 
-                    if (
-                        entityType === 'materialization' &&
-                        incompatibleSchemaChanges &&
-                        hasLength(incompatibleSchemaChanges[bindingUUID])
-                    ) {
+                    if (onIncompatibleSchemaChangeOverridden) {
```
travjenkins, line 264: What is driving us doing these updates manually here and right up above instead of using `addOrRemoveOnIncompatibleSchemaChange`?

src/components/incompatibleSchemaChange/Form.tsx
travjenkins, line 1: Forgot about this one - since we are rendering out invalid values we need to make sure to output objects correctly.

image

src/stores/Binding/Store.ts
```diff
@@ -866,6 +870,22 @@ const getInitialState = (
         );
     },
 
+    setBindingOnIncompatibleSchemaChange: (value) => {
+        set(
+            produce((state: BindingState) => {
+                const bindingUUID = state.currentBinding?.uuid;
```
travjenkins, line 876: We should pass in this value. Otherwise there is a _tiiiiny_ chance we call this function for one uuid and by the time this actually runs it is another uuid.

src/components/materialization/OnIncompatibleSchemaChange/Form.tsx
kiahna-tucker: This is a strange defect. The stored copy of the drafted specification has an `onIncompatibleSchemaChange` value that is equivalent to the selected value despite the fact that `mutate` was not called by the hook specified in `updateServer`.

kiahna-tucker: This issue should be resolved but I am going to leave this thread open in the event you have additional comments.

src/hooks/OnIncompatibleSchemaChange/useSpecificationIncompatibleSchemaSetting.ts
```diff
@@ -44,7 +39,7 @@ export default function useSpecificationIncompatibleSchemaSetting() {
                 return Promise.resolve();
             }
 
-            const spec: Schema = draftSpec.spec;
+            const spec: Schema = cloneDeep(draftSpec.spec);
```
travjenkins, line 42: What is driving cloning here?

travjenkins (approved):
overall last round of changes to code look good

still gonna try to break it as best I can with testing

src/hooks/OnIncompatibleSchemaChange/useSpecificationIncompatibleSchemaSetting.ts
kiahna-tucker: The issue noted in this [comment](https://github.com/estuary/ui/pull/1372#discussion_r1894056386).

src/components/incompatibleSchemaChange/Form.tsx
```diff
@@ -0,0 +1,168 @@
...
+                    <Typography>
+                        {intl.formatMessage(
+                            {
+                                id: 'incompatibleSchemaChange.error.message',
+                            },
+                            {
+                                currentSetting:
+                                    typeof currentSetting === 'string'
+                                        ? currentSetting
+                                        : JSON.stringify(currentSetting),
```
travjenkins, line 75: We should use `stringifyJSON`.

Issue #1371, opened 2024-11-20
Some users get stuck with a page loading when opening capture details


travjenkins:
**Bug**
User is on the dashboard
Goes to the `captures` list
Clicks on a specific capture
Loading spinner shows and does not go away

**Expected**
The page loads like normal
AND
If it does not a proper error page is displayed to them

**Screenshots**
Cannot include here - but they provided them

**Desktop (please complete the following information):**
 - OS: Apple
 - Browser Safari with AdGuard

**Additional context**

PR #1373, opened 2024-11-22
Better handling of import errors

travjenkins:
Stop trying and eventually show an error
Cleaning up old shared stuff and moving to proper location

## Issues

https://github.com/estuary/ui/issues/1371

### 1371

- Handle error in the react component and not the import function to make more reactive
- Only try to reload once and then show an error so people don't get infinite spinners
- Drop wait time just a little bit to speed things up

### Manually tested

- Manually tested with `build` and `preview`

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

After trying to reload once we will throw and error
image


src/components/shared/ErrorImporting/index.tsx
travjenkins, line 1: Moved most of the logic over from the service to the component. Also added a few more LR events for better tracking of this issue.

src/components/shared/ErrorImporting/shared.ts
travjenkins, line 1: Moved this over and removed a lot of the flexibility as they is not being used anywhere else.

src/components/shared/ErrorImporting/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { FallbackProps } from 'react-error-boundary';
+import { getWithExpiry, setWithExpiry } from './shared';
+
+export function ErrorImporting({ error }: FallbackProps) {
+    const [stopTrying, setStopTrying] = useState(false);
+
+    // TODO (error importing) get rid of this logging.
+    //  Keep an eye on the error so we know the list for
+    //  failedToLazyLoad is valid.
+    logRocketConsole('Error Importing Error', error);
```
kiahna-tucker, line 19: Is the repetition of the word _error_ intentional?

kiahna-tucker (approved):

src/components/shared/ErrorImporting/index.tsx
travjenkins: yeah - I'll add a colon... this is actually logging the `name` of the entity and then saying there is an error.

Issue #1409, opened 2024-12-20
Warning when user backfills materialization without storage mapping


dyaffe:
**Proposal**
We have a ton of users that try to backfill a materialization rather than a capture.  This works fine if they have storage mappings but doesn't work well if they don't have them and their collection is more than 20 days old.

It would be nice if we could pop up a warning message in two scenarios:

1. They backfill a collection that's over 20 days old which uses our trial bucket.
- "Your account uses Estuary's Trial bucket which includes 20 days of storage and this collection is older than that.  Data will be missing if you backfill from the materialization so we recommend backfilling from the source."
2. They add a collection that is over 20 days old to a materialization.
- "Your account uses Estuary's Trial bucket which includes 20 days of storage and this collection is older than that.  To ensure you have all data, please also backfill this collection from the source after adding it to the materialization."

jwhartley:
Is it possible to link to the capture that we're suggesting they backfill as part of the message?

travjenkins:
Linking to the capture is a bit outside of this request. Since there could be `n` captures needing backfilled the UX will need to be fleshed out a bit.

kiahna-tucker:
Related child issue(s): https://github.com/estuary/ui/issues/1488.

PR #1440, opened 2025-01-31
Display alert when an old, trial-only binding is added or backfilled

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1409

### 1409

* Extend binding store state to include a top-level, collection-specific metadata property (i.e., `collectionMetadata`). The property is a dictionary type that is keyed by the catalog name of the collection. The following pieces of metadata are currently stored there: `added`, `sourceBackfillRecommended`, `trialOnlyStorage`, and `updatedAt`.

* Create a hook, `useTrialPrefixes`, to evaluate, store, and return prefixes that only have the trial storage mapping defined.

* Create a hook, `useTrialCollections`, to evaluate and return metadata for collections under trial prefixes.

* Display a warning alert in the binding-level _Backfill_ section when the selected binding is over 20 days old, the corresponding collection is under a trial prefix, and it is backfilled in the current workflow.

* Display a warning alert in the top-level _Backfill_ section when the at least one backfilled binding is over 20 days old and the corresponding collection is under a trial prefix.

* Display a warning alert in the right pane of the binding selector when the selected binding is over 20 days old, the corresponding collection is under a trial prefix, and it was added in the current workflow.

* Store the numeric value of the trial duration of 20 days in an environment variable.

* Create component for the alert indicator displayed in the `CollectionConfig` header.

### Manually tested

* Validate that the indicator on the _Source Collections_ section header displays in a warning state at least one binding is over 20 days old, the corresponding collection is under a trial prefix, and backfilled or added in the current workflow.

* Validate that the indicator on a binding list item displays in a warning state when the binding is over 20 days old, the corresponding collection is under a trial prefix, and backfilled or added in the current workflow.

* Validate that the error state of the indicator on a binding list item and the _Source Collections_ header takes precedence over the warning state.

* Validate that a warning alert displays in the top-level _Backfill_ section when at least one binding is over 20 days old, the corresponding collection is under a trial prefix, and backfilled in the current workflow.

* Validate that a warning alert displays in the binding-level _Backfill_ section when the binding is over 20 days old, the corresponding collection is under a trial prefix, and backfilled in the current workflow.

* Validate that a warning alert displays above the _Resource Config_ header when the binding is over 20 days old, the corresponding collection is under a trial prefix, and added in the current workflow.

* Validate that source backfill recommended alerts are preserved when the _Next_ button is clicked and the browser refreshed.

### Automated tests

_N/A_

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

**_Source Collections_ section header when section contains a source backfill recommended alert**

pr_screenshot-1440-trial_collections-backfill-form_section_warning_indication

<br />
<br />

**Binding list item when a backfill is recommended for its source**

pr_screenshot-1440-trial_collections-backfill-selected_binding_warning_indication

<br />
<br />

**Top-level alert | Backfill**

pr_screenshot-1440-trial_collections-backfill-top_level_alert

<br />
<br />

**Binding-level alert | Backfill**

pr_screenshot-1440-trial_collections-backfill-binding_level_alert

<br />
<br />

**Binding-level alert | Added**

pr_screenshot-1440-trial_collections-added-binding_level_alert

<br />
<br />


src/components/editor/Bindings/Backfill/SectionWrapper.tsx
```diff
@@ -1,23 +1,48 @@
...
+
+                {entityType === 'materialization' ? (
+                    <TrialOnlyPrefixAlert
+                        bindingUUID={bindingUUID}
+                        messageId={alertMessageId}
+                        triggered={
+                            bindingUUID
+                                ? backfilledBindings.includes(bindingUUID)
+                                : backfillAllBindings
+                        }
```
kiahna-tucker, line 40: To display an alert in the top-level _Backfill_ section when at least one aged, trial-only binding is backfilled, a hook must be defined which filters the set of resource configs to extract the UUIDs of trial-only bindings and the trigger condition will require comparing the two arrays.

src/components/collection/Config.tsx
```diff
@@ -45,7 +51,7 @@ function CollectionConfig({
         resourceConfigErrorsExist ||
         fullSourceErrorsExist;
 
-    const hasWarnings = bindingErrorsExist;
+    const hasWarnings = bindingErrorsExist || sourceBackfillRecommended;
```
travjenkins, line 54: This is now possibly going over way more stuff. I think it is time to move it into a stand alone component.

src/api/liveSpecsExt.ts
```diff
@@ -429,6 +430,44 @@ const getLiveSpecShards = (tenant: string, entityType: Entity) => {
         .eq('spec_type', entityType);
 };
 
+export interface TrialCollectionQuery {
+    catalog_name: string;
+    updated_at: string;
+}
+
+const getTrialCollections = async (prefixes: string[]) => {
```
travjenkins: Gonna "invert" this approach. Noting here in writing since we discussed on a video chat.

src/components/editor/Bindings/Row/ErrorIndicator.tsx
```diff
@@ -22,14 +23,28 @@ function BindingsSelectorErrorIndicator({ bindingUUID }: Props) {
...
         return (
             <Typography>
                 <WarningCircle
                     style={{
                         marginRight: 4,
                         fontSize: 12,
-                        color: theme.palette.error.main,
+                        color:
+                            bindingErrors?.length > 0 ||
+                            configErrors?.length > 0
```
travjenkins, line 45: I think we could just check `!Boolean(sourceBackfillRecommended)` here

src/lang/en-US/Workflows.ts
travjenkins, line 1: We should try to keep the `20` in a property in the .env file so we can share here and the helper util for checking the time stamps.

src/components/editor/Bindings/Row/ErrorIndicator.tsx
kiahna-tucker: You can. That is what was there initially. I incorrectly assumed you would be opposed to that since you tend to prefer the evaluation of more severe error conditions.

src/stores/Binding/types.ts
```diff
@@ -145,7 +152,8 @@ export interface BindingState
     // and bindings are added to the specification via the collection selector.
     prefillResourceConfigs: (
         targetCollections: string[],
-        disableOmit?: boolean
+        disableOmit?: boolean,
+        trackAddition?: boolean
```
travjenkins, line 156: `trackAddition` seems to always be getting passed as `true`... we might not need to pass it anymore?

src/stores/Binding/Store.ts
```diff
@@ -139,6 +142,13 @@ const hydrateSpecificationDependentState = async (
         );
     }
 
+    const trialCollections = await evaluateTrialCollections(
+        Object.keys(get().bindings),
+        getTrialOnlyPrefixes
+    );
+
+    get().setCollectionMetadata(trialCollections);
```
travjenkins, line 150: I don't think this belongs in here as it is not relying on the draft spec at all

kiahna-tucker: It is hydrating binding dependent state.

src/stores/Binding/types.ts
kiahna-tucker: I was on the fence as to whether it is _safe_ to let this be the default behavior of `prefillResourceConfigs`. It definitely can be removed today, but will we forget this behavior exists in the future if removed.

kiahna-tucker: This input argument was removed from this store action.

src/stores/Binding/shared.ts
```diff
@@ -88,6 +89,22 @@ export const sortResourceConfigs = (
     return sortedResources;
 };
 
+export const updateCollectionMetadata = (
```
travjenkins, line 92: We should remove this

src/stores/Binding/Store.ts
```diff
@@ -945,22 +811,19 @@ const getInitialState = (
...
 
-                    updateCollectionMetadata(state, catalog_name, {
+                    state.collectionMetadata[catalog_name] = {
                         added,
                         sourceBackfillRecommended:
-                            triggered && isBeforeTrialInterval(updated_at),
+                            isBeforeTrialInterval(updated_at) &&
+                            (added ||
+                                backfilledCollections.includes(catalog_name)),
                         trialStorage: true,
```
travjenkins, line 824: Can you throw in a comment why this is hardcoded to `true`? I _think_ I remember us talking about this but cannot recall.

src/stores/Binding/Store.ts
```diff
@@ -532,6 +387,17 @@ const getInitialState = (
...
+
+        return {
+            addedCollections: draftedBindings
+                ? difference(
+                      draftedBindings.map((binding) =>
+                          getCollectionName(binding)
+                      ),
+                      liveBindings.map((binding) => getCollectionName(binding))
+                  )
+                : [],
```
travjenkins, line 399: We don't need to change anything here... but we could _maybe_ make this a bit eaiser on processing since if we are here we have already done through at least one of these arrays. However, it could end up getting more complex to "cache" off one of these lists while supporting multiple cases.

travjenkins: Actually I guess we always know we have a `drafterBindings` here... so potentially worth it to make a shared array that is built up in the loop above. But not sure how that would impact performance with the `set`

src/components/editor/Bindings/Row/ErrorIndicator.tsx
travjenkins, line 1: I think there is a bug. If I manually backfill a few items and then click on `backfill all` twice so that nothing is being backfilled then the warning icons still show.

image

---

Also when using `backfill all` to first backill none of the inline warnings show up
image

src/stores/Binding/Store.ts
```diff
@@ -765,6 +798,39 @@ const getInitialState = (
         );
     },
 
+    setCollectionMetadata: (values, addedCollections) => {
+        if (!hasLength(values)) {
+            return;
+        }
```
travjenkins, line 804: I think we need a way to handle this to be used to clear out the settings when all the backfills are removed.

Was this here just to be safe or are we relying on this for something?

kiahna-tucker: No, this is very important and remember there are multiple triggers for the alert at hand and often the action is called over a subset of collections.

src/components/editor/Bindings/Backfill/BackfillButton.tsx
```diff
@@ -151,6 +157,16 @@ function BackfillButton({
...
                         setBackfilledBindings(increment, targetBindingUUID);
+
+                        evaluateTrialCollections(
+                            bindingMetadata.map(({ collection }) => collection)
+                        ).then(
+                            (response) => {
+                                setCollectionMetadata(response, []);
+                            },
+                            () => {}
+                        );
```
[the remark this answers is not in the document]
travjenkins: Yup - that is what I was thinking.

src/components/editor/Bindings/Backfill/useUpdateBackfillCounter.ts
```diff
@@ -112,6 +125,14 @@ function useUpdateBackfillCounter() {
                                     backfilled && increment === 'false';
 
                                 if (shouldIncrement || shouldDecrement) {
+                                    increment === 'true'
```
travjenkins, line 128: Since we need to handle this logic in two places... I think we could just store off which array to populate by storing the key up above before we start looping.

travjenkins (approved):
lgtm

Issue #1443, opened 2025-02-03
Marketplace Billing Updates


dyaffe:
**Background**
We currently have a billing integration with GCP and AWS, but don't signify in the UI that an account is paying through them.

**Proposal**
1. Add additional payment provider options besides "Stripe" and "External" -- "AWS Marketplace" and "GCP Marketplace"
2. When they're selected we should replace "Cloud Tier" in the billing UI with "AWS Marketplace" and "GCP Marketplace" respectively

travjenkins:
I know we have the `gcm` account column. Are we adding one for `aws` as well?

dyaffe:
Yes, that would be great if possible.  We would update it manually.

PR #1444, opened 2025-02-03
Let the users know if they are using marketplace for payments

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1443

### 1443

-  Check for the marketplace stuff when checking external payment being used
- Add new content to let the user know they are using GCP
- Check if `not stripe` in case we end up expanding column
- Remove `.` from end of all short payment messages for billing

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

### GCP
image

### DIFF
--- a/src/components/admin/Billing/PricingTierDetails.tsx
+++ b/src/components/admin/Billing/PricingTierDetails.tsx
@@ -7,22 +7,33 @@ import { useTenantStore } from 'stores/Tenant/Store';
 
 function PricingTierDetails() {
     const selectedTenant = useTenantStore((state) => state.selectedTenant);
-    const externalPaymentMethod = useTenantUsesExternalPayment(selectedTenant);
+    const [externalPaymentMethod, marketPlaceProvider] =
+        useTenantUsesExternalPayment(selectedTenant);
 
     const billingStoreHydrated = useBillingStore((state) => state.hydrated);
     const paymentMethodExists = useBillingStore(
         (state) => state.paymentMethodExists
     );
 
-    const messageId = useMemo(
-        () =>
-            externalPaymentMethod
-                ? 'admin.billing.message.external'
-                : paymentMethodExists
-                ? 'admin.billing.message.paidTier'
-                : 'admin.billing.message.freeTier',
-        [externalPaymentMethod, paymentMethodExists]
-    );
+    const messageId = useMemo(() => {
+        if (externalPaymentMethod) {
+            if (marketPlaceProvider === 'gcp') {
+                return 'something goes here';
+            }
+
+            if (marketPlaceProvider === 'aws') {
+                return 'something goes here';
+            }
+
+            return 'admin.billing.message.external';
+        }
+
+        if (paymentMethodExists) {
+            return 'admin.billing.message.paidTier';
+        }
+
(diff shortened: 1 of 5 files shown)


src/context/fetcher/TenantBillingDetails.tsx
```diff
@@ -66,17 +66,28 @@ export const useTenantBillingDetails = () => {
...
+        }
+
+        const selectedTenantDetails = tenantBillingDetails.find(
+            (tenantBillingDetail) =>
+                tenantBillingDetail.tenant === selectedTenant
+        );
+
+        const marketplaceProvider = selectedTenantDetails?.gcm_account_id
+            ? 'gcp'
+            : null; //add aws support
```
kiahna-tucker, line 86: Is the comment on line 86 a future TODO or in scope for this branch? If the former, could you mark it as such?

src/components/admin/Billing/PricingTierDetails.tsx
```diff
@@ -7,22 +7,33 @@ import { useTenantStore } from 'stores/Tenant/Store';
...
+        if (externalPaymentMethod) {
+            if (marketPlaceProvider === 'gcp') {
+                return 'admin.billing.message.external.gcp';
+            }
+
+            if (marketPlaceProvider === 'aws') {
+                return 'admin.billing.message.external.aws';
+            }
+
+            return 'admin.billing.message.external';
```
kiahna-tucker, line 28: This could be simplified to one of the following to increase scalability:

```ts
// If the ID of the generic content had an identifying suffix.
return `admin.billing.message.external.${marketPlaceProvider ?? 'generic }`;
```

OR

```ts
return marketPlaceProvider ? `admin.billing.message.external.${marketPlaceProvider}` : 'admin.billing.message.external';
```

src/context/fetcher/TenantBillingDetails.tsx
```diff
@@ -66,17 +66,28 @@ export const useTenantBillingDetails = () => {
     return useContext(TenantContext);
 };
 
-export const useTenantUsesExternalPayment = (selectedTenant: string) => {
+export const useTenantUsesExternalPayment = (
+    selectedTenant: string
+): [boolean, string | null] => {
```
kiahna-tucker, line 71: Should the string type be narrowed to the set of supported providers since its value is, ultimately, determined by the client?

kiahna-tucker (commented):
There is one _critical_ [comment](https://github.com/estuary/ui/pull/1444#discussion_r1941259563) that needs to be addressed before an approval can be given.

src/components/admin/Billing/PricingTierDetails.tsx
travjenkins: unless it is deeply important we should try not to dynamically build translate strings.

src/context/fetcher/TenantBillingDetails.tsx
travjenkins: It will eventually for sure

kiahna-tucker (approved):

Issue #1445, opened 2025-02-04
When enabling or disabling a binding, we should display that resource


dyaffe:
**Background**
When editing a capture or materialization, if a user enables (or disables) a binding that isn't currently enabled, we don't select it in the editor on the right side.  That can be pretty confusing and it would be great to automatically select it when that action is taken.

**2nd thing**
~~When a user searches for a specific binding, it would be nice to show the first one in the filtered list automatically.~~ 
moved over to https://github.com/estuary/ui/issues/1426

PR #1446, opened 2025-02-04
Toggling binding disable/enable selects the binding

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1445

### 1445

- No longer stopping event bubbling
- Minor refactor to make function more usable
- Adding the `disable` field to select the binding

### Manually tested

-  Toggled and clicked around in the binding selector

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

N/A

### DIFF
--- a/src/components/collection/Selector/List/index.tsx
+++ b/src/components/collection/Selector/List/index.tsx
@@ -161,6 +161,28 @@ function CollectionSelectorList({
         [isCapture]
     );
 
+    const selectBinding = useCallback(
+        (id: any) => {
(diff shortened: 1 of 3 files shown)


kiahna-tucker (approved):
Approved while testing is underway. This change is seemingly innocuous so I don't foresee any issues that could be caused by it. I would still like to test it but I am comfortable with the PR being merged if it is needed ASAP.

Issue #1470, opened 2025-02-17
Break Up IPV4 and IPV6 addresses


dyaffe:
**Background**
Users only need to whitelist either our IPV4 or IPV6 addresses for a data plane.  As such, we should display them separately and make it obvious that they only need to whitelist one of them.

**Proposal**
Break out "IP Whitelisting" under the "Data Planes" section of the admin UI for each data plane.

Under it text should say "All traffic from this data plane is dual stack such that either the IPV4 or IPV6 can be whitelisted".

Then show IPV4's and IPV6's as separate copy/paste-able options.

travjenkins:
Make sure each option can be copied on their own

travjenkins:
We'll also handle showing the table to normal users with this

PR #1474, opened 2025-02-25
Show users data plane details and split up cidr blocks

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1470

### 1470

- Split apart the IPv4 and IPv6 in the cidr block arrays
- Allow user to select which version to see/copy
- Save off selection to local storage
- Show all users the `Data Planes` section in admin

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

### Admin table public ipv4
image

### Admin table public ipv6
image

### Config shows ipv4
image

### config showing ipv6
image


src/components/admin/Settings/index.tsx
```diff
@@ -44,13 +39,11 @@ function Settings() {
                 <StorageMappings />
             </Stack>
 
-            {hasSupportRole ? (
-                <Stack>
-                    <Divider sx={{ mt: 2, mb: 4 }} />
+            <Stack>
```
travjenkins, line 42: We'll just always show to users now

src/components/shared/CopyCidrBlocks.tsx
travjenkins, line 1: Previously rendering cidr blocks was simple. This isn't too complex but best to keep it in a single place since we render these in the table and the config forms.

Some stuff is hardcoded - but I think it is fine since IPs do not change _too often_

src/context/DataPlaneScopeContext.tsx
```diff
@@ -1,21 +1,33 @@
...
 const DataPlaneScopeContext = createContext<DataPlaneScope | null>(null);
 
 const DataPlaneScopeContextProvider = ({ children }: BaseComponentProps) => {
-    const [dataPlaneScope, setDataPlaneScope] =
-        useState<DataPlaneScopes>('private');
+    const [defaultDataPlane, setDefaultDataPlane] =
+        useLocalStorage<DataPlaneScopes>(
+            LocalStorageKeys.ADMIN_DATAPLANE_CHOICE,
+            defaultOption
+        );
```
travjenkins, line 20: Trying to make the UX better and not make users have to constantly switch between public and private. Eventually - this should be a setting stored with the select tenant.

src/lang/en-US/JsonForms.ts
```diff
@@ -13,7 +13,8 @@ export const JsonForms: Record<string, string> = {
     'dateTimePicker.picker.footer': `Timezone: UTC`,
     'datePicker.button.ariaLabel': `Open date picker for {label}`,
     'informational.sshEndpoint.title': `All Estuary traffic comes from a group of IPs that can be whitelisted:`,
-    'informational.sshEndpoint.ip': `34.121.207.128, 35.226.75.135, 34.68.62.148`,
+    'informational.sshEndpoint.ipv4': `34.121.207.128, 35.226.75.135, 34.68.62.148`,
+    'informational.sshEndpoint.ipv6': `does not support IPv6`,
```
travjenkins, line 17: Very rare this would ever hit for something not related to the public data plan - but still kept the message open ended.

src/lang/en-US/JsonForms.ts
```diff
@@ -13,7 +13,8 @@ export const JsonForms: Record<string, string> = {
     'dateTimePicker.picker.footer': `Timezone: UTC`,
     'datePicker.button.ariaLabel': `Open date picker for {label}`,
     'informational.sshEndpoint.title': `All Estuary traffic comes from a group of IPs that can be whitelisted:`,
-    'informational.sshEndpoint.ip': `34.121.207.128, 35.226.75.135, 34.68.62.148`,
+    'informational.sshEndpoint.ipv4': `34.121.207.128, 35.226.75.135, 34.68.62.148`,
```
travjenkins, line 16: Only the key changed

Issue #1468, opened 2025-02-14
Cannot click `back` while on `materialization` `create`


travjenkins:
**Bug**
Enter materialization create
select a connector
land on create page
click back
land on create page again

**Expected**
go back to connector select page

**Screenshots**
n/a

**Desktop (please complete the following information):**
n/a

**Additional context**
n/a

PR #1479, opened 2025-02-26
Travjenkins/bug/cannot click back on mat create

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1468

### 1468

-  always set `advanceToForm` to true

### misc

-  cleaned up the old options to be an object
- removed a setting that was no longer used

### Manually tested

- created materialization and then clicked back

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

### After clicking back on materialization create
image

### DIFF
--- a/src/components/shared/Entity/hooks/useEntityCreateNavigate.ts
+++ b/src/components/shared/Entity/hooks/useEntityCreateNavigate.ts
@@ -7,19 +7,20 @@ import { ENTITY_SETTINGS } from 'settings/entity';
 import { EntityWithCreateWorkflow } from 'types';
 import { getPathWithParams, hasLength } from 'utils/misc-utils';
 
+export interface HookEntityCreateNavigateProps {
+    advanceToForm?: boolean;
+    dataPlaneId?: string | null;
+    id?: string | null | undefined;
+}
+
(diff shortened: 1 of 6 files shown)


src/components/shared/Entity/hooks/useEntityCreateNavigate.ts
```diff
@@ -7,19 +7,20 @@ import { ENTITY_SETTINGS } from 'settings/entity';
 import { EntityWithCreateWorkflow } from 'types';
 import { getPathWithParams, hasLength } from 'utils/misc-utils';
 
+export interface HookEntityCreateNavigateProps {
+    advanceToForm?: boolean;
+    dataPlaneId?: string | null;
+    id?: string | null | undefined;
+}
```
travjenkins, line 14: `replace` was not being used so I removed it. It may make it harder to track back the original settings if we want to put it back in. However, I'd rather not keep it hanging around.

kiahna-tucker (approved):
Approved with minimal testing. Confirmed that the browser navigation buttons prompt and traverse application workflows as desired.

Issue #1488, opened 2025-03-04
Adding a collection via source capture warns the user about previously added collections


dyaffe:
**Bug**
For accounts using the Estuary trial bucket, if you add a collection to your materialization by adding a "Source Capture", we'll currently warn about needing to backfill previously added collections.

**Expected**
We should never warn about needing to backfill a collection that was previously part of the spec.  

cc @kiahna-tucker 

<img width="1736" alt="Image" src="https://github.com/user-attachments/assets/e6c9cb3b-1cff-4f91-a975-5ffe2544de3c" />

dyaffe:
I think it may actually be warning on the opposite of what we want -- collections that were previously there and not ones that were added.

kiahna-tucker:
> I think it may actually be warning on the opposite of what we want -- collections that were previously there and not ones that were added.

This warning will be displayed for any associated collection that is under a trial-only tenant, updated greater than 20 days ago, and added in the context of this workflow. It sounds like the addition-related condition of this trigger must be narrowed: only bound collections that are _new_ to the task should be eligible for this warning.

dyaffe:
It isn't actually doing that though.  Some collections are omitted from the warning (see that screenshot).

Yes, you are correct though...the feature would be much more useful if it only warned about newly added collections.

kiahna-tucker:
Was this a session of yours? If the collection was not added in the context of the current workflow it would not be flagged. Looking at the collections discovered by the linked capture, the two collections believed to be an indication of an error in the screenshot are not written to by the linked capture.

PR #1489, opened 2025-03-04
Present source backfill recommended warning for newly bound collections

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1488

### 1488

* Extend the collection metadata object to include an indication of whether the collection is bound to the live task.

* Narrow the client's understanding of an _added_ collection. The `added` collection metadata property now reflects whether the collection is newly introduced to the task specification.

### Manually tested

### Automated tests

_N/A_

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

_N/A_

### DIFF
--- a/src/stores/Binding/Store.ts
+++ b/src/stores/Binding/Store.ts
@@ -360,11 +360,6 @@ const getInitialState = (
                             targetBindingIndex
                         );
 
-                        state.collectionMetadata[collection].previouslyBound =
-                            liveBindingIndex > -1;
-
-                        state.collectionMetadata;
-
                         const liveBackfillCounter =
                             liveBindingIndex > -1
                                 ? getBackfillCounter(
@@ -378,6 +373,23 @@ const getInitialState = (
                                 draftedBackfillCounter > 0)
                         ) {
                             state.backfilledBindings.push(UUID);
+                        }
+
+                        console.log(collection);
+                        console.log(liveBindingIndex);
+
+                        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+                        if (state.collectionMetadata?.[collection]) {
+                            state.collectionMetadata[
+                                collection
+                            ].previouslyBound = liveBindingIndex > -1;
+                        } else {
+                            state.collectionMetadata = {
+                                ...state.collectionMetadata,
+                                [collection]: {
+                                    previouslyBound: liveBindingIndex > -1,
+                                },
+                            };
                         }
                     }
                 });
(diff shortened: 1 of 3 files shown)


src/stores/Binding/Store.ts
```diff
@@ -374,6 +374,20 @@ const getInitialState = (
                         ) {
                             state.backfilledBindings.push(UUID);
                         }
+
+                        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+                        if (state.collectionMetadata?.[collection]) {
+                            state.collectionMetadata[
+                                collection
+                            ].previouslyBound = liveBindingIndex > -1;
```
kiahna-tucker, line 382: This condition only exists as a safeguard in the event this action is called outside of hydration.

src/stores/Binding/Store.ts
```diff
@@ -714,6 +732,33 @@ const getInitialState = (
...
+                        )
+                        .map(
+                            ([_uuid, resourceConfig]) =>
+                                resourceConfig.meta.collectionName
+                        );
+
+                    resetPartialCollectionMetadata(state, evaluatedCollections);
+                } else {
+                    resetPartialCollectionMetadata(state);
+                }
```
travjenkins, line 755: In this scenario I think I would be better to just have a single call of `resetPartialCollectionMetadata`.

src/stores/Binding/shared.ts
```diff
@@ -43,6 +43,33 @@ export const getEnabledCollectionNames = (
         .map(({ meta }) => meta.collectionName);
 };
 
+export const resetPartialCollectionMetadata = (
```
travjenkins, line 46: This is an interesting one. I think it is fine for now - but was confusing upon first reading it as most `reset...` functions use a default getter. However, `collectionMetadata` doesn't have or need one.

We'll need to eventually think how we want to handle defaults for transient state settings.

src/stores/Binding/shared.ts
```diff
@@ -43,6 +43,33 @@ export const getEnabledCollectionNames = (
...
+            }
+        });
+
+        return;
+    }
+
+    Object.keys(state.collectionMetadata).forEach((collection) => {
+        state.collectionMetadata[collection].added = false;
+        state.collectionMetadata[collection].sourceBackfillRecommended = false;
+    });
```
travjenkins, line 70: I kind of hate for even mentioning this... but I feel this is a spot where recursion might actually help. That way we don't have to keep the two things setting values in sync.

travjenkins (approved):
latest lgtm - static analysis

Issue #1503, opened 2025-03-11
Update field selection table row actions


kiahna-tucker:
**Last Updated:** Mar. 19, 2025

<hr />

A field selection table row should have three actions: select, require, and exclude.

Below are the base requirements for this feature:

* Replace _Include_ row action with _Require_ and _Select_ row actions.

* Order the row actions as follows starting from left to right: select, require, and exclude.

* Apply the styling of the former _Include_ row action to the new  _Select_ row action.

* Use yellow/orange as the base color of the new _Require_ row action.

* Initialize a field according to the following logic:

    * A field is _selected_ if the `recommended` flag is true and the field is not listed in the `include`/`require` object or `exclude` array in the binding's `fields` stanza.
    
    * A field is _required_ if is listed in the `include`/`require` object in the binding's `fields` stanza.

    * A field is _excluded_ if is listed in the `exclude` array in the binding's `fields` stanza.
    
## Design

**Last Updated:** Mar. 11, 2025

The first, mocked row in each image corresponds to the row actions available when the `recommended` flag is false. The other mocked rows in each image correspond to the row actions available when the `recommended` flag is true.

**No row actions permanently disabled**

This applies to the following constraint types: `LOCATION_RECOMMENDED`, `FIELD_OPTIONAL`.

<img width="205" alt="Image" src="https://github.com/user-attachments/assets/ae3f7859-ab77-4fe1-a74c-b01c05495f0e" />

<br />
<br />

**_Require_ row action permanently disabled**

This applies to the following constraint types: `FIELD_FORBIDDEN`.

<img width="203" alt="Image" src="https://github.com/user-attachments/assets/3e7ec279-f078-4652-ad8d-7f956f97fc66" />

<br />
<br />

**_Exclude_ row action permanently disabled**

This applies to the following constraint types: `FIELD_REQUIRED`, `LOCATION_REQUIRED`.

<img width="202" alt="Image" src="https://github.com/user-attachments/assets/2aa3ced2-4b7a-4075-9e06-706b1d06ab63" />

<br />
<br />

**NOTE:** No row actions are presented for fields with the `UNSATISFIABLE` constraint type.

kiahna-tucker:
## Questions

[...]

PR #1506, opened 2025-03-11
Update field selection table row actions

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1503

https://github.com/estuary/ui/issues/1533

### 1503

* Replace _Include_ row action with _Require_ and _Select_ row actions.

* Order the row actions as follows starting from left to right: select, require, and exclude.

* Apply the styling of the former _Include_ row action to the new  _Select_ row action.

* Use yellow/orange as the base color of the new _Require_ row action.

* Initialize a field according to the following logic:

    * A field is _selected_ if the `recommended` flag is true and the field is not listed in the `include`/`require` object or `exclude` array in the binding's `fields` stanza.
    
    * A field is _required_ if is listed in the `include`/`require` object in the binding's `fields` stanza.

    * A field is _excluded_ if is listed in the `exclude` array in the binding's `fields` stanza.

* Prevent a draft update when the selection mode does not change for a **single** field.

### 1533

* Move interfaces, variables, and functions defined outside of components in field selection component files to `types.ts` and `shared.ts` files. The following components were touched: `TableColumnSelector`, `Rows`, `Row`, and `FieldSelectionTable`.
    
### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

**Field selection initialized | `recommended` true | No existing selections**

pr_screenshot-1506-field_actions-recommended_true-init_init-fop_lreq_lrec-reorder_2

<br />
<br />

**Field selection initialized | `recommended` true | Existing selections**

pr_screenshot-1506-field_actions-recommended_true-init-fop_lreq_lrec-reorder_2

<br />
<br />

**Field selection initialized | `recommended` true | Tooltip on disabled _Exclude_ row action**

pr_screenshot-1506-field_actions-recommended_true-exclude_tooltip-fop_lreq_lrec-reorder_2

<br />
<br />

**Field selection initialized | `recommended` false | No existing selections**

pr_screenshot-1506-field_actions-recommended_false-init-fop_lreq_lrec-reorder_2

<br />
<br />

**Field selection initialized | `recommended` false | Existing selections**

pr_screenshot-1506-field_actions-recommended_false-existing_selections-fop_lreq_lrec-reorder_2

<br />
<br />


src/components/tables/cells/fieldSelection/FieldActionButton.tsx
travjenkins, line 1: Not for this PR - but for the field selection overhaul in general.

I think we should switch all these functions over to objects we just store the settings in.

travjenkins (commented):
static code looks good - testing now

src/components/tables/cells/fieldSelection/FieldActionButton.tsx
kiahna-tucker: This thought crossed my mind as well.

src/components/tables/cells/fieldSelection/FieldActions.tsx
```diff
@@ -30,8 +30,8 @@ const evaluateSelectionType = (
     recommended: boolean,
     toggleValue: FieldSelectionType,
     selectedValue: FieldSelectionType | null,
-    singleValue: FieldSelectionType | null
-) => (selectedValue === toggleValue && recommended ? 'default' : singleValue);
+    targetValue: FieldSelectionType | null
+) => (selectedValue === toggleValue && recommended ? 'default' : targetValue);
```
travjenkins, line 34: Can we toss a call to LR here so we can keep track.

travjenkins (approved):
lgtm

src/components/tables/cells/fieldSelection/FieldActionButton.tsx
travjenkins: I'd say the spirit of this is now generally done with the new `alpha` colors.

src/components/shared/buttons/OutlinedToggleButton.tsx
```diff
@@ -1,85 +1,49 @@
...
+            ? {
+                  backgroundColor: theme.palette[colorKey].alpha_05,
+                  border: `1px solid ${theme.palette[colorKey].alpha_12}`,
+                  color: theme.palette[colorKey].alpha_26,
+              }
+            : {
+                  backgroundColor: 'none',
+                  border: defaultOutline[theme.palette.mode],
+                  color: disabledButtonText[theme.palette.mode],
+              },
```
travjenkins, line 45: What is the case where `selected` is true but the class for selected is not applied? Hoping we could get the same styling with more static styling.

src/components/shared/buttons/OutlinedToggleButton.tsx
travjenkins, line 1: This should eventually live somewhere else. We probably want to put all the `styled` stuff in a single spot near the theme (and break that thing up). We should also probably just use our own ToggleButton that can be outlined or not and then stop using the MUI one all together.

src/components/tables/cells/fieldSelection/FieldActions.tsx
```diff
@@ -73,21 +77,48 @@ function FieldActions({ bindingUUID, field, constraint }: Props) {
...
+                            singleValue
+                        );
+
+                        setSingleSelection(
+                            bindingUUID,
+                            field,
+                            selectionType,
+                            selection?.meta
+                        );
+                    }}
```
travjenkins, line 100: I think all these `onClicks` should be shared.

travjenkins: They should also probably use the `value` from the function args.

src/components/tables/cells/fieldSelection/FieldActions.tsx
```diff
@@ -73,21 +77,48 @@ function FieldActions({ bindingUUID, field, constraint }: Props) {
...
+                    value="default"
+                >
+                    {intl.formatMessage({
+                        id: 'fieldSelection.table.cta.selectField',
+                    })}
+                </OutlinedToggleButton>
+
+                <OutlinedToggleButton
+                    color="warning"
+                    selected={selection?.mode === 'require'}
```
travjenkins, line 112: Can we use the `value' prop on the group so we don't have to manage this ourselves?

src/components/graphs/DataByHourGraph/StatTypeSelector.tsx
```diff
@@ -33,7 +28,7 @@ function StatTypeSelector() {
             >
                 <FormattedMessage id="data.docs" />
```
travjenkins, line 29: While we're updating this can you switch this out to use `intl`

src/components/shared/OutlinedToggleButtonGroup.tsx
```diff
@@ -0,0 +1,29 @@
+import {
+    styled,
+    ToggleButtonGroup,
+    ToggleButtonGroupProps,
+} from '@mui/material';
+
+const OutlinedToggleButtonGroup = styled(ToggleButtonGroup)<
+    ToggleButtonGroupProps & { buttonSelector?: string }
+>(({ buttonSelector }) => ({
+    [`&${buttonSelector ?? ' .MuiToggleButton-root'}`]: {
```
travjenkins, line 10: I am thinking that if we allow a `selector` to be passed in then the person doing it needs to handle nesting themselves. I would say requiring an entire string is best.

So the `&` should be passed in.

src/components/tables/cells/fieldSelection/FieldActionButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+                    { id: 'fieldSelection.table.tooltip.disabledRowAction' },
+                    {
+                        reason: intl.formatMessage({
+                            id: getConstraintMessageId(constraint.type),
+                        }),
+                    }
+                )}
+                placement={tooltipPlacement}
+            >
+                <span className="toggle-button">
```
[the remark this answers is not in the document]
travjenkins: The other idea could be that we just hardcode the styling to support the two options and don't worry about passing it in. I'll leave it up to you.

src/components/shared/buttons/OutlinedToggleButton.tsx
```diff
@@ -23,10 +23,14 @@ const OutlinedToggleButton = styled(ToggleButton)(({
...
-            'color': theme.palette[colorKey].dark,
+            'color':
+                theme.palette.mode === 'light'
+                    ? theme.palette[colorKey].dark
+                    : theme.palette[colorKey].main,
             '&:hover': {
                 border: `1px solid ${
-                    colorKey === 'success'
+                    theme.palette.mode === 'light' &&
+                    (colorKey === 'success' || colorKey === 'info')
```
travjenkins, line 33: Fine with two - but I think these are usually best handles with `array` and then using `includes`. So we can easily add and remove items from the list.

src/components/shared/buttons/OutlinedToggleButton.tsx
```diff
@@ -23,10 +23,14 @@ const OutlinedToggleButton = styled(ToggleButton)(({
         [`&.${toggleButtonClasses.selected}`]: {
             'backgroundColor': theme.palette[colorKey].alpha_12,
             'border': `1px solid ${theme.palette[colorKey].alpha_50}`,
-            'color': theme.palette[colorKey].dark,
+            'color':
+                theme.palette.mode === 'light'
+                    ? theme.palette[colorKey].dark
+                    : theme.palette[colorKey].main,
             '&:hover': {
                 border: `1px solid ${
```
travjenkins, line 31: We eventually might need to look at not using the shorthand settings when doing more complex settings and just break the properties up.

src/components/tables/cells/fieldSelection/FieldActionButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { FieldActionButtonProps } from './types';
+
+export default function FieldActionButton({
+    bindingUUID,
+    constraint,
+    disabled,
+    field,
+    labelId,
+    selection,
+    tooltipPlacement,
```
travjenkins, line 16: This is fine for now - but I think if we allow one prop we should probably just go ahead and allow all props to be passed in.

We'd still _probably_ spread it into the component so `title` could not be overwritten but that way we never have to update the component to expand capability.

src/components/tables/cells/fieldSelection/shared.ts
travjenkins, line 1: What is driving keeping these in functions and not access a static object?

src/context/Theme.tsx
```diff
@@ -146,39 +190,40 @@ const lightMode: PaletteOptions = {
     },
     contrastThreshold,
     error: {
-        alpha_05: alpha(errorMain, 0.05),
-        alpha_12: alpha(errorMain, 0.12),
-        alpha_26: alpha(errorMain, 0.26),
-        alpha_50: alpha(errorMain, 0.5),
+        alpha_05: `rgba(${rgbTranslations.light.error.main}, 0.05)`,
```
travjenkins, line 193: Noting down what was said during dailies.

We'll want to eventually make something that helps generate the `rgba` string.

travjenkins (commented):
a few small tweaks needed

src/components/shared/buttons/OutlinedToggleButton.tsx
kiahna-tucker: The disabled styling is applied last and, thus, it overrides the selected styling.

travjenkins: How? They have two different classes.

src/components/tables/cells/fieldSelection/shared.ts
kiahna-tucker: The benefit of these functions is the unknown constraint type handling. I replaced the functions with a constraint message id object so you can see the difference. I am fine either way.

travjenkins (commented):
Looks good besides the one prop thingy

src/components/shared/OutlinedToggleButtonGroup.tsx
```diff
@@ -0,0 +1,29 @@
+import {
+    styled,
+    ToggleButtonGroup,
+    ToggleButtonGroupProps,
+} from '@mui/material';
+
+const OutlinedToggleButtonGroup = styled(ToggleButtonGroup)<
+    ToggleButtonGroupProps & { buttonSelector?: string }
+>(({ buttonSelector }) => ({
```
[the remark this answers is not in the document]
kiahna-tucker: Ah, good catch. Thanks!

Issue #1520, opened 2025-03-20
Sometimes `create` will not load


travjenkins:
**Bug**
Got on a call with a customer and saw in their network panel that no call to `connector_tags` was being made during the loading of `capture create`

**Expected**
The page loads correctly

**Screenshots**
Include a screenshot of what you saw when the bug happened.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**

PR #1521, opened 2025-03-20
[PROD] - add some extra logging and minor logic tweaks for loading create

travjenkins:
Do not load past details hydrator until it is done Using a ref in hopes it reduces the chance of renders

## Issues

https://github.com/estuary/ui/issues/1520

### 1520

-  Adding some LR logging to see if that can help
- No longer loading endpoint until details is fully hydrated. Hoping this will reduce the chance of something odd happening during loading. Also, all the other hydrators are dependent on that one so waiting is not a huge deal.
- 

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

### DIFF
--- a/src/stores/EndpointConfig/Hydrator.tsx
+++ b/src/stores/EndpointConfig/Hydrator.tsx
@@ -1,6 +1,7 @@
 import { useEntityType } from 'context/EntityContext';
 import { useEntityWorkflow } from 'context/Workflow';
-import { useEffect, useState } from 'react';
+
+import { useEffect, useRef } from 'react';
 import { logRocketConsole } from 'services/shared';
 import { useDetailsFormStore } from 'stores/DetailsForm/Store';
 import {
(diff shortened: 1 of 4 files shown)


src/stores/EndpointConfig/Hydrator.tsx
```diff
@@ -16,7 +17,7 @@ export const EndpointConfigHydrator = ({ children }: BaseComponentProps) => {
     const entityType = useEntityType();
     const workflow = useEntityWorkflow();
 
-    const [runHydration, setRunHydration] = useState(true);
+    const runHydration = useRef(true);
```
travjenkins, line 20: This never needed to be state. Doubt it causes any issues - but better to be safe.

Issue #437, opened 2023-01-05
Able to enter the edit workflow for read-only tasks


kiahna-tucker:
## Bug

A user with read-only capabilities to a specific tenant is able to enter the edit workflow for its tasks. A task that falls under a read-only tenant will have a row in the corresponding task table that behaves no differently than tasks belonging to write-enabled tenants. This means that the options menu for that table row will contain an _edit specification_ option, granted an edit workflow exists[^1]. Selecting the _edit specification_ option will transition a user into the edit workflow for that task.

Since the user does not have the write capability for this tenant, a field-level error will appear in the details form section that indicates there is a problem with the catalog name. Despite being able to edit the _Endpoint Config_ and _Output Collections_ sections of the form, this error prevents the submission of the form when the _Next_ CTA is clicked.

[^1]: An edit workflow does not exist for collections in the UI therefore, while derivations are _tasks_, it should be explicitly noted that this bug does not impact them at this point in time.

### Recreation Steps:

This is easy to recreate in a standard, local environment. The steps below will outline a simple way to reproduce this bug locally, but can be broadened to be reproduced in various environments.

1. Log into the application as a user that is **only** granted the `read` capability to the `/ops` tenant.

1. Navigate to the _Materializations_ page.

1. Open the options menu for the `ops/<user>/catalog-stats-view` materialization table row (i.e., click the icon consisting of three, vertical dots on the right-hand side of the row).

1. Select the _edit specification_ menu option.

1. Observe that the edit workflow for this materialization is initiated. The details form section accordion summary will display an error icon to the left of text reading _Materialization Details_. When expanded, a field-level error appears on the catalog name field of the details form section.

## Expected

The expected UX is still, _technically_, undetermined. For the sake of kicking off the brainstorming, a few approaches come to mind:

1. The _edit specification_ menu option could be disabled for tasks belonging to read-only tenants. When hovering over the option, a tooltip could appear identifying the reigning tenant, that they were granted read-only access to this tenant, and potentially containing a link to access grant/permission-related docs or encouraging them to visit the _User_ tab of the _Admin_ page to review their access grants.

[...]

PR #1528, opened 2025-03-24
Show `error' when user tries to `edit` entity they have no `write' access to

travjenkins:
## Issues

https://github.com/estuary/ui/issues/437

### 437

- Taking a very simplified approach and showing a message to user and removing the test/ save buttons. There are other actions the user can still take that will yell - but this is a pretty rare case anyway I do not think it matters that much.

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots


src/components/shared/Entity/Header/Actions.tsx
travjenkins, line 1: This and the progress do not 100% need to be broken out - but it did reduce renders a solid bit so went ahead and kept them broken out.

src/context/EntityContext.tsx
```diff
@@ -27,4 +28,27 @@ const useEntityType = () => {
...
+        switch (entityTypeValue) {
+            case 'capture':
+                return intl.formatMessage({ id: 'terms.capture' });
+                break;
+            case 'materialization':
+                return intl.formatMessage({ id: 'terms.materialization' });
+                break;
+            case 'collection':
+                return intl.formatMessage({ id: 'terms.transformation' });
+                break;
```
kiahna-tucker, line 46: Either the hook should be renamed to indicate the limited scope to which it applies or a comment should be left behind indicated its intended scope.

Issue #1542, opened 2025-03-31
Sometimes the "you don't have edit access" message can be shown at wrong times


travjenkins:
**Bug**
Dave was seeing this message when using edit on a materialization. This was not a support account - but his personal account. When we reloaded the message went away.

**Expected**
Message should not be displayed.

**Screenshots**
Include a screenshot of what you saw when the bug happened.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**

Issue #1426, opened 2025-01-20
Collection Naming Updates Prevent Searching Schemas


dyaffe:
**Problem**
After the updates to how we name collections, users can no longer search for schemas in the capture edit search box.

It would be ideal to change how we show collection names in that portion of the UI to show both the schema and table name so that users could see and search for schemas as well as table names.

travjenkins:
🤦 ... yeah... I did not think about this at all. 

Tech Notes:
We'll need to see how this can work with the MUI table we're using to show the list. We might have to pull filtering out of the table and into stand alone code.

travjenkins:
We should also handle the second request from https://github.com/estuary/ui/issues/1445 at this time

**When a user searches for a specific binding, it would be nice to show the first one in the filtered list automatically.**

PR #1552, opened 2025-04-09
Make the binding search UX better

travjenkins:
## Issues

Contributes to https://github.com/estuary/ui/issues/1426
Contributes to https://github.com/estuary/ui/issues/1542

### 1426

- Highlight the words that match
- Convert from `DataGrid` to a `Table` approach

### 1542
- make sure we have data before showing the warning

### To Do
- [x] Have group toggle react to the filter list
- [x] Get border around scroll bar working
- [x] Hide footer with flag
- [x] **Transformation create** - renders correctly
- [x] **Transformation create** - works correctly
- [x] Test adding collections
- [x] ~~Decide on when to scroll in the list?~~
- [x] ~~Add top/bottom controls?~~
- [x] Get remove all working
- [x] Keep input focused when filtering the first time
- [x] Update disable/enable all to be a single toggle
- [x] Make sure row selection does not fire on remove

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

### Capture
Initial view did not much has changed here. We no longer have pagination so the footer shows some general stats
image

Search now shows the full name (which is used in search) and highlights the match
image

### Materialization
Major difference here is we show the remove button
image

### Footer
If all entities are disabled or enabled we mention it
image
image

If only some are disabled we show the count
image

### Group Actions
These always apply to the list being shown. So either everything or whatever is being filtered to
image
image

Toggle all state runs against the filtered list and also responds to the list. So if the filtered list is all enabled the toggle will show as enabled.
image


src/components/collection/Selector/List/CollectionSelectorBody.tsx
```diff
@@ -0,0 +1,141 @@
...
+                            itemKey={(index, data) =>
+                                data[index][COLLECTION_SELECTOR_UUID_COL]
+                            }
+                            itemSize={DEFAULT_ROW_HEIGHT} // Row height
+                            overscanCount={10}
+                            width={width}
+                            onItemsRendered={() => {
+                                checkScrollbarVisibility();
+                            }}
+                        >
```
travjenkins, line 80: This is a major change to take us from `DataGrid` to using a normal `Table`.

On item rendering checking the scrollbar helps with performance a _tiny_ bit.

src/components/collection/Selector/List/CollectionSelectorBody.tsx
```diff
@@ -0,0 +1,141 @@
...
+                            {({ index, style, data }) => {
+                                const row = data[index];
+
+                                return (
+                                    <TableRow
+                                        key={row[COLLECTION_SELECTOR_UUID_COL]}
+                                        component={Box}
+                                        style={style}
```
travjenkins, line 88: Sending in the virtualization styling here

src/components/collection/Selector/List/CollectionSelectorBody.tsx
```diff
@@ -0,0 +1,141 @@
...
+                                                              }
+                                                    }
+                                                >
+                                                    {column.renderCell?.(
+                                                        {
+                                                            row,
+                                                        },
+                                                        filterValue
+                                                    )}
+                                                </TableCell>
```
travjenkins, line 127: Tried to keep this close to how datagrid works and use the same functions and stuff like that.

Using a `className` here so we don't have to have a lot of repeating styling.

src/components/collection/Selector/List/CollectionSelectorTable.tsx
travjenkins, line 1: Moved this out mainly cause the styling is very large and houses a lot of styling that is applied to children based on selectors. Since we are using virtualization it helps keeping as much styling in the parent as possible.

src/components/collection/Selector/List/Header/Name.tsx
travjenkins, line 1: We don't need this anymore so removing since it can mess up when a user deletes all bindings while there is a filter value and we try to clear out the input.

src/components/collection/Selector/index.tsx
```diff
@@ -21,38 +21,35 @@ function CollectionSelector({
         itemType ?? intl.formatMessage({ id: 'terms.collections' });
 
     return (
-        <Box>
-            <Box sx={{ height: '100%', width: '100%' }}>
```
travjenkins, line 25: Switched the list to render with `flex` and stop using hardcoded heights.

src/components/editor/Bindings/NoResults.tsx
travjenkins, line 1: We no longer have `DataGrid` to handle empty state

src/components/editor/Bindings/Row/Name.tsx
```diff
@@ -30,10 +29,21 @@ function BindingsSelectorName({
...
-                {shortenName ? stripPathing(collection) : collection}
+            <Typography component="span" {...typographyTruncation}>
+                <Highlighter
+                    autoEscape={false}
+                    highlightClassName={HIGHLIGHT_CLASS_NAME}
+                    highlightTag={NameHighlight}
+                    searchWords={[filterValue ?? '']}
+                    textToHighlight={collection.join('')}
+                    unhighlightClassName={UNHIGHLIGHT_CLASS_NAME}
+                />
```
travjenkins, line 46: We now will highlight what is matched during the filter.

src/components/editor/Bindings/Row/NameHighlight.tsx
travjenkins, line 1: Dave liked the bolding but not the "highlighter" styling that is applied by the browser when using the `mark` tag.

src/components/editor/Bindings/Selector.tsx
```diff
@@ -1,147 +1,40 @@
...
-import { useBindingStore } from 'src/stores/Binding/Store';
-import { useDetailsFormStore } from 'src/stores/DetailsForm/Store';
+import { useBinding_setCurrentBinding } from 'src/stores/Binding/hooks';
 import { useFormStateStore_isActive } from 'src/stores/FormState/hooks';
-import { hasLength } from 'src/utils/misc-utils';
 
 interface BindingSelectorProps {
     disableSelect?: boolean;
-    height?: number | string;
+    legacyTransformHeightHack?: number;
```
travjenkins, line 12: I really want to remove this property so marking it as something we want to remove.

src/components/shared/Entity/List/index.tsx
```diff
@@ -12,7 +12,7 @@ function EntityList({ content, PrimaryCTA, toggle }: EntityListProps) {
     const intl = useIntl();
 
     return (
-        <Box>
+        <Stack direction="column" sx={{ height: '100%' }}>
```
travjenkins, line 15: Mainly using `Stack` to keep the `flex` repeat styling limited.

src/components/tables/EntityTable/TableHeader.tsx
```diff
@@ -107,7 +103,9 @@ function EntityTableHeader({
                             }
                             sx={tableCellSX}
                         >
-                            {selectData && column.field && !hide ? (
+                            {column.renderInlineHeader && !hide ? (
+                                column.renderInlineHeader(index)
+                            ) : selectData && column.field && !hide ? (
```
travjenkins, line 108: Need to add function that can handle rendering within the default `TableCell` tags

src/hooks/useBindingSelectorCells.tsx
travjenkins, line 1: Brought in as much as I could from the old component into this hook.

src/hooks/useBindingSelectorCells.tsx
```diff
@@ -0,0 +1,161 @@
...
+                cellRenderer: (params, filterValue) => {
+                    const filteringActive = Boolean(filterValue);
+
+                    const collectionParts = filteringActive
+                        ? splitPathAndName(
+                              params.row[COLLECTION_SELECTOR_NAME_COL]
+                          )
+                        : isCapture
+                          ? [params.row[COLLECTION_SELECTOR_STRIPPED_PATH_NAME]]
+                          : [params.row[COLLECTION_SELECTOR_NAME_COL]];
```
travjenkins, line 100: On captures if the user is filtering we will show the full name

src/hooks/useBindingSelectorCells.tsx
```diff
@@ -0,0 +1,161 @@
...
+                                                COLLECTION_SELECTOR_UUID_COL
+                                            ]
+                                        }
+                                        collection={
+                                            params.row[
+                                                COLLECTION_SELECTOR_NAME_COL
+                                            ]
+                                        }
+                                    />
+                                ),
```
travjenkins, line 120: Allow passing in the start icon so that transformation create won't show an error indicator that has no details.

src/hooks/useBindingSelectorNotification.tsx
travjenkins, line 1: Taken from the original component.

src/hooks/useCanEditEntity.ts
travjenkins, line 1: This is a weird issue I ran into where due to HMR we might show the "you don't have access" warning which has been happening in prod (https://github.com/estuary/ui/issues/1542)

src/hooks/useBindingSelectorCells.tsx
travjenkins, line 1: Taken from the original component.

src/hooks/useReactWindowReadyToScroll.ts
travjenkins, line 1: Shared from logs

src/hooks/useReactWindowScrollbarGap.ts
travjenkins, line 1: Added to keep the columns lined up like `DataGrid` does. This does not solve it for all tables and just for virtualized once - hence the name including `ReactWindow`

src/components/collection/Selector/List/CollectionSelectorFooter.tsx
travjenkins, line 1: Showing total count and disabled count in the footer

kiahna-tucker (approved):

Issue #1605, opened 2025-05-07
Edit fails if a user cannot see the data plane it is on


travjenkins:
**Bug**
When a user does not have read access on the data plane that is selected for an edit - everything blows up.

**Expected**
Make the data-plane selector an optional field (which would allow it to be rendered without an input as well as the entire form) in the event a data-plane ID specified in the URL could not be found, use the data-plane ID specified in the `live_spec_ext` row for the task to create publications, and add a LogRocket event for this specific scenario to aid support.

**Screenshots**
N/A - they include private info. See the slack thread

**Desktop (please complete the following information):**
N/A

**Additional context**
https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1746634680688679

PR #1607, opened 2025-05-07
Ensure the absence of data-plane options in workflows is non-blocking

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1605

### 1605

* Include the `data_plane_name` in the `live-specs-ext` call used to prefill the details form store state in edit workflows. This is required because the `publications` table references data-planes by name and not ID. In the event no data-plane options are returned, `data_plane_name` and `data_plane_id` are used to create a fallback data-plane option (i.e., an element in the `dataPlaneOptions` array, a details form store state property) and prefill the selected data-plane (i.e., the details form store state property, `details.data.dataPlane`).

* Create a fallback option for the data-plane selector `oneOf` whose value is a dummy `DataPlaneOption` object and label is `No options found`. This primarily safeguards create workflows; in the event no data-plane options are returned, the data-plane selector can still be rendered but the task will be effectively forced into the default, public data-plane.

* Track instances where data-plane options are not returned with a LogRocket event.

**NOTE:** The client will continue to display an error in the place of the form if the call to fetch the series of data-plane options **fails**.

### Manually tested

* Validate that the form is initialized in the event no data-plane options were returned in all workflows.

**NOTE:** Testing this scenario locally is difficult given there is only one functional data-plane available and knowledge of how to create a functional, local data-plane is held by a very small subset of the engineering team.

### Automated tests

_N/A_

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

### DIFF
--- a/src/api/liveSpecsExt.ts
+++ b/src/api/liveSpecsExt.ts
@@ -214,6 +214,7 @@ export interface LiveSpecsExtQuery_DetailsForm {
     spec_type: Entity;
     spec: any;
     data_plane_id: string;
+    data_plane_name: string;
     connector_tag_id: string;
     connector_image_name: string;
     connector_image_tag: string;
@@ -226,6 +227,7 @@ const DETAILS_FORM_QUERY = `
     spec_type,
     spec,
     data_plane_id,
+    data_plane_name,
     connector_tag_id,
     connector_image_name,
     connector_image_tag,
--- a/src/api/liveSpecsExt.ts
+++ b/src/api/liveSpecsExt.ts
@@ -215,6 +215,7 @@ export interface LiveSpecsExtQuery_DetailsForm {
     spec: any;
     data_plane_id: string;
     data_plane_name: string | null;
+    reactor_address: string | null;
     connector_tag_id: string;
     connector_image_name: string;
     connector_image_tag: string;
@@ -228,6 +229,7 @@ const DETAILS_FORM_QUERY = `
     spec,
     data_plane_id,
     data_plane_name,
+    reactor_address,
     connector_tag_id,
     connector_image_name,
     connector_image_tag,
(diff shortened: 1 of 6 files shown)


src/api/liveSpecsExt.ts
```diff
@@ -214,6 +214,7 @@ export interface LiveSpecsExtQuery_DetailsForm {
     spec_type: Entity;
     spec: any;
     data_plane_id: string;
+    data_plane_name: string;
```
travjenkins, line 217: This type needs updates as this can be `null`

src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
```diff
@@ -50,6 +50,26 @@ export default function useDataPlaneField(
...
+                    dataPlaneName: {
+                        cluster: '',
+                        prefix: '',
+                        provider: '',
+                        region: '',
+                        whole: '',
+                    },
+                    id: '',
+                    reactorAddress: '',
+                    scope: 'public',
```
travjenkins, line 67: We should keep this default in the data plane utils.

src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
```diff
@@ -50,6 +53,15 @@ export default function useDataPlaneField(
...
             });
+        } else {
+            // The details form store hydrator does not fail loudly when no data-plane options are found
+            // and the create workflow does not have a fallback data-plane option to use in that scenario.
+            dataPlanesOneOf.push({
+                const: DATA_PLANE_OPTION_TEMPLATE,
+                title: intl.formatMessage({
+                    id: 'workflows.dataPlane.label.noOptionsFound',
+                }),
+            });
```
travjenkins, line 64: So adding this will mean the initial value is no right and will switch the autocomplete between `controlled` and `uncontrolled`. We'll need to make sure either the renderer handles this and knows when to default the value or we set the value ahead of time.

PR #1617, opened 2025-05-16
adding data planes to table

travjenkins:
## Issues

### issue_number

-   changes...

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

### DIFF
--- a/src/components/tables/StorageMappings/Rows.tsx
+++ b/src/components/tables/StorageMappings/Rows.tsx
@@ -2,6 +2,7 @@ import type { StorageMappings, StorageMappingStore } from 'src/types';
 
 import { TableCell, TableRow } from '@mui/material';
 
+import ChipListCell from 'src/components/tables/cells/ChipList';
 import ChipStatus from 'src/components/tables/cells/ChipStatus';
 import TimeStamp from 'src/components/tables/cells/TimeStamp';
 
@@ -28,6 +29,10 @@ export const tableColumns = [
     },
     {
         field: null,
+        headerIntlKey: 'entityTable.data.dataPlanes',
+    },
+    {
+        field: null,
         headerIntlKey: 'entityTable.data.provider',
     },
     {
@@ -45,9 +50,11 @@ export const tableColumns = [
 ];
 
 function DataCells({ store }: DataCellProps) {
-    const { provider, bucket, prefix } = store;
+    const { provider, bucket, prefix, data_planes } = store;
     return (
         <>
+            <ChipListCell values={data_planes} maxChips={3} stripPath={false} />
+
             <TableCell>{provider}</TableCell>
 
             <TableCell>{bucket}</TableCell>
--- a/src/lang/en-US/EntityTable.ts
+++ b/src/lang/en-US/EntityTable.ts
@@ -24,6 +24,7 @@ export const EntityTable: Record<string, string> = {
     'entityTable.data.provider': `Provider`,
     'entityTable.data.bucket': `Bucket`,
     'entityTable.data.prefix': `Prefix`,
+    'entityTable.data.dataPlanes': `Data Planes`,
     'entityTable.data.storagePrefix': `Prefix`,
     'entityTable.data.sharedPrefix': `Shared Prefix`,
     'entityTable.data.sharedWith': `Shared With`,
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -142,6 +142,7 @@ export interface StorageMappingStore {
     provider: string;
     bucket: string;
     prefix: string;
+    data_planes: string[];
 }
 export interface StorageMappings {
     id: string;


src/forms/renderers/DataPlaneSelector/AutoComplete.tsx
```diff
@@ -107,25 +110,29 @@ export const DataPlaneAutoComplete = ({
...
-                        {intl.formatMessage({ id: `common.${group}` })}
-                    </Typography>
-
-                    <MenuList style={{ padding: 0 }}>{children}</MenuList>
-                </li>
-            )}
+            renderGroup={({ group, children }) =>
+                children === null ||
+                (isArray(children) &&
+                    children.every((node) => node === null)) ? null : (
```
kiahna-tucker, line 116: This was done to hide group headers when no child nodes can be found. Since the rendered options are filtered, `children` can be null or a null array.

src/components/shared/Entity/DetailsForm/useDataPlaneOptions.ts
kiahna-tucker, line 1: This hook is only used to scope the data-plane options inside `DataPlaneAutoComplete` (i.e., _src/forms/renderers/DataPlaneSelector/AutoComplete.tsx_).

src/hooks/dataPlanes/useEvaluateDataPlaneOptions.ts
```diff
@@ -0,0 +1,80 @@
...
+
+                setDataPlaneOptions(fallbackOptions);
+
+                return fallbackOptions;
+            }
+
+            const options = dataPlanes
+                ? dataPlanes.map((dataPlane) =>
+                      generateDataPlaneOption(dataPlane, dataPlaneNames[0])
+                  )
```
kiahna-tucker, line 70: See related [comment](https://github.com/estuary/ui/pull/1617/files#r2164416290).

src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
```diff
@@ -75,27 +80,36 @@ export default function useDataPlaneField(
...
+    const dataPlaneUISchema = useMemo(
+        () => ({
+            label: intl.formatMessage({
+                id: 'workflows.dataPlane.label',
+            }),
+            scope: `#/properties/${DATA_PLANE_SCOPE}`,
+            type: 'Control',
+            options: {
+                readOnly: objectRoles.length > 1 && entityName.length === 0,
+            },
```
travjenkins, line 92: This is okay for now - but if this ends up living awhile we'd want to make use of the [rules api](https://jsonforms.io/docs/uischema/rules/)

src/components/shared/Entity/DetailsForm/useEvaluateStorageMapping.ts
```diff
@@ -0,0 +1,108 @@
...
+                    ? options.find(
+                          (option) =>
+                              option.dataPlaneName.whole === dataPlaneNames[0]
+                      )
+                    : undefined;
+            }
+
+            return undefined;
+        },
+        [hasSupportRole, options]
```
travjenkins, line 55: We should just pass in `hasSupportRole` to the function since it is "private" and the caller already listens to changes on that property.

src/forms/renderers/DataPlaneSelector/AutoComplete.tsx
travjenkins: This seems backwards though... cause when it isn't true it is going to be checking the entire array every time. I think we should just check if `some` are `not null` instead

src/utils/misc-utils.ts
```diff
@@ -61,7 +61,7 @@ export const splitPathAndName = (stringVal: string): string[] => {
     if (!stringVal) return [];
 
     const lastSlash = stringVal.lastIndexOf('/') + 1;
-    return [stringVal.substr(0, lastSlash), stringVal.substr(lastSlash)];
+    return [stringVal.substring(0, lastSlash), stringVal.substring(lastSlash)];
```
travjenkins, line 64: This is the right call - but making sure you're aware these are different calls now. I don't think they have any consequence for us though.

src/stores/Entities/hooks.ts
```diff
@@ -43,8 +45,27 @@ export const useEntitiesStore_atLeastOneAdminTenant = () => {
...
+            if (!restrictByStorageMappings || hasSupportRole) {
+                return state.capabilities.admin;
+            }
+
+            return Object.keys(state.storageMappings).filter(
+                (storageMappingPrefix) =>
+                    state.capabilities.admin.some((adminPrefix) =>
+                        storageMappingPrefix.startsWith(adminPrefix)
+                    )
+            );
```
travjenkins, line 66: On the next update to storage mappings we should make it so this is not checked all the time.

Issue #1627, opened 2025-05-22
During Capture Create you cannot disable a binding


travjenkins:
**Bug**
Create capture
run a discover
use the binding disable toggle
nothing changes

**Expected**
the field should get disabled

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Additional context**
https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1747919402314139

PR #1628, opened 2025-05-22
Do not query for everything in a draft | Update `disable` directly on `draft`

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1627

### 1627

- Do not query for all items in a `draft spec` and just get the entity we're creating/editing
    - This prevents issues when there are over 1k bindings and the draft would show as "missing" in the workflow
- Make `disable` on the bindings just directly update the `draft` when it can so that we do not need to require users to generate a spec again after just altering disable/enable
    - We still show the generate button if the user enables a previously disabled collection

### Misc

- Added clean up in for `debounce` stuff that could be running after unmount

### Bugs

- [ ] The list of bindings is getting duplicated
     - This does not seem related to these changes and super rare
     - image

- [x] knowing when a specific binding has an invalid setting is not working
- [x] Removing a binding and messing with disable with/without draft in Materialization
- [x] The row specific disable toggle is not rendering properly without draft in Materialization

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots


src/hooks/useDraftSpecs.ts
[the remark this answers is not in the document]
kiahna-tucker: This does **not** seem like an issue to me. Curious if this was done before `useInitializeCollectionDraft` was created, perhaps it was used to extract collection information for the _Collection_ tab of the binding selector.

src/components/editor/Bindings/Row/Toggle.tsx
```diff
@@ -5,39 +5,33 @@ import { Button, Switch } from '@mui/material';
...
 
-    const toggleDisable = useBinding_toggleDisable();
-    const disabled = useBinding_resourceConfigOfMetaBindingProperty(
-        bindingUUID,
-        'disable'
-    );
+    const { currentSetting, updateDraft } = useDisableUpdater(bindingUUID);
+
+    console.log('bindingUUID', bindingUUID);
+    console.log('currentSetting', currentSetting);
```
kiahna-tucker, line 19: Calling out the presence of this logging statement for general awareness.

src/hooks/bindings/useBindingDisabled.ts
kiahna-tucker, line 1: I would prefer for `index` to always be a number. That way the memo below would be equipped to handle the edge case where `bindingIndex` is `-1`. If you would prefer not to adjust the selector logic, I would update the memo to ensure the index is nonnegative to be safe.

src/hooks/useServerUpdateRequiredMonitor.ts
```diff
@@ -56,14 +56,7 @@ const useServerUpdateRequiredMonitor = (draftSpecs: DraftSpecQuery[]) => {
...
 
-                                // Do a quick simple disabled check before comparing the entire object
-                                if (
-                                    resourceConfigs[bindingUUID].meta
-                                        .disable !==
-                                    getDisableProps(binding?.disable).disable
-                                ) {
-                                    return true;
-                                }
+                                // Do no longer check disable here because disable directly updates the draft
```
kiahna-tucker, line 59: Small grammatical error in this comment: _No longer check ..._ or _Do not check ..._ are suitable corrections.

src/lang/en-US/Workflows.ts
```diff
@@ -52,6 +52,7 @@ export const Workflows: Record<string, string> = {
     'workflows.collectionSelector.notifications.remove': `{count} {itemType} removed`,
     'workflows.collectionSelector.notifications.toggle.enable': `{count} {itemType} enabled`,
     'workflows.collectionSelector.notifications.toggle.disable': `{count} {itemType} disabled`,
+    'workflows.collectionSelector.notifications.toggle.disable.error': `Changes Reverted. Unable to update server. Please try again.`,
```
kiahna-tucker, line 55: Why is _reverted_ capitalized?

src/stores/Binding/hooks.ts
kiahna-tucker, line 1: We love narrowed type annotations! Thank you. To make the `useBinding_resourceConfigOfMetaBindingProperty` hook even _more_ useful, could it be typed akin to `useBinding_collectionMetadataProperty` at this time? This way the return type will reflect that of the desired property.

src/stores/Binding/slices/ToggleDisable.ts
```diff
@@ -0,0 +1,113 @@
...
+    set: NamedSet<StoreWithToggleDisable>
+): StoreWithToggleDisable => ({
+    toggleDisable: (updates) => {
+        set(
+            produce((state: BindingState) => {
+                updates.forEach(({ uuid, val }) => {
+                    const collectionName =
+                        state.resourceConfigs[uuid].meta.collectionName;
+
+                    if (val === true) {
```
kiahna-tucker, line 33: If `val` is a required boolean property of `ToggleDisableUpdate`, why do we need to explicitly check that it equates to `true`? I have seen this condition in a couple places in this PR.

src/stores/Binding/slices/ToggleDisable.ts
```diff
@@ -0,0 +1,113 @@
+import type { BindingState } from 'src/stores/Binding/types';
+import type { NamedSet } from 'zustand/middleware';
+
+import produce from 'immer';
+import { isBoolean } from 'lodash';
+
+import { hasLength } from 'src/utils/misc-utils';
+
+export type ToggleDisableUpdate = {
+    uuid: string;
```
kiahna-tucker, line 10: The generic nature of `uuid` reminds me of `id`. Should this property be renamed to `bindingUUID` to clarify the state to which the UUID corresponds?

src/stores/Binding/types.ts
```diff
@@ -57,10 +58,17 @@ export interface BindingChanges {
     addedCollections: string[];
 }
 
+export type BindingDisableUpdate = {
+    uuid: string;
+    bindingIndex: number;
+    val: boolean;
+};
```
kiahna-tucker, line 65: How does this type differ from `ToggleDisableUpdate`?

kiahna-tucker (commented):
This PR has been moderately tested without issue. The code review inspired some questions that should be addressed before approval can be given.

src/stores/Binding/types.ts
travjenkins: It doesn't. I initially had two cause I wasn't sure if the shape would stay the same.

Issue #1629, opened 2025-05-27
Make sure users are creting a NEW task when using create


travjenkins:
Background:
A user can accidentally clobber an existing task by re-using a name.

Request:
We should add an `expectPubId: 00` to the drafts.

PR #1632, opened 2025-05-28
Default expected publication ID of task in create workflows

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1629

### 1629

* Set the expected publication ID in create workflows to prevent task collisions.

### Manually tested

* Validate that the expected publication ID of a task is set to `00:00:00:00:00:00:00:00` in create workflows.

* Validate that the edit workflow behavior is unchanged.

### Automated tests

_N/A_

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

_N/A_


src/hooks/useStoreDiscoveredCaptures.ts
```diff
@@ -69,7 +72,10 @@ function useStoreDiscoveredCaptures() {
                 const supabaseConfig: SupabaseConfig | null =
                     editWorkflow && lastPubId
                         ? { catalogName: entityName, lastPubId }
-                        : null;
+                        : {
+                              catalogName: '',
+                              lastPubId: NEW_TASK_PUBLICATION_ID,
+                          };
```
travjenkins, line 78: Eventually the `catalog name` is optional. What is the value of adding a blank one here?

src/hooks/useStoreDiscoveredCaptures.ts
```diff
@@ -69,7 +72,10 @@ function useStoreDiscoveredCaptures() {
                 const supabaseConfig: SupabaseConfig | null =
```
travjenkins, line 72: This `null` type can be removed now.

src/hooks/useStoreDiscoveredCaptures.ts
kiahna-tucker: `catalogName` is not an optional property of a `SupabaseConfig`.

travjenkins: Yup.

travjenkins (approved):
lgtm

Issue #1626, opened 2025-05-22
`resource config` fields with type `duration` are rendering incorrectly


travjenkins:
**Bug**
Discover from netsuite
Look at binding with `interval`

**Expected**
The input should show either 1 day or 24 hours

**Screenshots**

![Image](https://github.com/user-attachments/assets/d0162d06-eff1-4504-8d42-6146255c8997)

![Image](https://github.com/user-attachments/assets/cb0618df-957b-4531-b160-cd8ad2cc5563)

**Desktop (please complete the following information):**
n/a (I think / hope)

**Additional context**
https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1747918785642229

travjenkins:
Related to how we set `currentBinding` when searching.

PR #1639, opened 2025-05-30
Prevent old `resource config` from rendering when filtering bindings

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1626

### 1626

- Share the `hacky timeout` into a hook
- Always use the hook when setting current binding

#### Bugs

- [x] When filtering list if user clicks on another binding it will be selected and then immediately be switched back to the one auto selected during filtering.

### Manually tested

- create/edit of Capture and Materialization while switching between bindings

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Hard to show with screenshots. But now when you filter the bindings there will be a quick flash like when you click on a binding manually.


src/components/collection/Selector/List/index.tsx
```diff
@@ -154,15 +157,25 @@ function CollectionSelectorList({
...
+            // Selection disabled
+            !selectionEnabled ||
+            !setCurrentBinding ||
+            // Filter has not changed so we can skip. Otherwise while the bindings are filtered
+            //  if a user clicks manually on a binding then it will be selected and a split second
+            //  later this effect will run and set it back to a default value.
+            (filterValue.length > 0 &&
+                previousFilterValue &&
+                previousFilterValue.length > 0 &&
+                filterValue === previousFilterValue)
```
travjenkins, line 170: This fixes the first bug found during testing. Comment explains what it is doing.

src/stores/Binding/hooks.ts
travjenkins, line 1: Needing this is still annoying. Moving bindings selector to a base table and then opening the settings in a modal/slide out/etc. could solve needing this all together... and that would be nice.

Issue #1660, opened 2025-06-19
Some materialization edits fail due to `00` pub id being set


travjenkins:
**Bug**
There is _some_ steps a user is doing that causes edit to set the `NEW_TASK_PUBLICATION_ID` on the draft.

**Expected**
The `00` id is not set as this is edit and for sure a new

**Screenshots**
Include a screenshot of what you saw when the bug happened.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**

PR #1661, opened 2025-06-19
Better handling of draft updating

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1660

### 1660

- Check for `materialization` edit for `evolution` as that can be run there
- Keep binding `disable` setting in store and skip `spec` update if the binding is new

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

N/A

### DIFF
--- a/src/hooks/useStoreDiscoveredCaptures.ts
+++ b/src/hooks/useStoreDiscoveredCaptures.ts
@@ -26,7 +26,8 @@ function useStoreDiscoveredCaptures() {
     const [lastPubId] = useGlobalSearchParams([GlobalSearchParams.LAST_PUB_ID]);
 
     const workflow = useEntityWorkflow();
-    const editWorkflow = workflow === 'capture_edit';
+    const editWorkflow =
+        workflow === 'capture_edit' || workflow === 'materialization_edit';
 
     // Binding Store
     const evaluateDiscoveredCollections =


src/hooks/bindings/useDisableUpdater.ts
```diff
@@ -44,7 +47,24 @@ function useDisableUpdater(bindingUUID?: string) {
...
+                            missingIndex ||
+                            // Make sure there is a binding there at least
+                            !spec.bindings[bindingIndex] ||
+                            // Try to be safe and make sure we have the right name
+                            getCollectionName(spec.bindings[bindingIndex]) ===
+                                collectionName
+                        ) {
+                            missingIndex = true;
+                            return;
+                        }
```
travjenkins, line 66: Need to make sure what we are trying to disable is there. This is mainly just for `materialization` `edit` flow where a user enters the flow, adds some bindings, and then toggles the disable/enable without generating a catalog spec.

src/hooks/useDraftUpdater.ts
```diff
@@ -41,17 +41,28 @@ function useDraftUpdater() {
...
+                // Means we are trying to update something on the spec that isn't there yet so just store off
+                //  into the store and keep carrying on
+                logRocketEvent('PropertyUpdater', {
+                    updatedSpecMissing: true,
+                });
+
+                // If we do not have these yet do not fail - just keep going and hopefully the calling code
+                //  will save the value into a store
+                return Promise.resolve();
+            }
```
travjenkins, line 56: New functionality that should have always been there. We need a way for the specific calls to tell the update function to skip the update without error. This is like how `useTimeTravel` checks if the binding is in a spec or we just need to store things in a store.

PR #1662, opened 2025-06-19
PROD - disable toggle failing to show

travjenkins:
Need to match that the name is NOT equal :facepalm: 

https://github.com/estuary/ui/issues/1663


psFried (approved):
LGTM

Issue #1675, opened 2025-06-25
`disable` toggle not rendering right during `capture` create


travjenkins:
**Bug**
User discovers a capture
Click the disable all toggle
See none of the individual toggles change

**Expected**
The individual toggles should change

**Screenshots**
![Image](https://github.com/user-attachments/assets/a35d5dbb-e9fb-4b12-8b81-f8ca5a304e97)

**Desktop (please complete the following information):**
N/A (probably)

**Additional context**
N/A

PR #1676, opened 2025-06-25
Making the check work for toggle all call

travjenkins:
Putting current setting into memo

## Issues

https://github.com/estuary/ui/issues/1675

### 1675

- Need to ignore collection name checking when there is no name to compare to
- Only use the draft settings when there is a draft id. This is because if we have a draft id we are NOT going to show the `next` button and have the spec be generated again. So whatever is in the draft will be used in publication.

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

image

### DIFF
--- a/src/hooks/bindings/useDisableUpdater.ts
+++ b/src/hooks/bindings/useDisableUpdater.ts
@@ -5,7 +5,10 @@ import { useCallback, useMemo } from 'react';
 import { useSnackbar } from 'notistack';
 import { useIntl } from 'react-intl';
 
-import { useEditorStore_queryResponse_draftSpecs } from 'src/components/editor/Store/hooks';
+import {
+    useEditorStore_id,
+    useEditorStore_queryResponse_draftSpecs,
+} from 'src/components/editor/Store/hooks';
 import { useEntityType } from 'src/context/EntityContext';
 import useDraftUpdater from 'src/hooks/useDraftUpdater';
 import { useBindingStore } from 'src/stores/Binding/Store';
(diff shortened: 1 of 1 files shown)


src/hooks/bindings/useDisableUpdater.ts
```diff
@@ -57,9 +57,12 @@ function useDisableUpdater(bindingUUID?: string) {
...
                             !spec.bindings[bindingIndex] ||
-                            // Try to be safe and make sure we have the right name
-                            getCollectionName(spec.bindings[bindingIndex]) !==
-                                collectionName
+                            // Try to be safe and make sure we have the right name when we are updating
+                            //  a specific binding
+                            (collectionName &&
+                                getCollectionName(
+                                    spec.bindings[bindingIndex]
+                                ) !== collectionName)
```
travjenkins, line 65: Only need to check if the collection name is the same when the name is there to check :facepalm:

src/hooks/bindings/useDisableUpdater.ts
```diff
@@ -142,24 +145,26 @@ function useDisableUpdater(bindingUUID?: string) {
...
-    let currentSetting: boolean | null | undefined = null;
-
-    if (bindingIndex !== null) {
-        if (
-            draftSpecs?.[0]?.spec?.bindings &&
-            draftSpecs?.[0]?.spec?.bindings.length > 0 &&
-            draftSpecs[0].spec.bindings[bindingIndex]
-        ) {
-            currentSetting = draftSpecs[0].spec.bindings[bindingIndex].disable;
+    const currentSetting: boolean | null | undefined = useMemo(() => {
```
travjenkins, line 148: No real change here but putting it into a memo.

kiahna-tucker (approved):

Issue #1692, opened 2025-07-02
When trying to edit an entity the user is sometimes taken to create


travjenkins:
**Bug**
Not really sure what exactly causes this but when a user is trying to edit something they can be taken to create erroneously. 

**Expected**
They should only land on edit if they are editing an existing entity.

**Screenshots**
Include a screenshot of what you saw when the bug happened.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**

Issue #1690, opened 2025-07-01
Edit on data planes can show error


travjenkins:
**Bug**
Edit an entity
See data plane hows error about `oneOf`

**Expected**
data plane renders without error

**Screenshots**

![Image](https://github.com/user-attachments/assets/68ec76d0-f4d9-446d-843b-93e046d83cc4)

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**

PR #1693, opened 2025-07-02
Prevent going from edit to create and ensuring selected data plane is available during edit

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1690
https://github.com/estuary/ui/issues/1692

### 1690

- During edit checking if the stored value is on the list

### 1692

- replaced the `navigateToCreate` with a `setQuery`

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

1692 - URL updating when data plane is changed. This uses the `id` that is from the option so sometimes this will be a catalog name and sometimes a UUID for now.
image
image

### DIFF
--- a/src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
+++ b/src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
@@ -64,7 +64,9 @@ export default function useDataPlaneField(
             });
 
             // TODO (data-planes) - this should not really be needed but a solution to triage the prod issues
-            //  some users are seeing. Need an actual solution.
+            //  some users are seeing. Need an actual solution. We potentially do not even really want to make a proper
+            //  list of options during edit and rather just use the value from the live_spec_ext view as the ONLY option
+            //  since during edit the user should not be changing the data plane if one is already provided.
             // If we're in edit AND we have a stored data plane then it more than likely came from
             //  the live_specs_ext view and we should make sure that it is an option
             if (


src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
```diff
@@ -25,18 +26,21 @@ interface OneOfElement {
 export default function useDataPlaneField(
     entityType: EntityWithCreateWorkflow
 ) {
-    const dataPlaneIdInURL = useGlobalSearchParams(
-        GlobalSearchParams.DATA_PLANE_ID
-    );
+    const [query, setQuery] = useQueryParams({
+        [GlobalSearchParams.DATA_PLANE_ID]: StringParam,
+    });
+    const dataPlaneIdInURL = query[GlobalSearchParams.DATA_PLANE_ID];
```
travjenkins, line 32: We can just directly use the query param and read it from there so that we get easy access to the `set` function.

src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
```diff
@@ -58,6 +62,29 @@ export default function useDataPlaneField(
...
+            //  the live_specs_ext view and we should make sure that it is an option
+            if (
+                isEdit &&
+                storedDataPlane &&
+                dataPlanesOneOf.findIndex(
+                    (datum) =>
+                        datum.const.dataPlaneName.whole ===
+                        storedDataPlane.dataPlaneName.whole
+                ) === -1
+            ) {
```
travjenkins, line 78: Trying to keep this triage scoped heavily. If we're editing and have the stored data plane then make sure it is in the list. While testing this I never saw `storedDataPlane` being undefined while loading in edit.

src/components/shared/Entity/DetailsForm/useDataPlaneField.ts
```diff
@@ -105,21 +132,18 @@ export default function useDataPlaneField(
...
                     setEntityNameChanged(details.data.entityName);
 
-                    // TODO (data-plane): Set search param of interest instead of using navigate function.
-                    navigateToCreate(entityType, {
-                        id: details.data.connectorImage.connectorId,
-                        advanceToForm: true,
-                        dataPlaneId: selectedDataPlaneOption?.id ?? null,
+                    setQuery({
+                        [GlobalSearchParams.DATA_PLANE_ID]:
+                            selectedDataPlaneOption?.id ?? null,
```
travjenkins, line 137: Set the query just as we did before but no longer using the navigate function.

kiahna-tucker (approved):

Issue #1677, opened 2025-06-25
Sometimes `logs` are shown from the wrong entity


travjenkins:
**Bug**
Jonathan was able to view a previous entities logs while trying to see the `stats-view` logs.

**Expected**
Only the logs for the current entity should be shown

**Screenshots**
See the Slack thread

**Desktop (please complete the following information):**
N/A

**Additional context**
Slack Link: https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1750862835336539

PR #1711, opened 2025-07-10
Do not update journal data if unmounted

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1677

### 1677

-  I was not able to reproduce the issue but I could get React to complain about a memory leak. So I think these are tightly coupled.

### Misc

- Cleaning up some old code for loading LogRocket manually that we have not been wanting for awhile.

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Slowed the network and then clicked on `logs` and quickly clicked on `spec`
image


src/hooks/journals/useJournalData.ts
```diff
@@ -175,19 +177,33 @@ const useJournalData = (
...
-                    setLoading(false);
+                    if (fetchCancelHack.current) {
+                        setLoading(false);
+                    }
+                }
+
+                if (!fetchCancelHack.current) {
+                    logRocketEvent('JournalData', {
+                        skippingFetch: true,
+                    });
```
travjenkins, line 206: Putting this in just in case we run into issues

Issue #1798, opened 2025-11-03
Multiple catalog names can match on alert history page


travjenkins:
Since we are using the `Prefix` query if there are two tasks with names that are the same and the two connectors only differ on the ending `suffix` they will both show up on the alerts history page.

Overview page
<img width="153" height="98" alt="Image" src="https://github.com/user-attachments/assets/4edad593-c3d4-48d0-8936-171b151c9e2b" />

Details page of the entity with a shorter name
<img width="124" height="340" alt="Image" src="https://github.com/user-attachments/assets/ebdd3d3e-4600-405f-a44a-1fe6ea48ca6a" />

PR #1799, opened 2025-11-03
Update `Alerts` on `Details` to use `LiveSpecs`

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1798

### 1798

- Switch over to the query that matches on the spec name
- Update the alerts history table to work for the new query
- 

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Half active and half resolved for one entity
All resolved for another
image

Overview still works
image

Alert History page works
image

Pagination works
image

Limited (faked being 3... normal setting is 5)
image

Only Resolved
image

### DIFF
--- a/src/components/tables/AlertHistory/index.tsx
+++ b/src/components/tables/AlertHistory/index.tsx
@@ -1,9 +1,9 @@
 import type { AlertHistoryTableProps } from 'src/components/tables/AlertHistory/types';
 import type { TableState } from 'src/types';
 import type {
-    AlertEdge,
+    AlertHistoryQueryResponse,
     AlertNode,
-    LiveSpecsQueryResponse,
+    AlertNodeEdge,
     LiveSpecVariables,
     WithPagination,
 } from 'src/types/gql';
@@ -43,7 +43,7 @@ const PAGE_SIZE = 5;
 const MAX_ACTIVE_ALERTS = 5;
 
 // Query for active alerts
-const activeAlertsQuery = gql<LiveSpecsQueryResponse, LiveSpecVariables>`
+const activeAlertsQuery = gql<AlertHistoryQueryResponse, LiveSpecVariables>`
     query ActiveAlertsQuery($catalogName: String!) {
         liveSpecs(by: { names: $catalogName }) {
             edges {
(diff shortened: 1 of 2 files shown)


src/components/tables/AlertHistory/index.tsx
travjenkins: Since this is big - overall this "just" splits the two queries apart into different queries. That then causes a lot of smaller tweaks and and changes.

src/components/tables/AlertHistory/index.tsx
```diff
@@ -91,39 +111,60 @@ function AlertHistoryTable({
...
-            before: beforeCursor,
-
-            [!afterCursor ? 'last' : 'first']: PAGE_SIZE,
+            catalogName: catalogName,
+            ...(active
+                ? {}
+                : {
+                      before: beforeCursor,
+                      last: PAGE_SIZE,
+                  }),
```
travjenkins: `active` query does not support pagination so not sending.

src/components/tables/AlertHistory/index.tsx
```diff
@@ -91,39 +111,60 @@ function AlertHistoryTable({
     const [beforeCursor, setBeforeCursor] = useState<string | undefined>(
         undefined
     );
-    const [afterCursor, setAfterCursor] = useState<string | undefined>(
-        undefined
+    const [cursorHistory, setCursorHistory] = useState<(string | undefined)[]>(
+        []
```
travjenkins: The `AlertHistory` pagination does not have `after` and `first` so keeping track of the cursors so we can pagination both ways (some what). There are some limitations if the resolved alerts change after we have these cached.

src/components/tables/AlertHistory/index.tsx
```diff
@@ -149,6 +190,13 @@ function AlertHistoryTable({
         status: TableStatuses.LOADING,
     });
 
+    // Extract the data based on active prop
+    const specNode = data?.liveSpecs?.edges?.[0]?.node;
+    const alerts = active
+        ? specNode?.activeAlerts
+        : specNode?.alertHistory?.edges;
+    const pageInfo = active ? null : specNode?.alertHistory?.pageInfo;
```
travjenkins: Now that there are two queries we have to manage to different shapes for now.

Issue #1822, opened 2025-11-26
One fetch from a data plane failing causes all status to not load


travjenkins:
**Bug**
On the entity table we wait for all shards to load correctly

**Expected**
We can load the ones we have access to and ignore the others

**Screenshots**
Include a screenshot of what you saw when the bug happened.

**Desktop (please complete the following information):**
 - OS: [e.g. Windows]

**Additional context**
https://estuaryworkspace.slack.com/archives/C03Q9A3GFEE/p1764174221448049

PR #1823, opened 2025-11-26
We don't need ALL shard lists to come back - display what we have

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1822

### 1822

- Let what shards we get back to come back
- Do not spam calls forever if we have issues
- Check if everything failed and return an error
- Only wait `3 seconds` when fetching shards

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Show a special error for details page
image

One failure does not block the others
image
image

### DIFF
--- a/src/components/editor/EditorWithFileSelector.tsx
+++ b/src/components/editor/EditorWithFileSelector.tsx
@@ -1,4 +1,4 @@
-import type { MonacoEditorProps } from 'src/components/editor/MonacoEditor';
+import type { MonacoEditorProps } from 'src/components/editor/MonacoEditor/types';
 
 import EditorFileSelector from 'src/components/editor/FileSelector';
 import ListAndDetails from 'src/components/editor/ListAndDetails';
--- a/src/components/editor/MonacoEditor/index.tsx
+++ b/src/components/editor/MonacoEditor/index.tsx
@@ -1,5 +1,5 @@
 import type * as monacoEditor from 'monaco-editor/esm/vs/editor/editor.api';
-import type { Entity } from 'src/types';
+import type { MonacoEditorProps } from 'src/components/editor/MonacoEditor/types';
 
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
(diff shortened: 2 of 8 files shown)


src/hooks/shards/useShardsList.ts
```diff
@@ -40,18 +66,53 @@ const useShardsList = (catalogNames: string[]) => {
...
+
+            if (
+                isPromiseFulfilledResult(shardResponse) &&
+                shardResponse.value.shards.length > 0
+            ) {
+                allCallsFailed = false;
+                failureCountsRef.current[catalogName] &&= 0;
+                response.shards = response.shards.concat(
+                    shardResponse.value.shards
+                );
```
travjenkins: If we fulfilled something we can mark that we got at least one success and then reset any counters if they exist

src/hooks/shards/useShardsList.ts
```diff
@@ -40,18 +66,53 @@ const useShardsList = (catalogNames: string[]) => {
...
-        const response: { shards: Shard[] } = { shards: [] };
+                const failureCount = failureCountsRef.current[catalogName];
 
-        shardResponses.forEach(({ shards }) => {
-            response.shards = response.shards.concat(shards);
+                logRocketEvent('ShardsList', {
+                    catalogName,
+                    failureCount: failureCountsRef.current[catalogName],
+                    willRetry: failureCount < MAX_FAILURES,
+                });
```
travjenkins: If there was no fulfillment then keep a count of that and then fire an event so we can keep an eye on this.

src/hooks/shards/useShardHydration.ts
```diff
@@ -12,10 +14,16 @@ import {
     useShardDetail_setShards,
 } from 'src/stores/ShardDetail/hooks';
 
+interface UseShardHydrationSettings {
+    defaultMessageId?: ShardStatusMessageIds;
+}
```
travjenkins: Initially I needed to add a setting but no longer need to but leaving this in for future proof.

src/utils/dataPlane-utils.ts
travjenkins: Main change is the adding of a "timeout" since the DPG does not have one and adding one there is kinda tough. Also added some eventing so we can keep an eye on this. As called out in the comments there is nothing special about `3000` so may need to update that.

kiahna-tucker (commented):
Releasing review comments while manual testing is underway.

Issue #1826, opened 2025-12-02
Help users share the right access


travjenkins:
### Background
Our UX for sharing access is confusing to users. Basically everywhere else in the dashboard if they see an input with a `prefix` name they must also provide the rest of the name. 

The input that needs changed
<img width="946" height="193" alt="Image" src="https://github.com/user-attachments/assets/2d83c755-cf2f-405f-98c5-77a11449d450" />

We added a general warning but this does not seem to be enough
<img width="954" height="302" alt="Image" src="https://github.com/user-attachments/assets/a4890b1a-a683-44a7-8f54-a6dbd90c98c3" />

### Request
We should update the UX so that the user must take a specific action that allows them to limit access.

The general thought is the input should just be prefix until the user interacts with _something_ before entering more.

Content ideas: Show a checkbox / button that says `limit access further`

PR #1833, opened 2025-12-05
Greg/enhancement/new user access

GregorShear:
## Issues

https://github.com/estuary/ui/issues/1826

## Changes

- make it more obvious that the user is granting full or limited access to the selected prefix
- replace "type" selector with clearer "reusable invite" checkbox
- disable tenant selector if only one tenant available

### Manually tested
- very long prefix
- dark / light mode
- invalid characters in name

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots
old:
521435384-a4890b1a-a683-44a7-8f54-a6dbd90c98c3

new:
Screenshot 2025-12-08 at 4 43 43 PM
Screenshot 2025-12-08 at 4 44 52 PM

### DIFF
--- a/src/components/tables/AccessGrants/AccessLinks/Dialog/GenerateInvitation.tsx
+++ b/src/components/tables/AccessGrants/AccessLinks/Dialog/GenerateInvitation.tsx
@@ -282,7 +282,6 @@ function GenerateInvitation({
                                         ref={subPrefixInputRef}
                                         value={name}
                                         placeholder="example"
-                                        // disabled={!limitedAccessScope}
                                         onChange={(event) => {
                                             prefixHandlers.setName(
                                                 event.target.value


src/components/tables/AccessGrants/AccessLinks/Dialog/GenerateInvitation.tsx
```diff
@@ -91,82 +163,218 @@ function GenerateInvitation({
...
+                                        id: 'admin.users.prefixInvitation.label.scope.full',
+                                    })}
+                                    <span
+                                        style={{
+                                            fontFamily: 'monospace',
+                                            fontSize: 12,
+                                        }}
+                                    >
+                                        {clampedPrefix}
+                                    </span>
```
travjenkins: We have `TechnicalEmphasis` which _might_ be close to what you are wanting here. So I'd see if that can do what you're hoping.

src/components/tables/AccessGrants/AccessLinks/Dialog/GenerateInvitation.tsx
```diff
@@ -63,6 +134,7 @@ function GenerateInvitation({
         generateInvitation: (event: React.MouseEvent<HTMLElement>) => {
             event.preventDefault();
 
+            const objectRole = prefix + name;
```
travjenkins: We prefer string templates... which I am actually surprised isn't an eslint rule already.

This should not always add the `name`
Image

src/components/tables/AccessGrants/AccessLinks/Dialog/GenerateInvitation.tsx
```diff
@@ -91,82 +163,218 @@ function GenerateInvitation({
...
+                        changeHandler={handlers.setGrantCapability}
+                    />
+                </Box>
+
+                <Divider flexItem orientation="vertical" />
 
-            <Grid item xs={4} md={2}>
-                <AutocompletedField
+                <Box sx={{ flex: 1 }}>
+                    <RadioGroup
```
travjenkins: This is an interesting thing we'll wanna chat about in dailies on Monday.

Up until now normally selecting from a group of options like this go into a raio drop down. We could use some of these styles and approaches... however an actual menu here would be a big step back I think.
Image

src/components/tables/AccessGrants/AccessLinks/Dialog/GenerateInvitation.tsx
```diff
@@ -21,12 +32,59 @@ import { appendWithForwardSlash, hasLength } from 'src/utils/misc-utils';
...
+}) => {
+    return (
+        <FormControlLabel
+            value={value}
+            control={
+                <Radio
+                    size="small"
+                    sx={{
+                        '&.MuiRadio-root:hover': {
+                            bgcolor: 'transparent !important',
```
travjenkins: I would just leave this. It does look a bit wonky but we leave these default states in most spots. Also, this will still show with keyboard controls and we _generally_ try to keep keyboard and mouse interactions pretty close.

Example is the table below this form:
Image

src/components/tables/AccessGrants/AccessLinks/Dialog/GenerateInvitation.tsx
travjenkins: On load the modal does not populate correctly with the default option in the dropdown.

Image

src/components/tables/AccessGrants/AccessLinks/Dialog/GenerateInvitation.tsx
```diff
@@ -21,12 +32,59 @@ import { appendWithForwardSlash, hasLength } from 'src/utils/misc-utils';
 // outside of advanced cases.
 const capabilityOptions = ['admin', 'read'];
 const typeOptions = ['single-use', 'multi-use'];
+const MAX_PREFIX_LENGTH = 12;
```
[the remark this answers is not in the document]
travjenkins: Mainly thinking of this input.
Image

Issue #1870, opened 2026-01-14
Endpoint config will show empty without error


travjenkins:
**Bug**
If JSONForms deref does not work then the endpoint config is just empty

**Expected**
We should show an error to the user and have them reach out to support

**Screenshots**

<img width="576" height="142" alt="Image" src="https://github.com/user-attachments/assets/5c61f78b-0219-49ae-9f98-a60fbcceaf9b" />

<img width="752" height="103" alt="Image" src="https://github.com/user-attachments/assets/4c2a6f6b-ff08-49b5-b0d3-b03245f70fa6" />

**Desktop (please complete the following information):**
Do not think this is related to browser/os right now

**Additional context**

PR #1872, opened 2026-01-16
Show error when endpoint cannot `deref`

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1870

### 1870

- Replacing a place where we call `set` from a proxy object
- I think this is the last `produce => state.set...()` call in the system.

### Manually tested

- Ran a `npm audit fix` and then removed the `queryString` node polyfill and then entered `create` and `edit`

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

### Create
image

### Edit
image

image

### DIFF
--- a/src/forms/renderers/MissingType.tsx
+++ b/src/forms/renderers/MissingType.tsx
@@ -21,8 +21,9 @@ const MissingTypeRenderer = (props: LayoutProps) => {
     } = props;
 
     // jsonforms - missing type
-    // This is here so when connectors like Google Sheets have an Authentication
-    //  that only has discriminator, title, and oneOf.
+    // This handles when connectors:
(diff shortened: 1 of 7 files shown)


src/stores/Binding/Store.ts
```diff
@@ -1098,13 +1098,13 @@ const getInitialState = (
     setResourceSchema: async (val) => {
         const resolved = await getDereffedSchema(val);
 
+        if (!resolved) {
+            get().setHydrationErrorsExist(true);
+            return;
+        }
```
travjenkins: [Issue Fix] - this is part of the fix. Both are needed to truly fix the issue.

Issue #1904, opened 2026-02-19
Add support for details to show yearly stats


travjenkins:
**Please propose features in the Discussions area as an `idea`**
These should show the stats per month over the last 12 months.

PR #1905, opened 2026-02-19
 Adding in a `Year` selection for data

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1904

### 1904

- Added in new `year` options
- Allowed for special "selected label" key. Showing `13 months` as the selection felt weird to me.
- Formatting `monthly` selection. This we store these as the first of the month showing a full timestamp felt very weird

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

New option for `Year`
image

These selects `13 months` so the user can compare the previous year
image

When hovering we just show `month` and `year`
image

Using `short` format for months so they fit on page even while narrow
image

Hiding the `unit` from here for `monthly` as it looked odd
image
image

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -3,6 +3,7 @@ module.exports = {
         'eslint-config-kentcdodds/import',
         'eslint-config-kentcdodds/jsx-a11y',
         'eslint-config-kentcdodds/react',
+        'plugin:storybook/recommended',
     ],
     ignorePatterns: ['vite.config.ts', '__mocks__', 'playwright-tests/'],
     plugins: ['formatjs', 'unused-imports', 'no-relative-import-paths'],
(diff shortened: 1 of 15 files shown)


src/components/filters/DetailsRange.tsx
```diff
@@ -61,7 +62,9 @@ function DetailsRange() {
             >
                 {intl.formatMessage(
                     {
-                        id: `detailsPanel.recentUsage.filter.label.${relativeUnit}`,
+                        id:
+                            selectedLabelKey ??
+                            `detailsPanel.recentUsage.filter.label.${relativeUnit}`,
```
travjenkins: Quick way to allow to custom key here as the default approach would show `13 months` when `Year` was selected.

Issue #1817, opened 2025-11-20
Field selection validation fails for previously disabled bindings


travjenkins:
## Bug

Field selection validation errors can be reported for previously disabled bindings. Presently, the client does not reliably obtain a copy of the live specification of the source collection of a previously disabled binding. This is a necessary resource for the WASM fuction, `evaluate_field_selection`, therefore the client flags its absence as an erroneous validation attempt.

### Reproduction Steps

1. Enter the edit workflow for a materialization with at least one disabled binding.

1. Enable the disabled binding.

1. Click the _Next_ CTA to update the drafted specification on the server.

1. Click the _Refresh_ CTA in the _Field Selection_ section in the right pane of the binding selector. 

1. Observe that a toast notification communicating a field selection validation error for the previously disabled binding is displayed.

## Expected

Enabling a binding should prompt the client to fetch the live specification of its source collection.

## Screenshots

<img width="801" height="779" alt="Image" src="https://github.com/user-attachments/assets/a9833bcc-e58c-41eb-adcd-3d13f66c6620" />

## Desktop
- **OS:** Windows
- **Browser:** Chrome, Edge

## Additional Context

* The first time this defect was reported in the [`front-end` Slack channel](https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1763645282592469) on Nov. 20, 2025. The defect was introduced by changes made to how source collection live specifications were fetched in [PR 1791](https://github.com/estuary/ui/pull/1791).

PR #1906, opened 2026-02-20
Fetch the live specification for source collections of previously disabled bindings

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1817

### 1817

* Allow field selection validation to be retried a single time in the event of an initial failure.

* Terminate collection hydration in the Workflow store when none of the drafted source collections need their live specifications fetched prior to the collection-related state initialization. This is partially awkward wording but the distinction is important. The Workflow state properties of concern are `collectionsHydrating` and `collectionsInited`.

* Display an error icon next to the source collection name in the binding list when field selection validation fails.

* Display a section-level, error alert in the _Field Selection_ section when field selection validation fails, replacing the toast notification.

* Expand LogRocket coverage for the field selection validation process.

### Manually tested

Approaches to testing include, but are not limited to, the following:

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

### DIFF
--- a/src/stores/Binding/slices/__tests__/FieldSelection.test.ts
+++ b/src/stores/Binding/slices/__tests__/FieldSelection.test.ts
@@ -1,4 +1,7 @@
-import type { BindingFieldSelection, StoreWithFieldSelection } from 'src/stores/Binding/slices/FieldSelection';
+import type {
+    BindingFieldSelection,
+    StoreWithFieldSelection,
+} from 'src/stores/Binding/slices/FieldSelection';
 
 import { createStore } from 'zustand/vanilla';
 
@@ -36,8 +39,8 @@ const mockSelection = (
 const createTestStore = (
     initialSelections: Record<string, BindingFieldSelection> = {}
 ) => {
-    const store = createStore<StoreWithFieldSelection>(
-        (set) => getStoreWithFieldSelectionSettings(set as any)
+    const store = createStore<StoreWithFieldSelection>((set) =>
+        getStoreWithFieldSelectionSettings(set as any)
     );
 
     if (Object.keys(initialSelections).length > 0) {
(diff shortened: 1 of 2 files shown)


src/hooks/fieldSelection/useValidateFieldSelection.ts
```diff
@@ -335,47 +377,69 @@ export default function useValidateFieldSelection() {
...
+                        if (validationEligible) {
+                            setValidationFailure([uuid]);
 
-                                setValidationFailure([uuid]);
+                            return;
+                        }
 
-                                return;
-                            }
+                        advanceHydrationStatus('VALIDATING', uuid);
```
travjenkins: I am still just doing static review... can this be hit?

It is a `let` so there is a possibility something in between really messes with the array... but all the `filter` calls I see _seem_ alright 🤷  and wouldn't change this always being true.

Image

src/hooks/fieldSelection/useValidateFieldSelection.ts
```diff
@@ -258,16 +277,35 @@ export default function useValidateFieldSelection() {
...
-                        formFailureDetected: failureDetected,
-                        missingServerData: true,
-                        targetCollection: meta.collectionName,
-                    });
+                if (
+                    meta.disable ||
+                    targetBindingContext.find(
+                        (context) => context.uuid === uuid
+                    )?.validationAttempts ===
+                        MAX_FIELD_SELECTION_VALIDATION_ATTEMPTS
```
travjenkins: To be safer we should really check greater/equal.

travjenkins (approved):
lgtm - paired with Kiahna.

PR #1934, opened 2026-03-20
fix codegen config

GregorShear:
I don't know what was going on the first time around with this - the documented config wasn't working, now it is. This is cleaner.

The thing that caught my attention was that the [invite links PR](https://github.com/estuary/ui/pull/1921) isn't failing CI even though it's expecting `inviteLinks` query which hasn't been released in production yet


travjenkins (approved):
lgtm

PR #1949, opened 2026-04-03
Switch `connectors` over to using GQL

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1929

### 1929

- We now fetch all of the connector details from GQL on load of edit/create and the admin section (test page still uses PostgREST). This means we stop making extra calls for one offs like resource schema. This is the new `useConnectorTag` context.
- More forceful with checking that connectors are valid. The GQL endpoint requires a lot more info so the UI _should_ be good to trust the responses more. This impacts things like the "valid connector" wrapper and the `hasValidConnectors` hook we used to need.
- `test/jsonforms` is now available in production
- No longer tracking `unsupportedConnector` as we just fired an event for that and didn't show anything to the user.
- Cleaned up the old `details` hydrator as only the test page was using that

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Edit now shows the connector image tag in the URL
image

Connector cards still showing as they did before
image


src/components/capture/Create/index.tsx
```diff
@@ -62,12 +54,6 @@ function CaptureCreate() {
         }
     }, [mutateDraftSpecs, mutate_advancedEditor]);
 
-    // Reset the catalog if the connector changes
-    useEffect(() => {
-        setDraftId(null);
-        setInitiateDiscovery(true);
-    }, [setDraftId, setInitiateDiscovery, imageTag]);
-
```
travjenkins, line 70: Since we prevented users from changing the connector awhile ago I think this is totally safe to remove.

src/components/connectors/Grid/ConnectorCards.tsx
```diff
@@ -40,23 +40,53 @@ export default function ConnectorCards({
...
+            );
+
+        if (!searchQuery) return nodes;
+
+        const q = searchQuery.toLowerCase();
+        return nodes.filter(
+            (node) =>
+                node.title?.toLowerCase().includes(q) ||
+                node.shortDescription?.toLowerCase().includes(q)
+        );
```
travjenkins, line 76: We now need to do some of the filtering client side.

src/components/shared/Entity/DetailsForm/useFormFields.ts
travjenkins, line 1: We used to need to set stuff here but seems it is good now.

Only thing to call out is the `setConnector` used to have a side effect of setting `setEntityNameChanged`. I think Kiahna is right and this was back when the connector could change and we needed to set the name changed because the connector being suffixed was going to change.

src/context/Router/index.tsx
travjenkins, line 1: Connector folks could use the jsonforms test page in prod.

src/components/shared/guards/SupportRole.tsx
travjenkins, line 1: New basic guard for support role. Potentially we'l add a special error or something here but not major concern right now.

src/validation/index.ts
```diff
@@ -29,7 +29,7 @@ export const DATE_TIME_RE = new RegExp(
     /^([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z)$/
 );
 
-export const MAC_ADDR_RE = new RegExp(/^([0-9A-F]{2}:){7}([0-9A-F]{2})$/i);
+export const MAC_ADDR_LIKE_RE = new RegExp(/^([0-9A-F]{2}){7}([0-9A-F]{2})$/i);
```
travjenkins: GQL returns UUIDs without the colon vs Postgrest that returns the UUIDs with colons.

kiahna-tucker (approved):

Issue #1955, opened 2026-04-21
Show `backfill warning`  on individual bindings


travjenkins:
**Please propose features in the Discussions area as an `idea`**
Want to show the "backfill from the Capture" warning on binding backfill button like we do with the "backfill all" button.

https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1776733551947019

PR #1956, opened 2026-04-21
Show `backfill warning` on individual binding

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1955

### 1955

- Added message to both the all binding and individual binding
- Made a component should it would be shared

### Manually tested

- Edit for Capture and Materialization
- Create just to be safe nothing leaked out

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

image


src/components/editor/Bindings/Backfill/BackfillDescription.tsx
```diff
@@ -0,0 +1,35 @@
...
+import AlertBox from 'src/components/shared/AlertBox';
+import { useEntityType } from 'src/context/EntityContext';
+
+interface Props {
+    allBindings?: boolean;
+}
+
+export default function BackfillDescription({ allBindings }: Props) {
+    const intl = useIntl();
+    const entityType = useEntityType();
```
GregorShear, line 14: I'm wishing there were a way for `allBindings` and `entityType` to both be props or both accessed via hooks, but moving `useEntityType()` to the parent doesn't really feel like an improvement...

just thinking out loud - no blockers here

GregorShear (approved):
Tested on my local, seems good.

it took me a while to figure out why there are two subtly different backfill options - pretty confusing, but not necessarily in scope for this PR.

Image

jwhartley:
It would be good to update the copy for the top level / all bindings backfill too
I noted in [slack](https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1776826480368589?thread_ts=1776733551.947019&cid=C02G25WJCNS) that "If you need to backfill, we recommend doing so from the Source since it's faster and uses less data"  is only correct if you have a lot of updates to existing records, it doesn’t apply for append-only, plus you lose history if it’s a snapshot source backfill (select * from existing data, vs streamed changes over time)

src/components/editor/Bindings/Backfill/BackfillDescription.tsx
travjenkins: I am hoping with React 18 and it helping reduce extra renders when using context we could start moving more things like `allBindings` into context so we could use hooks.

Issue #1737, opened 2025-08-13
Require targetNaming selection on materialization creation


jwhartley:
Following up from [slack thread](https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1752715784776499), can we make the selection of a Target Resource Naming required when setting up a materialization. 

Currently it defaults to `mirrorSchemas` (for materializations that support `x-schema-name`), but many users have not gotten the results they expected and needed to backfill. Since there are many customers that don't want this default, and it's complex UX, forcing users to choose their desired option should avoid some misconfigured materializations.

PR #1957, opened 2026-04-21
Support new `target naming` in materializations

travjenkins:
## Issues

https://github.com/estuary/ui/issues/1737

### 1737

- Added `target_naming.md` to cover migrations stuff and scenarios I covered
- Upgrade `flow_web` dep
- Create a reusable `target naming` component and hooks.
- Wired up with confirmation dialog to force users to select how to name things
- Moved the setting down in the advanced spec editor
- Hiding the old method while leaving it there if someone needs it
- Moved the prefilling source capture up into the gate

### Misc

- Add support for `MessageWithEmphasis` to support "hard coded" content
- We had a bunch of random `700` `fontWeights` so switched those over to `500`

### Manually tested

- please see `target_naming.md` for manual tests

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

### Editing 
Once there is a setting you can modify  from advanced options
image

### Using `materialize` 
User gets prompted
image

After confirming the sourceCapture and collections are populated
image


src/components/editor/Bindings/UpdateResourceConfigButton.tsx
```diff
@@ -1,107 +1,103 @@
...
+                        />
+                    ),
+                },
+                true
+            );
+
+            if (!confirmed) return;
+            await handleConfirm(defaultStrategyRef.current, () =>
+                close(defaultStrategyRef.current)
+            );
```
travjenkins, line 87: This will be the common approach for showing the confirmation with the form.

src/components/materialization/OnIncompatibleSchemaChange/index.tsx
travjenkins, line 1: With some of the new sections and options displaying next to each other wanted to keep headers looking a bit more consistent.

src/components/materialization/targetNaming/FormContent.tsx
travjenkins, line 1: This component has a bunch of state within because that makes it easier to handle if the user `cancel`s the selection. Also, allows them to tinker with the options/settings without messing up the store.

src/components/materialization/targetNaming/FormContent.tsx
```diff
@@ -0,0 +1,271 @@
...
+                    invalidSettingsMessageId="specPropUpdater.error.message.withRemove"
+                />
+            ) : null}
+
+            <RadioGroup
+                value={strategyKey}
+                onChange={(e) => setStrategyKey(e.target.value as StrategyKey)}
+            >
+                <Stack spacing={2} sx={{ width: '100%' }}>
+                    <StrategyOption
```
travjenkins, line 137: These options repeat a fair amount - but change _just_ enough that I felt it was okay to repeat it a bit. Still not my favorite but it works 🤷

src/components/materialization/targetNaming/UpdateWrapper.tsx
travjenkins, line 1: This is the button that opens up the modal. This is the only place we use the entire Modal that is in the `targetNaming` directory - everything else just uses the form component.

src/hooks/materialization/useApplyCollectionSelections.ts
travjenkins, line 1: Previously we had some similar/repeated code to handle adding collections. These were different because in the past adding a collection through setting `sourceCapture` never had to worry about settings and prefilling resource configs with WASM.

Now that we are getting `targetNaming` we need to ensure these both add collections the same way.

src/stores/Binding/PrefillSourceCaptureGate.tsx
```diff
@@ -0,0 +1,72 @@
...
+            applyCollectionSelections(strategy, response);
+        }
+
+        setDialogHandled(true);
+    };
+
+    return (
+        <>
+            {children}
+            {!dialogHandled && shouldShowDialog ? (
```
travjenkins: Safer to just render the `children` right away and then include the dialog if needed. Otherwise the child hydration gets a bit wonky.

src/components/collection/Selector/Add/index.tsx
```diff
@@ -61,6 +61,10 @@ function BindingsEditorAdd({
         setOpen(typeof args === 'boolean' ? args : !open);
     };
 
+    const handleAddClick = () => {
+        toggleDialog(true);
+    };
```
kiahna-tucker: Is this function definition necessary? If this function is only invoked once, why not pass an inline arrow function to the tooltip `onClick` method?

src/components/content/MessageWithEmphasis.tsx
```diff
@@ -4,23 +4,23 @@ import { FormattedMessage } from 'react-intl';
...
-                    <b>
-                        <FormattedMessage id={`${messageID}.emphasis`} />
-                    </b>
-                ),
-            }}
+            values={
+                emphasisContent ?? {
+                    emphasis: (
+                        <b>
+                            <FormattedMessage id={`${messageID}.emphasis`} />
```
kiahna-tucker: Would you like to replace this instance of `FormattedMessage` at this time?

kiahna-tucker (commented):
It is minor but I do wonder whether _Set Source Capture_ would be a more fitting button label than _Select Source Capture_ when the _Destination Layout_ dialog opens in the source capture flow.  In the previous step, the user _selects_ a capture from a table of options so I could see grounds for light confusion with the current button label. This is not something that needs to be changed but I did want to call it out.

### References

**_Destination Layout_ modal within the context of the source capture flow | Primary CTA**

Image

kiahna-tucker (approved):

PR #1981, opened 2026-05-26
gql/inviteLinks: Remove `Capability` enum type reference

jshearer:
## Summary

The `CreateInviteLink` mutation declared `$capability: Capability!`, embedding the GraphQL enum type name in the query string. The upcoming [authz refactor](https://github.com/estuary/flow/pull/2936) renames `Capability` to `LegacyCapability`, which would cause this query to fail.

This change replaces the typed variable with a fixed per-value mutation that inlines the capability as an unquoted GraphQL enum literal (e.g. `capability: admin`), so the query body no longer references the enum type by name. After this lands, the backend rename can proceed without a coordinated UI deploy.

## Implementation

* `src/api/gql/inviteLinks.ts`: three static mutations (one per allowed value), exported as `CREATE_INVITE_LINK_BY_CAPABILITY` map. Static mutations rather than `${capability}` interpolation because the `client` codegen preset extracts query strings statically.
* `isCreateInviteLinkCapability` runtime guard prevents any value outside `admin | read | write` from reaching the map lookup.
* `GenerateInvitation.tsx`: selects the mutation by current capability state and drops `capability` from the variables payload. The `capability` `useState` declaration moved above `useMutation` to satisfy hook ordering.

Other queries that select `userCapability` fields and the hand-written `Capability` type import in `GenerateInvitation.tsx` are intentionally untouched. Those are wire-format stable across the rename and only need a codegen rerun in a follow-up PR.


GregorShear (approved):

jshearer:
closing because https://github.com/estuary/flow/pull/2936 no longer renames `Capability`, as `flowctl` depended on it as well and it was going to be too annoying to change

PR #1990, opened 2026-06-03
Refresh token UI to use GQL

GregorShear:
## Summary

Migrates the refresh token UI from PostgREST to GraphQL.

## Test plan

- [ ] Verify refresh token pagination (10 tokens/page)
- [ ] Create a new refresh token with a description - confirm the new token appears and secret can be copied
- [ ] Dismiss and reopen the dialog - confirm it resets to the create form
- [ ] Revoke a token - confirmation dialog appears and the token is removed after confirming
- [ ] Expire a token (edit its updated_at to over a year ago) - the row should be disabled and say "expired"
- [ ] Verify error states render correctly (e.g., create failure, revoke failure, list load failure) and clear when dialog is dismissed and reopened


src/components/admin/Api/RefreshToken/CreateDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        setToken(encodedToken);
+        onCreated();
+    };
+
+    return (
+        <Dialog
+            open={open}
+            onClose={token ? undefined : onClose}
```
jshearer: Claude review found this, I think it's quite minor mostly because refresh tokens last for a long time so even if you are shown an old token it doesn't really matter, but it's probably an easy fix

---

`onClose={token ? undefined : onClose}` disables backdrop/Escape dismissal only once `token` is set. During the in-flight create mutation `token` is still `''`, so Escape or a backdrop click closes the dialog mid-request (the X button is correctly gated on `generating`, but this path is not). When the awaited `createRefreshToken` then resolves, `setToken(encodedToken)` (line 91) runs on the now-closing dialog and `onExited: resetDialog` (line 106) clears it, so the token was created server-side but its secret is never shown. Worse second-order effect: because the dialog is persistently mounted in `Table.tsx`, the post-await `setToken` can land *after* the reset, leaving `token` populated while `open=false`; the next time the user opens the Create dialog it renders the previous token's secret instead of a fresh form. Fix: gate dismissal on the in-flight state too, e.g. `onClose={token || generating ? undefined : onClose}`.

src/components/admin/Api/RefreshToken/Table.tsx
```diff
@@ -0,0 +1,230 @@
...
+                    <TableHead>
+                        <TableRow>
+                            <TableCell>
+                                <FormattedMessage id="entityTable.data.created" />
+                            </TableCell>
+                            <TableCell>
+                                <FormattedMessage id="admin.cli_api.refreshToken.table.column.label" />
+                            </TableCell>
+                            <TableCell>
+                                <FormattedMessage id="entityTable.data.status" />
```
jshearer: I have not run this PR locally, so sorry if this isn't actually true or not (also from Claude):

---

Headers render `Created / Label / Status / (blank) / (blank)` (lines 133-142) but each row renders `created / detail / uses / expired-status / revoke` (TimeStamp, then lines 55 and 63). So "Used N times" sits under the **Status** header and the actual "Expired" status sits under a blank header, on every row. The body has one more labeled content column than the head accounts for. Fix: give the uses column its own header and move "Status" over the expired column.

src/components/admin/Api/RefreshToken/Table.tsx
```diff
@@ -0,0 +1,230 @@
...
+                        {fetching && refreshTokens.length === 0 ? (
+                            <TableRow>
+                                <TableCell
+                                    colSpan={5}
+                                    sx={{ textAlign: 'center' }}
+                                >
+                                    <FormattedMessage id="common.loading" />
+                                </TableCell>
+                            </TableRow>
+                        ) : refreshTokens.length === 0 ? (
```
jshearer: On a load error the table shows the error banner and the empty state at the same time

---

On query failure `useRefreshTokens` returns `DEFAULT_TOKENS` (`[]`) and `fetching` is false, so the `error` banner (line 122) renders and control falls through to the `refreshTokens.length === 0` branch (line 156), which shows "No refresh tokens found / Create one now" directly under the red error text. The empty-state branch should also check `!error`. The deleted `EntityTable` UI avoided this by distinguishing `NETWORK_FAILED` from `NO_EXISTING_DATA`.

src/components/admin/Api/RefreshToken/Table.tsx
```diff
@@ -0,0 +1,230 @@
...
+                                        <FormattedMessage id="admin.cli_api.refreshToken.table.noContent.cta" />
+                                    </Typography>
+                                </TableCell>
+                            </TableRow>
+                        ) : (
+                            refreshTokens.map((row) => (
+                                <Row
+                                    key={row.id}
+                                    row={row}
+                                    onRevoked={
```
jshearer: The sibling `AccessLinksTable` recovers from an emptied page in a `useEffect` keyed on the post-fetch length (`if (currentPage > 0) goToPage(currentPage - 1)`, lines 90-103). This table instead decides at revoke time: `onRevoked = currentPage > 0 && refreshTokens.length === 1 ? () => goToPage(currentPage - 1) : undefined`. Two problems follow:
- That callback lives on a `Row`, and the footer (the only Previous control) is gated on `refreshTokens.length > 0` (line 197). If a page at `currentPage > 0` ever returns zero edges (a concurrent revoke in another session, TTL refetch, or future server-side filtering), no `Row` mounts, `onRevoked` can't fire, and there is no Previous button, so the user is stranded on a blank page.
- The page-back assumes the revoked row disappears from the list. Whether it does depends on whether the `refreshTokens` query returns revoked tokens. The UI is built to render `expired` rows dimmed, which suggests revoked tokens *stay* (shown as "Expired") — in which case the page never actually empties and the back-navigation fires spuriously. If instead the backend excludes revoked tokens from the list, the heuristic is right but the stranding case above still stands.

Adopting the `AccessLinksTable` `useEffect` pattern resolves both regardless of which list semantics the backend uses. Worth confirming the intended behavior: does `refreshTokens` return revoked tokens?

src/components/admin/Api/RefreshToken/RevokeDialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const [saving, setSaving] = useState(false);
+    const [error, setError] = useState<CombinedError | null>(null);
+
+    const revokeToken = async () => {
+        setSaving(true);
+        setError(null);
+
+        const result = await revokeRefreshToken({ id });
+
+        if (result.error) {
```
jshearer, line 39: Does this ever return false? I kind of doubt it...

---

`revokeRefreshToken` is typed `Boolean!` and the schema documents "already-zeroed (revoked) tokens are treated as not found", but `revokeToken` only branches on `result.error` and never reads `result.data?.revokeRefreshToken`. If the not-found path returns `false` with no GraphQL error, the dialog reports success on a no-op (e.g. revoking a token already revoked elsewhere). This is partly inherited from the old PATCH-based flow and the cache invalidation makes the end state roughly correct, so it's low impact, but checking the boolean alongside `result.error` is the defensive fix. Confirming requires the resolver's not-found contract (returns `Ok(false)` vs raises an error).

jshearer (commented):
Got some almost entirely 🤖🦾  review feedback for ya

jshearer (approved):
Once you address the above review feedback this should be good to go 👍

src/components/admin/Api/RefreshToken/Table.tsx
GregorShear: query filters out revoked tokens, but yeah, this is a better approach.

src/components/admin/Api/RefreshToken/RevokeDialog.tsx
GregorShear: oh yeah - it can technically return false if you revoke and already-revoked token, but all we care about in this case is whether the request succeeds (i.e. the user's intent either came true ...or already was true)

PR #1992, opened 2026-06-06
Move user menu, create org menu

GregorShear:
## Summary

Centralizes how the selected organization is chosen and reorganizes the side nav. The org-selection behavior is the substantive change; the nav rework is structural/visual with no functional impact.

## Functional changes

- **Tenant selection is bootstrapped app-wide** via `useInitializeSelectedTenant` (run once from `TenantGuard`), instead of as a side effect of a page-level dropdown.
- **One tenant switcher:** the side-nav org switcher replaces the per-page selectors on the home and admin pages.
- **Full-page errors always offer logout + reload**, so any error screen (including legal-check) can recover.

## Secondary (no behavior change)

Help/user menus moved into the side nav; all rows unified under one `ListItemLink`; dark/light toggle moved into the user menu.

ezgif-524cddc361fdc64e

user menu, now with more dark mode button:
Screenshot 2026-06-06 at 1 27 34 PM

org menu, currently only for switching tenants
Screenshot 2026-06-06 at 1 27 38 PM

org menu for estuary-support users (regular customers won't see this)
Screenshot 2026-06-06 at 1 27 46 PM

Adding a logout button to the FullPageError component since the user menu (where the logout button normally lives) moved to the side nav, and that isn't rendered with FullPageError:
Screenshot 2026-06-15 at 3 40 18 PM


src/components/fullPage/Error.tsx
```diff
@@ -56,6 +57,27 @@ function FullPageError({
...
+
+                    <Button
+                        variant="outlined"
+                        onClick={() => {
+                            void supabaseClient.auth.signOut();
+                        }}
+                    >
+                        <FormattedMessage id="cta.logout" />
+                    </Button>
+                </Stack>
```
GregorShear, line 79: The user menu (and logout menu item) moved to the sidenav, which the FullPageError omits, so putting a a logout button directly in the error dialog

src/components/admin/Billing/useTenantChangeReset.ts
```diff
@@ -0,0 +1,22 @@
+import { useCallback, useEffect, useRef } from 'react';
+
+import { useBillingStore } from 'src/stores/Billing';
+import { useTenantStore } from 'src/stores/Tenant';
+
+export function useTenantChangeReset() {
```
GregorShear, line 6: This hook is transitional - it extracts the billing reset logic from the deleted TenantOptions component.

A follow-up PR will key billing data on the selected tenant making this reset unnecessary.

GregorShear:
This is actually blocked by some tenant / prefixSelector refactor (which is blocked by billing gql migration...)

PR #1995, opened 2026-06-10
Clean up stores, remove orphaned billing code

GregorShear:
No functional changes, just consolidating some store code so we can have fewer exports, and removing dead billing code


jshearer (approved):
LGTM

PR #1997, opened 2026-06-11
Migrate task-creation flows to fetch data planes and storage mappings from gql

GregorShear:
Fixes the UI bug where users can't see data planes when creating new captures, materializations, etc...

This is a narrow change that swaps gql in, but also leaves a bunch of clunky architecture in place. I anticipate a broader rethinking of this part of the UI sometime later when we can pay down (or just delete) that tech debt.


GregorShear:
i'm not confident this is an easy fix for the bug - closing for now

PR #1998, opened 2026-06-12
RR7 1/5: Bump react-router to latest v6.x (6.18 → 6.30.4)

GregorShear:
Part of the React Router v7 upgrade stack (1 of 5). **Base: `main`.**

Bumps `react-router` and `react-router-dom` from 6.18.0 to 6.30.4. Same major version, no API changes — this just gets us onto current v6 so the six v7 `future` flags (introduced across later 6.x minors) are available for the next steps.

- React is already 18.3.1, which clears the v7 React-18 requirement.
- Typecheck clean; lockfile diff is small.


jshearer (approved):
👍

PR #2003, opened 2026-06-16
Key billing data on the selected tenant

GregorShear:
## What
Route billing invoice fetching through a single tenant-keyed hook (`useBillingInvoices`) instead of an imperative fetch into the Billing store.

## Why
Switching tenants required manually resetting the billing store (`useTenantChangeReset`) so the previous tenant's data wasn't left on screen. Keying the fetch on the selected tenant makes that reset unnecessary — the data refetches and supersedes on its own, matching how `useMonthlyUsage` / `AlertSubscriptions` already work.

## Changes
- New `useBillingInvoices` hook: fetches the selected tenant's invoices for a rolling six-month window and returns `{ invoices, selectedInvoice, isLoading, networkFailed, errorExists }`. `selectedInvoice` falls back to the newest invoice when the stored selection isn't in the current tenant's data, so an org switch self-corrects.
- Slim the `Billing` store to just the invoice selection and `paymentMethodExists`; drop `invoices`, the `StoreWithHydration` flags, and `resetState`.
- Delete `useTenantChangeReset` and the unmount reset in `AdminBilling`.
- Point every consumer (history table, line items, usage graphs, graph-state wrapper, load error) at the hook.

Still PostgREST under the hood — transport is unchanged here; the GraphQL migration follows in the stacked PR. Behavior-preserving, no consumer API changes.

Stacked on `greg/move-avatar`.


GregorShear:
stale... closing

PR #2008, opened 2026-06-19
Remove dead code from ListItemLink

GregorShear:
Simplify and delete dead code from ListItemLink. Remove react-intl


jshearer (approved):
LGTM

PR #2009, opened 2026-06-19
Deprecate i18n props

GregorShear:
Any component that expects a react-intl i18n id as an input prop is now deprecated, and a "pure" version of the component is a named export from the same file with a string literal prop. 

No functional changes, and it shouldn't break any existing callers.


jshearer (approved):
LGTM % merge conflict

PR #2014, opened 2026-06-23
Fix data preview "Server Error" for collections with more than 25 journals

jshearer:
## What

Data Preview showed `Server Error` instead of sample documents for any collection split into more than 25 journals.

## Root cause

Gazette's `List` RPC is server-streaming and returns the journal list as one or more newline-delimited `{result: ListResponse}` frames. gazette/core#466 lowered the per-frame cap (`maxJournalsPerListResponse`) from 1000 to 25, so once that build shipped, collections split more than 25 ways began coming back as multiple frames. The UI's `getJournals` parsed the body with a single `response.json()`, which handles only one frame and threw on the second. That surfaced as the generic `FETCH_DEFAULT_ERROR` ("Server Error"), and because the list call failed the preview never selected a journal to read.

## Fix

estuary/data-plane-gateway#51 fixes `JournalClient.list()` to drain the stream and merge every frame (matching what `read()` already does), returning the full `ProtocolListResponse`. This PR vendors that build and migrates onto it:

- Bump the vendored `data-plane-gateway` tarball.
- `useJournalsForCollection` calls `journalClient.list()` and reads `.journals` directly.
- Remove the now-dead `getJournals` and `isNestedProtocolListResponse` helpers.

Data preview still samples a single journal as before; handling multiple frames just lets the list parse.

Depends on estuary/data-plane-gateway#51.


src/hooks/journals/useJournalData.ts
GregorShear, line 105:
```suggestion
                    shouldRefreshToken(error?.message) &&
 ```

`fetcher` only ever rejects with an object so we'll be calling `shouldRefreshToken("[object Object]")` which will return false and `refreshAuthToken()` is never reached

File: src/hooks/journals/useJournalData.ts [hunk 1]
```diff
@@ -54,31 +51,30 @@ const useJournalsForCollection = (collectionName: string | undefined) => {
...
 
-                if (isEmpty(dataPlaneListResponse)) {
-                    return Promise.reject(dataPlaneListResponse);
+                    return {
+                        journals: result.unwrap().journals ?? [],
+                    };
+                } catch {
+                    // The gateway client throws on transport-level failures it
+                    // cannot parse into a ResponseError, e.g. a network error.
+                    return Promise.reject({ message: FETCH_DEFAULT_ERROR });
```
GregorShear, line 76:
```suggestion
                } catch (error: unknown) {
                    // The gateway client throws on transport-level failures it
                    // cannot parse into a ResponseError, e.g. a network error.
                    logRocketConsole('journals:list:error', error);
                    return Promise.reject({ message: FETCH_DEFAULT_ERROR });
```

failing calls to `getJournals` would be logged by the `client()`, so we could preserve that logging here

Issue #1960, opened 2026-04-28
Task alert configuration in admin


travjenkins:
**Please propose features in the Discussions area as an `idea`**

PR #2015, opened 2026-06-23
Update alert subscription UX

kiahna-tucker:
## Issues

https://github.com/estuary/ui/issues/1960

### 1960

-   changes...

### Manually tested

### Automated tests

-   unit testing covered

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

**NOTE:** Code that altered button label styling in the application was merged into this branch **_after_** screenshots were taken for this PR. The button labels in the _Organization Notifications_ section of the _Settings_ tab on the _Admin_ page are in title case.

### Table

**_Organization Notifications_ table**

pr_screenshot-2015-alert_subscription_ux-table

<br />
<br />

### Dialogs

**_Configure Notification Methods_ dialog | Initialized**

pr_screenshot-2015-alert_subscription_ux-dialog-configure-initialized

<br />
<br />

**_Configure Notification Methods_ dialog | Subscription summary expanded**

pr_screenshot-2015-alert_subscription_ux-dialog-configure-subscription_summary-expanded

<br />
<br />

**_Update Notification Methods_ dialog | Initialized**

pr_screenshot-2015-alert_subscription_ux-dialog-update-initialized

<br />
<br />

**_Update Notification Methods_ dialog | Subscription summary expanded**

pr_screenshot-2015-alert_subscription_ux-dialog-update-subscription_summary-expanded

<br />
<br />

**_Recipients_ section | No recipients found**

pr_screenshot-2015-alert_subscription_ux-dialog-no_subscribers

<br />
<br />

### _Alert Types_ List Item Interactions

**Standard alert type | Deselected**

pr_screenshot-2015-alert_subscription_ux-dialog-alert_types-item-deselected

<br />
<br />

**Standard alert type | Deselected | Hovered**

pr_screenshot-2015-alert_subscription_ux-dialog-alert_types-item-deselected-hovered

<br />
<br />

**Standard alert type | Selected**

pr_screenshot-2015-alert_subscription_ux-dialog-alert_types-item-selected

<br />
<br />

**Standard alert type | Selected | Hovered**

pr_screenshot-2015-alert_subscription_ux-dialog-alert_types-item-selected-hovered

<br />
<br />

**System alert type**

A list item corresponding to a system alert type does not have a hover state since it cannot be deselected.

pr_screenshot-2015-alert_subscription_ux-dialog-alert_types-item-system_alert

<br />
<br />

### Errors

**Duplicate subscribed email detected | Subscriber summary expanded**

pr_screenshot-2015-alert_subscription_ux-dialog-duplicate_email

<br />
<br />

**Duplicate subscribed email detected | Subscriber summary collapsed**

pr_screenshot-2015-alert_subscription_ux-dialog-duplicate_email-collapsed

<br />
<br />

**Multiple email input errors detected | Subscriber summary expanded**

pr_screenshot-2015-alert_subscription_ux-dialog-multiple_email_errors

<br />
<br />

**Multiple email input errors detected | Subscriber summary collapsed**

pr_screenshot-2015-alert_subscription_ux-dialog-multiple_email_errors-collapsed

<br />
<br />


GregorShear:
Not sure what was discussed, but I might expect to see some of the other alert configuration options here - at least `shardFailed`.

Screenshot 2026-07-06 at 12 51 46 PM

Related to this, the UI doesn't appear to show any config inherited from parent prefixes. For example, if I configure a 24 hour interval for `dataMovementStalled` on the acmeCo/ prefix, when i open the notification dialog to acmeCo/nested/ , the `dataMovementStalled` dropdown says "None", even though the parent prefix is in effect there.

(and i'm also seeing now that if i save that acmeCo/nested/ dialog without making any changes, both parent configs and subscriptions are saved as new rows with the acmeCo/nested/ prefix)

kiahna-tucker (commented):
> Not sure what was discussed, but I might expect to see some of the other alert configuration options here - at least shardFailed.

The data movement setting was the only global prefix setting requested at this time. Given the stage this feature is in and its UX breadth, I would advocate that support of additional, global prefix settings be considered _fast follows_.

> Related to this, the UI doesn't appear to show any config inherited from parent prefixes. For example, if I configure a 24 hour interval for dataMovementStalled on the acmeCo/ prefix, when i open the notification dialog to acmeCo/nested/ , the dataMovementStalled dropdown says "None", even though the parent prefix is in effect there.

The `EffectiveAlertConfig` is currently used to populate the _Interval_ field in the _Data Movement Stalled_ section. This object is reported to correspond to _the fully-resolved effective config at this scope, merging all ancestor prefix layers and controller defaults_. I will take another look at the network calls to ensure that the client is making requests as expected. Please let me know if different object should be used.

**UPDATE:** This behavior is expected as no row exists for `acmeCo/nested/` in the `alert_configs` database table at this time, leaving the server unaware of the new prefix until the user has clicked the _Save_ button in the _Configure Notification Methods_ dialog. The only sensible way to get around this in the absence of cleanup routines on/deletion APIs exposed by the backend would be to have a button (preferably inline) with the _Prefix_ field that would create a row in the `alert_configs` database table with an empty object in the `config` database table cell when clicked.

To my knowledge, a GraphQL endpoint that allows the client to delete rows from the `alert_configs` database table does not exist; therefore, there is a question of which entity (i.e., client or server) should be responsible for cleaning up "stale" rows in this table (i.e., a row where the `config` database table cell contains an empty object) and what the permitted scope of deletions should be. Regarding the latter open question, if a deletion API is the proper direction, it would be beneficial if the request could accept a prefix and, perhaps, a flag indicating the intent to clean up "stale" rows corresponding to as well as under that prefix. But I minimally digress.

**Tl;dr:** The GraphQL API will not return the "inherited", global settings for a prefix that does not have a corresponding row in the `alert_configs` table, resulting in this UX quirk that I raised around a month ago. At the time, I was directed to implement the interactive proposal as closely as possible, therefore I assume the broader team is comfortable addressing this post-release as a _fast-follow_ if we desire to take action.

> (and i'm also seeing now that if i save that acmeCo/nested/ dialog without making any changes, both parent configs and subscriptions are saved as new rows with the acmeCo/nested/ prefix)

This was ultimately an open UX question that is tightly coupled to that highlighted above. Allowing rows to be inserted into the `alert_configs` and `alert_subscriptions` tables for a new prefix when global prefix settings nor subscriptions are defined, allows for the new prefix to be visible in the _Organization Notifications_ table and the "inherited", global prefix settings (i.e., the effective alert config) to be viewed in _Update Notification Methods_ dialog.

GregorShear:
Along the same lines as my comment about redundant alert configs being created, if i have a `dataMovementStalled: 24h` config defined at `acmeCo/hello/world/` and I open the dialog to `acmeCo/hello/` (the parent) i see the dataMovementStalled value set to 24h even though i only ever set that value on the child. And then if i click the save button it does create the config with that value at `acmeCo/hello/`, effectively promoting that config to the parent

GregorShear:
I'm also seeing that if i make a change to the data movement stalled interval, and then click the "delete prefix alerts" button, the dialog closes and saves that config option - i could imagine this being surprising to a user. maybe we could remove the delete prefix alerts button entirely?

GregorShear:
If i change the global settings at acmeCo/ without having added a recipient, clicking "save" hangs on the loading indicator (i don't see any network requests being made)

GregorShear:
when i open the dialog and add a new recipient, the alert types respond to mouse over (with hover states) but not to mouse clicks. 

I have to interact with the email field first, and then the alert types are clickable.

GregorShear:
There's some dead space in the alert type boxes between the text and the edge of the box that responds to hover but isn't clickable at all

Screenshot 2026-07-06 at 2 51 41 PM

GregorShear:
probably unlikely but the dialog crashes if the API sends down `{"dataMovementStalled":{"enabled":true}}` (no "condition" key defined)

kiahna-tucker:
> Along the same lines as my comment about redundant alert configs being created, if i have a dataMovementStalled: 24h config defined at acmeCo/hello/world/ and I open the dialog to acmeCo/hello/ (the parent) i see the dataMovementStalled value set to 24h even though i only ever set that value on the child. And then if i click the save button it does create the config with that value at acmeCo/hello/, effectively promoting that config to the parent

> I'm also seeing that if i make a change to the data movement stalled interval, and then click the "delete prefix alerts" button, the dialog closes and saves that config option - i could imagine this being surprising to a user. maybe we could remove the delete prefix alerts button entirely?

The quoted UI/UX callouts above are consequences of the proposed and approved design I was instructed to follow closely. I made similar callouts in the past, however the most likely scenario will be that UI/UX adjustments will be made as _fast-follows_, either as other teammates encounter the friction in production or as customers do. 

> There's some dead space in the alert type boxes between the text and the edge of the box that responds to hover but isn't clickable at all.

There are 16 pixels of padding on the right and left of MUI's `ListItem` component that are not clickable. This is not a blocking and can be addressed as a fast-follow.


**Action Items:**
- [x]  when i open the dialog and add a new recipient, the alert types respond to mouse over (with hover states) but not to mouse clicks. I have to interact with the email field first, and then the alert types are clickable.
- [x] If i change the global settings at acmeCo/ without having added a recipient, clicking safe hangs on the loading indicator (i don't see any network requests being made)
- [x] probably unlikely but the dialog crashes if the API sends down {"dataMovementStalled":{"enabled":true}} (no "condition" key defined)
- [x] Implement the workaround for the `alertConfigs` GraphQL endpoint discussed in Slack.

GregorShear:
> UPDATE: This behavior is expected as no row exists for acmeCo/nested/ in the alert_configs database table at this time, leaving the server unaware of the new prefix until the user has clicked the Save button in the Configure Notification Methods dialog. The only sensible way to get around this in the absence of cleanup routines on/deletion APIs exposed by the backend would be to have a button (preferably inline) with the Prefix field that would create a row in the alert_configs database table with an empty object in the config database table cell when clicked.

I'm not following you here - there should be a way for a user to subscribe to alerts at `acmeCo/whatever/` without creating a new config row at that prefix. Creating a new config row there alongside the subscriptions defeats the inheritance scheme where the config can be managed at the parent prefix unless explicitly overridden at the child prefix. 

Hope i'm not over-explaining here - help me understand why we would need to write a new config at the child prefix? sounds like you're saying that's the intended behavior.

src/utils/notification-utils.ts
```diff
@@ -8,3 +14,82 @@ export const snackbarSettings: OptionsObject = {
...
+                    { ...query, id: crypto.randomUUID(), viewing: false },
+                ],
+            };
+        }
+    });
+
+    return subscriptionMetadata;
+};
+
+export const fromUnconventionalTimeFormat = (value: string | undefined) => {
```
GregorShear, line 44: a `dataMovementStalled` config with `"stalledFor: "30days"` will show up in the UI as "None".

Probably better to just fall back to the value sent by the server if it doesn't match one of the existing choices

src/components/admin/Settings/PrefixAlerts/Dialog/GlobalSettings/DataMovementSetting.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    {intl.formatMessage({
+                        id: 'details.settings.notifications.dataProcessing.noDataProcessedInInterval.message',
+                    })}
+                </Typography>
+            </Stack>
+
+            {loading ? (
+                <Skeleton height={38} />
+            ) : (
+                <Autocomplete
```
GregorShear, line 63: nit/IMO an Autocomplete is probably overkill here

src/components/admin/Settings/PrefixAlerts/useModifyAlertMetadata.ts
```diff
@@ -0,0 +1,177 @@
...
+                    : upsertSubscription({
+                          alertTypes,
+                          email,
+                          prefix: catalogPrefix,
+                      });
+            });
+
+        const serverErrors: (PostgrestError | CombinedError)[] = [];
+
+        Promise.allSettled(subscriptionQueries).then(
```
GregorShear:
```suggestion
        await Promise.allSettled(subscriptionQueries).then(
```

do we need an await here?

GregorShear:
i acknowledge this isn't represented in the mock, but we'll want a way to suppress alerts at a given prefix - e.g. `acmeCo/` set some `dataMovementStalled` condition, but we want to set `dataMovementStalled: { enabled: false }}` at `acmeCo/broken_prototype/`

GregorShear:
hey by the way - did you do a claude review on this? i'm thinking a lot of this stuff would be caught by automated PR review...

GregorShear:
email validation is pretty aggressive - could we wait until the user clicks away

Screenshot 2026-07-06 at 3 54 06 PM

also, the save button remains disabled until i blur the email input field

Screenshot 2026-07-06 at 3 55 40 PM

kiahna-tucker:
> Along the same lines as my comment about redundant alert configs being created, if i have a dataMovementStalled: 24h config defined at acmeCo/hello/world/ and I open the dialog to acmeCo/hello/ (the parent) i see the dataMovementStalled value set to 24h even though i only ever set that value on the child. And then if i click the save button it does create the config with that value at acmeCo/hello/, effectively promoting that config to the parent.

As aforementioned, the _Configure Notification Methods_ dialog **only** affects a single prefix. Changing the prefix will not reset the settings or subscriptions the user configured _unless_ there is existing information in the database tables to override it. This UX is peculiar, was brought to the attention of necessary parties by me, and it was decided that the UX would be released as-is and re-evaluated later on.

> email validation is pretty aggressive - could we wait until the user clicks away

Email validation is pretty aggressive as requested and agreed to in a UI/UX meeting when alert subscriptions was first introduced to the application. If you would like to see it relaxed now, that is fine but such a nit is not blocking.

> also, the save button remains disabled until i blur the email input field

This is intentional and necessary for reasons explained in dailies in May.

> i acknowledge this isn't represented in the mock, but we'll want a way to suppress alerts at a given prefix - e.g. acmeCo/ set some dataMovementStalled condition, but we want to set dataMovementStalled: { enabled: false }} at acmeCo/broken_prototype/

This is a fair point, however it is out-of-scope for this feature. Please review the interactive proposal to get a better understanding of the scope of this feature.

src/components/admin/Settings/PrefixAlerts/useModifyAlertMetadata.ts
kiahna-tucker: It is not _necessary_ but a worthwhile safeguard. Will update.

GregorShear:
> > Along the same lines as my comment about redundant alert configs being created, if i have a dataMovementStalled: 24h config defined at acmeCo/hello/world/ and I open the dialog to acmeCo/hello/ (the parent) i see the dataMovementStalled value set to 24h even though i only ever set that value on the child. And then if i click the save button it does create the config with that value at acmeCo/hello/, effectively promoting that config to the parent.
> 
> As aforementioned, the _Configure Notification Methods_ dialog **only** affects a single prefix. Changing the prefix will not reset the settings or subscriptions the user configured _unless_ there is existing information in the database tables to override it. This UX is peculiar, was brought to the attention of necessary parties by me, and it was decided that the UX would be released as-is and re-evaluated later on.

You're saying this is the intended behavior?


https://github.com/user-attachments/assets/e4103cc4-5266-4723-b786-76b465636dcd

kiahna-tucker:
No. I do not know what commit you are working off of, but it appears to be older than that of the current branch. Nevertheless, no, that is not what I am saying here. You and I are misinterpreting each other's words -- which is to be expected as we learn the other's communication styles and patterns. I believed you described behavior strictly within the dialog, therefore I described how changes to the prefix input affects the data in other dialog sections. This UX is something that _does_ feel odd to me, so I wanted to make sure you were aware that it is, presently, expected.

As stated in Slack, it does appear that we have an ethos to PRs that differ. I have not done a code review of the entire PR after the changes made yesterday. I will explicitly re-request a review in GitHub (and message you directly, if that would be beneficial) when it is time for you to take a second look at this PR.

kiahna-tucker:
@GregorShear, Here is the output of a medium, inline code review performed by Claude:

> Logic review summary
> 
> This branch refactors the PrefixAlerts feature from a single-subscription-per-dialog model to a multi-recipient-per-prefix model, and adds prefix-level global settings (data-movement config) backed by a new alertConfigs GraphQL query/mutation. I reviewed the hand-written logic across the store, hooks, and dialog components.
> 
> The logic is sound and internally consistent. Key areas worth a reviewer's attention, all of which hold up:
> 
> - Alert-config upserts are correctly gated. An empty config is written only to clear a prefix's own previously-saved settings — never for a prefix that merely inherits from an ancestor and never when no settings exist — which respects the "don't insert empty alert_configs rows" constraint. The explicit/implicit/directImplicitMatch evaluation keys off a single consistent prefix value.
> - Subscription state is keyed by stable id. Add/edit/delete/toggle operations and email/alert-type mutations all identify rows by id, so duplicate or in-progress (empty) emails don't cause the wrong row to be mutated.
> - Async save path is safe. Loading state is always reset (including early-exit paths), stale subscriptions are diffed against server state without double-deletes, and identical subscriptions short-circuit to a no-op.
> - Rendering guards are in place. Table rows guard against empty subscription arrays, and the debounced prefix evaluation is cancelled on unmount.
> 
> One cosmetic, non-blocking note: the from/toUnconventionalTimeFormat helpers in notification-utils.ts are not strict inverses ('1day' and '24h' both map to '24:00:00'), which only matters if the backend distinguishes those two values.
> 
> No correctness or blocking issues found. Note that this is a logic review only — a type-check/lint/test run is still recommended as a separate gate.

kiahna-tucker:
Latest Claude code review:

> This branch reworks PrefixAlerts from a single-subscription model to per-prefix, multi-recipient configuration with global (data-movement) settings. The logic is sound and ready to merge. Core flows were reviewed end-to-end:
> 
> - Store state management (useAlertSubscriptionsStore) — mutable/immutable subscription tracking, deletion marking, and prefix-selection handling are internally consistent; subscription IDs align correctly between the immutable source and mutable working copies.
> - Save/delete path (useModifyAlertMetadata, useUpsertAlertSubscription, useUpsertAlertConfig) — correctly distinguishes create vs. update vs. delete, skips no-op writes, and cleans up stale subscriptions.
> - Validation (useEvaluateSubscriptionIneligibility) — empty-email and duplicate-email detection correctly operate on active (non-deleted) subscriptions only.
> - Global settings — the data-movement interval round-trips correctly between the stored format and the UI option set.
> 
> No correctness defects remain. Two low-severity, non-blocking items are worth noting:
> 
> 1. Alert-config pagination — configs are fetched as a single page rather than paged (useAllPages); prefixes beyond the default page size could miss their global settings. Fine if the per-tenant config count is small.
> 2. Minor polish — one button handler returns a rejected promise with no catcher (harmless console noise; a bare return would be cleaner).

kiahna-tucker:
Latest Claude code review:

Code review summary

> This branch refactors prefix alerts from a single-subscription model to per-prefix metadata supporting multiple recipients plus prefix-level global settings. The logic is sound and ready for review.
> 
> Areas verified as correct:
> - Store reducers (useAlertSubscriptionsStore) — immer-based updates for adding/deleting/toggling subscriptions and alert types correctly distinguish DB-backed rows (marked for deletion) from client-only rows (removed outright).
> - Save reconciliation (useModifyAlertMetadata) — the stale-subscription delete + upsert flow handles email changes, deletions, and no-op-identical subscriptions without duplicate or orphaned mutations.
> - Prefix-settings inheritance — longest-matching-prefix resolution for explicit vs. inherited (implicit) global settings is internally consistent, and the config-format round-trip (from/toUnconventionalTimeFormat) is symmetric.
> - Duplicate/empty-email validation — blank emails are excluded from duplicate detection and gated separately.
> - UI plumbing — PrefixedName fixedPrefix behavior and the per-prefix table/dialog wiring are consistent; no dead props or dangling references.
> 
> One minor, non-blocking item:
> - The alert-config query (PrefixAlerts/index.tsx) fetches a single page capped at first: 100 and does not follow pageInfo.hasNextPage. This is safe for any realistic number of prefix configs; if a tenant ever exceeds 100, the overflow would be silently dropped. Optional follow-up: log when hasNextPage is true, or paginate via useAllPages.
> 
> No correctness defects block merge.

GregorShear:
> (and i'm also seeing now that if i save that acmeCo/nested/ dialog without making any changes, both parent configs and subscriptions are saved as new rows with the acmeCo/nested/ prefix)

Confirming that the config is no longer saved as a new row, but the subscription is

Screenshot 2026-07-16 at 2 28 16 PM

kiahna-tucker:
Greg and I discussed the scenario highlighted above on VC which breaks down as follows:

Let's say an alert subscription exists for `acmeCo/`.

1) Click the _Configure Notifications_ button (which opens the _Configure Notification Methods_ dialog prefilled with data related to `acmeCo/`).

1) Enter characters in the prefix field that correspond to a nested prefix (e.g., `acmeCo/anvils/`) without direct subscriptions (i.e., `alert_subscriptions` database table does not contain a row corresponding to the nested prefix).

1) Observe that the _Recipients_ section remains unchanged, reflecting the subscribers to `acmeCo/`, and the _Save_ CTA is enabled.

1) Click the _Save_ CTA.

This results in the creation of "superfluous" (or effectively duplicate) alert subscriptions for `acmeCo/anvils/`. To unsubscribe the recipient from those alerts for `acmeCo/anvils/`, the subscription at the `acmeCo/anvils/` **and** `acmeCo/` level would need to be removed.

That said, this UX quirk was presented on Jul. 10th and approved of but it is likely something we would like to address quickly. Given the prioritization of this feature and the approval, I would push for this quirk to be addressed in a rapid fast-follow.

GregorShear:
> > (and i'm also seeing now that if i save that acmeCo/nested/ dialog without making any changes, both parent configs and subscriptions are saved as new rows with the acmeCo/nested/ prefix)
> 
> Confirming that the config is no longer saved as a new row...

There's actually still a subtle issue here - setting a new value for dataMovementStalled should just save that as the config record, like: 

```
{
  "dataMovementStalled": {
    "condition": {
      "stalledFor": "2days"
    }
  }
}
```

but instead it freezes the entire effective config to the record:

```
{
  "taskIdle": {
    "enabled": true,
    "condition": {
      "idleFor": "30days"
    },
    "autoDisable": false
  },
  "shardFailed": {
    "enabled": true,
    "condition": {
      "per": "8h",
      "failures": 3
    }
  },
  "dataMovementStalled": {
    "condition": {
      "stalledFor": "2days"
    }
  },
  "taskChronicallyFailing": {
    "enabled": true,
    "condition": {
      "failingFor": "30days"
    },
    "autoDisable": false
  }
}
```

kiahna-tucker:
> Creates or replaces the alert config at `catalogPrefixOrName`.

The description for the `updateAlertConfig` mutation leads me to believe that updates can be destructive if the request does not contain the full config. However, rows inserted into the `alert_configs` table should _not_ contain all that data. Assuming that is the defect you would like to see patched, I am working on a fix.

src/components/tables/PrefixAlerts/index.tsx
```diff
@@ -69,15 +84,38 @@ function PrefixAlertTable() {
     });
 
     useEffect(() => {
-        if (!fetching && !alertTypeResponse.fetching) {
+        if (!fetching && !alertConfigResponse.fetching) {
             setInitializationError(error);
```
GregorShear: Setting this error disables the email field and alert type checkboxes - you can still click "add recipient" but the UI that expands is just non-functional. I'd rather replace the entire contents of the dialog with some "init failed" message instead of presenting a half-broken UI.

A second point here is that config query failures are silent and look like no configuration is set, but the user is still free to add new config that will overwrite and potentially delete existing config (since the API relies on the frontend to merge and send back the entire config object, as you've pointed out). My take is that either a subscription or config query error would result in the init failed state.

Image

GregorShear:
The email field loses focus if you backspace to the end.

also IIRC didn't we want to prevent editing the email of existing subscriptions?

https://github.com/user-attachments/assets/7bc6923d-df6a-4280-ab04-f29630471bc2

kiahna-tucker:
> also IIRC didn't we want to prevent editing the email of existing subscriptions?

No, a user is free to "edit" the email of an existing subscription. Changing the email field input of an existing subscription results in the same server changes as removing the subscription in the client (i.e., clicking the _X_ button in the top-right corner of subscriber summary) and adding a new one with the updated email. What I aimed to prevent was having more than one templated subscription in the client (i.e., a list item without a specified email).

> The email field loses focus if you backspace to the end.

This is a known issue in production. Happy to investigate the defect and patch it, however I would advocate that this production bug should not be blocking.

> Setting this error disables the email field and alert type checkboxes - you can still click "add recipient" but the UI that expands is just non-functional. I'd rather replace the entire contents of the dialog with some "init failed" message instead of presenting a half-broken UI.

Makes sense to me. Will update.

src/components/tables/PrefixAlerts/index.tsx
```diff
@@ -69,15 +84,38 @@ function PrefixAlertTable() {
...
+            setSubscriptionMetadata(data?.alertSubscriptions ?? []);
```
GregorShear: I don't quite understand why we're re-saving the urql cache to another store? couldn't we just interact with the query cache directly?

src/components/admin/Settings/PrefixAlerts/Dialog/index.tsx
GregorShear, line 51: this feels like a postgrest vestige. why do we need a globally scoped state to begin with? urql cache + an edit buffer that lives and dies with the dialog should be sufficient and much simpler, no?

src/components/tables/PrefixAlerts/index.tsx
kiahna-tucker: Query data would be prop drilled considerably to reach the components dependent upon it, needlessly increasing the sensitivity of higher order components. Actions within the store are able to reference relevant query data without the need for their callers to pass the data and, thus, increasing their sensitivity to changes to the data.

Happy to chat through refactor opportunities but I would push for this feature to get out the door given the prioritization.

src/components/admin/Settings/PrefixAlerts/Dialog/index.tsx
kiahna-tucker: In the Control Plane/UI Sync meeting on Monday, let's chat about how we would like to interact with GraphQL.

As for this PR, alert configs used to be fetched when the debounced prefix input changed yet still stored at that time. In light of the limitation of the `alertConfigs` endpoint that resulted in the need for the control-plane workaround, the timing of the call was adjusted but, yes, the data is still stored in a Zustand store. There are some benefits to that and, to my knowledge, the GraphQL query key issue is still present in the repo that would require the query data to be considerably prop drilled. I am not opposed to a refactor if you see a better approach to working with GraphQL data. However, given the urgency around this feature, I would like to land this ASAP and revisit at this point.

src/components/tables/PrefixAlerts/index.tsx
kiahna-tucker: This should be handled. However, in a second pass, I would like for the user to be able to retry these queries. Perhaps we could discuss this in the Control Plane/UI Sync meeting on Monday.

GregorShear:
One thing that would be worth doing (and is hopefully a small change) is replacing the hand-written gql queries with the codegen types

kiahna-tucker:
> At first glance, the return type of the graphql function in the client always reflects the broadest response shape. Is there a way for the implicit return type to reflect the properties selected by a query? Or would you like for GraphQL queries to always return the full response object and the client can narrow them at a lower level (i.e., when useQuery is called in a hook or component)?

For the record, alert-related queries use GraphQL interfaces fetched and exported by `codegen`. Most of these queries only select the data that is actually relevant to the client but the `graphql` function returns an object that always reflects the broadest response shape, resulting in a misleading type annotation. To make a change similar to that observed in [PR 2027](https://github.com/estuary/ui/pull/2027), answers to the questions above are required.

Where does this PR stand otherwise? I would like to see it get merged within the next 24 hours, if possible.

GregorShear:
I'm finding more strange behavior where a parent config leaks into a child prefix. for example, if a parent has this config:

```json
{
  "shardFailed": { "condition": { "failures": 5 } },
  "dataMovementStalled": { "condition": { "stalledFor": "3days" } }
}
```

and the user edits an existing child prefix, changes the `dataMovementStalled` and saves, that child config ends up with `shardFailed` defined, too.

GregorShear:
- duplicate errors when typing in the prefix field with no network connection: Screenshot 2026-07-31 at 11 33 57 PM
- clicking into and out of the prefix field resets the entire dialog
ezgif-8875f8062ec3c0c6

- email field blurs one second after last character typed - i think you've already called it out as a compromise
- alert select box padding responds to hover, but not click (i already called this out before)
- system alert checkbox responds to hover and click, but doesn't change state Screenshot 2026-07-31 at 11 39 26 PM

GregorShear (dismissed):

kiahna-tucker:
> duplicate errors when typing in the prefix field with no network connection

> system alert checkbox responds to hover and click, but doesn't change state

These have simple fixes. Happy to do so before release.

> alert select box padding responds to hover, but not click (i already called this out before)

For context, there are 16 pixels of padding on the left and right _edges_ or _ends_ of the list item that are not click-responsive.

> email field blurs one second after last character typed - i think you've already called it out as a compromise

Debouncing the input is a temporary solution to allow the field to capture or commit the input without the user needing to press the Tab/Enter keys or invoke the `onBlur` method by clicking elsewhere in the dialog. Personally, I am not fond of debouncing the field input; I would rather move forward with the previous field interactions for an initial release. My hunch is that the subscription client state, unfortunately, needs to following the pattern that exists in binding client state -- which I had originally hoped would be avoidable.

> clicking into and out of the prefix field resets the entire dialog

I would like to better understand the scope of this once I can get online. In the example you've shown, the _templated_ subscription is the only thing affected. If that is the case, the scope of impact is considerably smaller than the comment alludes and worth noting when assessing the UX of an initial release.

kiahna-tucker:
> duplicate errors when typing in the prefix field with no network connection

> system alert checkbox responds to hover and click, but doesn't change state

> clicking into and out of the prefix field resets the entire dialog

@GregorShear, the three commits directly above intend to address the callouts quoted directly above in this comment. Please let me know if you still see issues along those lines. Additionally, could you let me know your stance on whether the email field would be best served without having its input debounced? I, personally, find the debounce disruptive.

**UPDATE:**

> alert select box padding responds to hover, but not click (i already called this out before)

The commit directly below intends to address the callout quoted directly above in this comment.

GregorShear (approved):

PR #2017, opened 2026-07-01
update Estuary IP fallback list

jwhartley:
Updates the hardcoded IP fallback shown when a data plane's CIDR list is unavailable, so it matches current configuration.


skord (approved):
LGTM

PR #2018, opened 2026-07-01
Remove inaccessible "express" feature

GregorShear:
I anticipate revisiting this feature, and referencing this work, but for now we can remove a lot dead code


jshearer (approved):
Happy to merge as is, deleting code/cleaning up is always great. That being said, Claude found a few dangling threads that could net even more deletions!

---

**`Workflow` store still carries express-only state**

`customerId`, `redirectUrl`, and their setters (`setCustomerId`, `setRedirectUrl`) have zero remaining callers after this commit. `ExpressWorkflowGuard` was the only writer for both.

https://github.com/estuary/ui/blob/6d0f8f55d298f380f12b21f492429f926af00400/src/stores/Workflow/types.ts#L19-L25

https://github.com/estuary/ui/blob/6d0f8f55d298f380f12b21f492429f926af00400/src/stores/Workflow/Store.ts#L76-L94

**That leaves an unreachable branch in `useEntityWorkflowHelpers`**

Since `redirectUrl` can never be set again, `externalRedirect` in `HeaderLogs.tsx` is always `false`, so the `window.location.href` branch in `exit()` can't fire for any surviving flow. `HeaderLogs` is the only caller that ever passes `external=true`.

https://github.com/estuary/ui/blob/6d0f8f55d298f380f12b21f492429f926af00400/src/components/shared/Entity/HeaderLogs.tsx#L23-L31

https://github.com/estuary/ui/blob/6d0f8f55d298f380f12b21f492429f926af00400/src/components/shared/Entity/hooks/useEntityWorkflowHelpers.ts#L87-L109

**`docs/STATE.md` still documents the flow**

https://github.com/estuary/ui/blob/6d0f8f55d298f380f12b21f492429f926af00400/docs/STATE.md#L100-L107

**Orphaned translation string**

`expressFlowIntegration` had exactly one consumer, `PoweredByEstuaryWatermark`, which this commit deletes.

https://github.com/estuary/ui/blob/6d0f8f55d298f380f12b21f492429f926af00400/src/lang/en-US/CommonMessages.ts#L6

PR #2019, opened 2026-07-01
Add code-health PR check (dead-code delta)

GregorShear:
## Code-health check: dead-code delta on every PR

Adds a GitHub Action that runs [Knip](https://knip.dev) on the PR head and its base, then posts a sticky comment with the difference — which unused files / exports / types / dependencies this PR **introduces** vs. **removes**.

### DIFF
--- a/.github/scripts/knip-delta.mjs
+++ b/.github/scripts/knip-delta.mjs
@@ -1,0 +1,192 @@
+#!/usr/bin/env node
+// Diff two Knip JSON reports (base vs head) into a PR "code-health" delta: which
+// dead-code issues this PR introduces vs. removes, across the categories Knip
+// reports (unused files, exports, dependencies, …).
+//
+// The comparison is against the PR's base, so the bar ratchets forward on its
+// own as main improves — no committed baseline to maintain.
+//
+// Usage: node knip-delta.mjs <base.json> <head.json> <comment-out.md>
+// Always exits 0 (informational). Emits `introduced` / `fixed` as step outputs
+// so a future version can gate on `introduced > 0`.
+import { appendFileSync, readFileSync, writeFileSync } from 'node:fs';
+
+const [, , basePath, headPath, outPath] = process.argv;
+const MARKER = '<!-- knip-health -->';
+const CAP = 10; // max offenders listed per category before "…and N more"
+
+// Display order + labels for the Knip issue categories we report, using Knip's
+// own names. `files` is the file itself being unused; the rest are arrays of
+// symbols/names on a file. (unresolved imports and duplicate exports are
+// intentionally omitted for now — add here to include them.)
+const LABELS = {
+    files: 'Unused files',
+    exports: 'Unused exports',
+    types: 'Unused exported types',
+    nsExports: 'Exports in used namespace',
+    nsTypes: 'Exported types in used namespace',
+    enumMembers: 'Unused exported enum members',
+    namespaceMembers: 'Unused exported namespace members',
+    dependencies: 'Unused dependencies',
+    devDependencies: 'Unused devDependencies',
+    optionalPeerDependencies: 'Referenced optional peerDependencies',
+    unlisted: 'Unlisted dependencies',
+    binaries: 'Unlisted binaries',
+};
+const RAW_CATS = Object.keys(LABELS);
+
+// One-line explainer per category, shown when its section is expanded.
+const EXPLAIN = {
+    files: 'File imported nowhere — delete (or import) it.',
+    exports:
+        'Exported symbol with no references outside its own file — un-export it, or delete it if unused entirely',
+    types: 'Exported type with no references outside its own file — un-export it, or delete it if unused entirely',
+    nsExports:
+        'Symbol only reachable through a namespace import; nothing references it',
+    nsTypes:
+        'Type only reachable through a namespace import; nothing references it',
+    enumMembers: 'An enum member referenced nowhere',
+    namespaceMembers: 'A namespace member referenced nowhere',
+    dependencies: 'In package.json but never imported',
+    devDependencies: 'In package.json devDependencies but never used',
+    optionalPeerDependencies:
+        'Optional peer dependency in package.json that is actually referenced',
+    unlisted: 'Imported but missing from package.json',
+    binaries:
+        'A binary used in scripts but not provided by any listed dependency',
+};
+
+function load(path) {
+    try {
+        const data = JSON.parse(readFileSync(path, 'utf8'));
+        if (!data || !Array.isArray(data.issues)) return null;
+        return data;
+    } catch {
+        return null;
+    }
+}
+
+// Map every issue to a stable key so we can diff the *sets* (not just counts):
+// a PR that fixes one issue and adds another nets zero but still introduced one.
+function keyset(report) {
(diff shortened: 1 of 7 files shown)


package-lock.json
```diff
@@ -6151,12 +6851,12 @@
             "license": "MIT"
         },
         "node_modules/@types/node": {
-            "version": "20.17.30",
-            "resolved": "https://registry.npmjs.org/@types/node/-/node-20.17.30.tgz",
-            "integrity": "sha512-7zf4YyHA+jvBNfVrk2Gtvs6x7E8V+YDW05bNfG2XkWDJfYRXrTiP/DsB2zSYTaHX0bGIujTBQdMVAhb+j7mwpg==",
+            "version": "26.0.1",
```
jshearer: This lockfile silently bumps the **hoisted** `@types/node` from 20.17.30 to 26.0.1 — a six-major jump in the typings `tsc` resolves for `src/` — in a PR that only adds knip.

`package.json` doesn't pin `@types/node` and tsconfig has no `types` allowlist, so after merge everyone's next `npm ci` typechecks against Node 26 typings (e.g. `NodeJS.Timeout`/global signature changes), potentially breaking `npm run typecheck` on main for an unrelated-looking PR. Suggest pinning/deduping `@types/node` back to 20.x in the lockfile before merging

.github/scripts/knip-delta.mjs
```diff
@@ -0,0 +1,228 @@
...
+    'devDependencies',
+    'optionalPeerDependencies',
+    'unlisted',
+    'binaries',
+    'exports',
+    'types',
+    'nsExports',
+    'nsTypes',
+    'enumMembers',
+    'classMembers',
```
jshearer: `RAW_CATS` doesn't match knip v6's JSON reporter: `'classMembers'` is not an issue type knip 6.23.0 emits (removed in v6 — dead config here, plus its `LABELS`/`EXPLAIN` entries), while `'namespaceMembers'` **is** emitted and reported by default (only `nsExports`/`nsTypes` are default-excluded) but is missing from this list.

Net effect: a PR that introduces a dead exported namespace member reports "⚪ No change to the dead-code surface", and the "Unused class members" category can never populate. Verified against `knip@6.23.0` `dist/constants.js` (ISSUE_TYPES) and `dist/reporters/json.js`.

.github/scripts/knip-delta.mjs
```diff
@@ -0,0 +1,228 @@
...
+            for (const entry of issue[cat] ?? []) {
+                const name =
+                    typeof entry === 'string'
+                        ? entry
+                        : (entry?.name ?? JSON.stringify(entry));
+                const dcat = displayCat(cat, entry || {});
+                const key =
+                    dcat === 'files'
+                        ? `files ${name}`
+                        : `${dcat} ${file} ${name}`;
```
jshearer: The key drops the entry's `namespace` (parent enum/class): knip's JSON entries for `enumMembers` are `{namespace: parentSymbol, name, ...}` (see `dist/reporters/util/util.js` `convert()`), but the key here is built from `name` alone.

So two dead members named e.g. `Unknown` in `enum A` and `enum B` of the same file both map to `enumMembers src/x.ts Unknown`. Concretely: base has dead `A.Unknown`; a PR adds dead `B.Unknown` in the same file → `baseKeys.has(k)` is true and the new issue is silently netted away (`introduced=0`). Including `entry.namespace` in the key (and in `fmt()`, where the listing is currently ambiguous between parents) fixes it.

.github/scripts/knip-classify.mjs
```diff
@@ -0,0 +1,151 @@
...
+    visit(sf);
+
+    // Couldn't anchor to a declaration (missing pos, re-export, generated name):
+    // fall back to "more than one reference exists" — imperfect, but no worse
+    // than the old count heuristic.
+    if (!decl) return refs.length >= 2;
+
+    const stmt = topLevelStatement(decl, sf);
+    const lo = stmt.getStart(sf);
+    const hi = stmt.getEnd();
```
jshearer: Equating "the declaration's own subtree" with one top-level statement misclassifies in both directions (both reproduced empirically with fixture files run through this script):

1. **Function overloads**: a genuinely dead `export function foo` with overload signatures is 3 top-level statements. The `foo` identifiers in the sibling overload statements fall outside `[lo, hi)`, so `refs.some(...)` returns true → `privateUse=true`, and the delta labels it "unnecessary export — just drop `export`" — advice that then trips the unused-var linter. (Same failure shape for interface/function declaration merging.)
2. **Multi-declarator statements**: `export const a = 1, b = a + 1;` — the reference to `a` inside `b`'s initializer is *within* the same statement, so it's excluded → `a` is labeled fully dead, though deleting it breaks `b`.

Anchoring to the specific declarator/declaration node set for the symbol (rather than the enclosing top-level statement) would fix both.

.github/workflows/health.yml
```diff
@@ -0,0 +1,120 @@
...
+                  node "$RUNNER_TEMP/classify.mjs" "$RUNNER_TEMP/head.json"
+
+            - name: Knip on base (measured with the head's config)
+              run: |
+                  git checkout --force --quiet ${{ github.event.pull_request.base.sha }}
+                  # Pin the HEAD's Knip config for the base run too, so a PR that
+                  # only changes config (like the one that introduces this check)
+                  # doesn't manufacture a phantom delta. The ruler is constant;
+                  # only source differences register. Falls back to base's config.
+                  git checkout --quiet ${{ github.event.pull_request.head.sha }} -- knip.json 2>/dev/null || true
```
jshearer, line 58: The "constant ruler" pin covers only `knip.json`, but knip's effective config also includes `tsconfig.json` and `package.json` (plugin auto-detection, `package.json#knip`), which are never pinned — so a PR changing tsconfig `paths` or dependency lists can still manufacture a phantom delta today. And if config ever moves to `knip.ts`/`package.json#knip`, this checkout fails silently via `|| true` and base/head measure with different configs.

Not a today-bug (the check is informational and the fallback is documented), but the failure mode is exactly the one this pin was written to prevent — a red `+N` on a config-only PR that teaches people to distrust the check. Worth either pinning the wider config surface or at least echoing a warning into the report when the pin path diverges.

jshearer (changes requested):
Claude found at least one "structural" problem with the code that it claims will cause false-positive signals.

This (dead code detection) is a genuinely hard problem to get right, and isn't even wholly solvable in a dynamic language like Typescript. I'm not sure that the juice is worth the squeeze here...

Also, what's up with that typing version bump? Maybe a no-op, but seems relevant.

.github/scripts/knip-delta.mjs
GregorShear: Thanks - this is fixed

GregorShear:
I'm feeling pretty tolerant of an imperfect tool here - it's only meant to give visibility on existing unused code, and will help us identify code orphaned by future changes. 

false negatives (orphaned code missed by the check) will just leave dead code in the project, and false positives can be ignored - if we get so many false positives that the check is useless, we can just turn it off.

GregorShear:
Discussed this on a call w/ Joseph - we agreed that presenting the bare Knip output is a stronger first pass, and we can refine/reclassify the categories later, if needed

jshearer (approved):

PR #2020, opened 2026-07-01
Remove unnecessary exports

GregorShear:


GregorShear:
What's special about this diff? 

It represents _most_ of the unnecessarily-exported-but-still-used-privately symbols - i.e. we can just drop the `export` keyword without running into the linter.

jshearer (approved):

PR #2022, opened 2026-07-02
Add orphaned-props category to the Code Health check

GregorShear:
## Summary

Adds a new `orphaned-props` category to the existing Code Health PR check. A prop can be declared (and even read internally) on a component while no caller anywhere in `src/` ever passes it, so it's always its default/undefined value in practice — Knip doesn't catch this since it only tracks module-level exports, not JSX call sites.

`orphaned-props.mjs` resolves each component's Props type syntactically (inline object types, named interfaces/type aliases, `extends`, intersections, and `Pick`/`Omit`/`Partial`/`Required`/`Readonly`) and cross-references it against every JSX invocation in the tree, matching usage sites back to declarations via this repo's absolute `src/`-rooted imports — no full type-checker needed.

A component is skipped (never reported) rather than guessed at when:
- its Props type doesn't resolve cleanly (generics, mapped/conditional types, a type re-exported through a barrel, etc.)
- any call site spreads attributes (`<Foo {...rest} />`) — the spread could supply any prop
- it has zero call sites — that's dead-code territory Knip already covers

`children` is treated as satisfied by non-empty JSX content, not just an explicit attribute.

Results merge into the existing Knip JSON report as a new `orphanedProps` category, so `knip-delta.mjs`'s base-vs-head ratchet and PR-comment rendering pick it up for free — no changes to the diffing/commenting logic itself beyond registering the category's label and explainer text.

## Heads up

Run against the current tree, this finds ~98 pre-existing orphaned props. Nothing here is blocking (the whole pipeline is informational/ratchet-forward — only *new* issues a PR introduces show as `+`), but the first comment on this PR will show that as a sizeable "carried" baseline.

## Test plan

- [x] Ran the script directly against the full `src/` tree; spot-checked several findings by hand against the source (`ConnectionStatusBadge.compact`, `CardWrapper.sx`, `MessageWithEmphasis.emphasisContent`, `OnboardGuard.children`) — all genuine, no false positives found in the sample
- [x] Fixed a double-counting bug where components declared as `function Foo() {} export default Foo;` were registered under two keys pointing at the same entry
- [x] Verified `knip-delta.mjs` renders the new category correctly with a synthetic base/head diff
- [ ] Let the Code Health workflow run for real on this PR and review the actual comment

### DIFF
--- a/.github/scripts/knip-classify.mjs
+++ b/.github/scripts/knip-classify.mjs
@@ -1,0 +1,151 @@
+#!/usr/bin/env node
+// Tag each unused exported symbol in a Knip JSON report with whether it's still
+// referenced privately within its own file — beyond its own declaration. This
+// separates an *unnecessary export* (used privately — just drop `export`, a
+// clean diff) from a *dead exported symbol* (used nowhere — the export was only
+// hiding it from the unused-var linter, so delete it).
+//
+// We parse each file with the TypeScript compiler and ask whether any reference
+// to the symbol falls outside its own declaration's subtree. A name-count regex
+// can't answer that: it reads a recursive call or a self-referencing type as a
+// "use" and mislabels genuinely dead code as merely unnecessary.
+//
+// Must run while the report's ref is checked out, since it reads source files.
+// Rewrites the report in place. Usage: node knip-classify.mjs <report.json>
+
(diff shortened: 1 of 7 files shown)


GregorShear:
Punting on this for now

PR #2025, opened 2026-07-07
sops-utils: Overlay-aware endpoint config parsing

willdonnelly:
## Issues

https://github.com/estuary/flow/issues/2985

## Changes

Extends parseEncryptedEndpointConfig to recognize the sops.overlay property and merge it over the plaintext fields extracted from an encrypted config.

This is necessary so that configs with overlays will preserve the overlay values when viewing/editing in the UI, without this we'd see default or stale values showing up instead and saving would clobber any overlay values.

This does _not_ implement the whole "we can now modify nonsensitive fields without re-entering credentials" feature which overlays make possible, that seemed complicated and like something best left to the UI team. This commit is solely about making sure that nothing breaks when we start having configs with overlay properties.

### Manually tested

The capture and materialization edit UI for tasks modified to have overlay properties now shows the values from the overlays in the appropriate form fields after this change.

### Automated tests

Added unit tests of the overlay merge operation.

### DIFF
--- a/src/gql-types/graphql.ts
+++ b/src/gql-types/graphql.ts
@@ -654,11 +654,11 @@ export type DataPlane = {
   /** Name of this data-plane under the catalog namespace. */
   name: Scalars['String']['output'];
   /**
-   * Configured private link endpoints for this data-plane. Replacing this
-   * list (via `updateDataPlanePrivateLinks`) triggers reconvergence by the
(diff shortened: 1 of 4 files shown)


kiahna-tucker (dismissed):
LGTM. Your branch is currently behind `main`. Once it's updated and the build passes, you are clear to merge.

kiahna-tucker (approved):

PR #2026, opened 2026-07-10
Rename capabilities -> legacy_capabilities

GregorShear:
Migrates the frontend off the deprecated `Capability` GraphQL type in favor of `LegacyCapability`, ahead of the backend's new orthogonal capability model (`CapabilityBit`/`CapabilityBundle`).

- `CreateInviteLink` mutation now sends `capabilityLegacy: LegacyCapability!`, and `GenerateInvitation.tsx` uses the `LegacyCapability` type throughout
- `AuthRolesQuery` drops the deprecated `minCapability` filter (`prefixes(by: {})` — the empty filter has the same behavior)
- `codegen.ts` enables `inputValueDeprecation` during schema introspection so deprecated input fields/arguments appear in the generated schema snapshot
- Regenerated `src/gql-types/*` against the latest backend schema

No frontend usage of the new `CapabilityBit`/`CapabilityBundle` enums yet — they only exist in generated types. Typecheck and `codegen:local` pass clean.


src/stores/Entities/hooks.ts
```diff
@@ -137,7 +137,7 @@ const useEntitiesStore_populateState = () => {
 //  That should load in around 1 second for 4g
 const authRolesQuery = gql<AuthRolesQueryResponse, PaginationVariables>`
     query AuthRolesQuery($after: String) {
-        prefixes(by: { minCapability: read }, first: 7500, after: $after) {
+        prefixes(by: {}, first: 7500, after: $after) {
```
GregorShear, line 137: `by.minCapability` was just required by the original API, `read` was the lowest capability possible, so this was essentially just fetching all reachable prefixes (i'm not sure why the param was required in the first place. omitting the param entirely would have been more semantically clear IMO).

Anyway, `minCapability` is optional and deprecated now, replaced by withCapabilities.

GregorShear:
will come back to this later, closing for now

PR #2027, opened 2026-07-10
Replace hand-written GraphQL types with codegen types

GregorShear:
Replaces hand-written GraphQL response/variable types with codegen-generated ones. Codegen was already scanning these documents (`documents: 'src/**/*.{ts,tsx}'` picks up urql `gql` tags) — the hand-written generics on `gql<Response, Variables>` were *overriding* the generated types. Switching to the client-preset `graphql()` function lets the generated `TypedDocumentNode` flow through urql, and documents are now validated against the live schema at codegen time.

## Migrated (7 documents)

- `stores/Entities/hooks.ts` — AuthRolesQuery (deletes `AuthRolesQueryResponse`/`AuthRolesNode` from `src/types/gql.ts`)
- `api/gql/storageMappings.ts` — 4 mutations/queries (deletes `StorageMappingsQueryResponse`, `CreateStorageMappingResult`, `UpdateStorageMappingResult`; JSON payload shapes stay as explicit local interfaces cast at the boundary)
- `api/gql/liveSpecs.ts` — LiveSpecsQuery (deletes `LiveSpecsQueryResponse`)
- `api/gql/dataPlanes.ts` — DataPlanes (node type now derived from the generated `DataPlanesQuery`)

## Drift the hand-written types were hiding

- `CreateStorageMappingResult` claimed a `created` field the mutation never selects
- `PageInfo.endCursor` and `StorageHealthItem.error` typed non-null; both nullable in the schema
- `DataPlane.cloudProvider` is an enum that also allows `LOCAL`, which the local `CloudProvider` type doesn't model — now narrowed by one explicit cast in `toDataPlaneNode`
- `AuthRolesQueryResponse.userCapability` was plain `string`; it's now the schema enum

## Supporting changes

- Custom scalars that serialize as strings (`Prefix`, `DateTime`, `Id`, `Name`, `UUID`, …) now map to `string` instead of `any`; `JSON`/`JSONObject` intentionally stay `any`
- `useAllPages` accepts nullable `endCursor`/`after` (matching the schema's `PageInfo`)
- Enables `inputValueDeprecation` during introspection

## Alerts-related types left alone

The alerts cluster (9 raw `gql` documents across `api/alerts.ts`, `AlertHistory`, `AlertingOverview`, `AlertsAreActiveBadge`) is deferred until the in-flight alerts UI overhaul lands, to avoid conflicts.

### DIFF
--- a/codegen.ts
+++ b/codegen.ts
@@ -1,10 +1,18 @@
 import { loadEnvFile } from 'node:process';
 import type { CodegenConfig } from '@graphql-codegen/cli';
+import type { Types } from '@graphql-codegen/plugin-helpers';
+import type { IntrospectionOptions } from 'graphql';
 
 loadEnvFile(process.env.LOCAL ? '.env.development.local' : '.env');
 
 const config: CodegenConfig = {
-    schema: process.env.VITE_GQL_URL,
+    schema: [
+        {
+            [process.env.VITE_GQL_URL as string]: {
+                inputValueDeprecation: true,
+            } satisfies Partial<IntrospectionOptions> as Types.UrlSchemaOptions,
+        },
+    ],
     documents: 'src/**/*.{ts,tsx}',
     generates: {
         './src/gql-types/schema.graphql': {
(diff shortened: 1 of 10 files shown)


jshearer (approved):
LGTM

PR #2031, opened 2026-07-17
Migrate storage mappings table to GraphQL

GregorShear:
Migrate storage mappings table on the settings page to use the gql list query, and add pagination


src/components/tables/StorageMappings/index.tsx
```diff
@@ -1,82 +1,117 @@
...
-        sortDirection,
-        setSortDirection,
-        columnToSort,
-        setColumnToSort,
-    } = useTableState('sm', 'catalog_prefix');
+    const { currentPage, cursor, onPageChange } = useCursorPagination();
 
-    const selectedTenant = useTenantStore((state) => state.selectedTenant);
+    const { storageMappings, fetching, error, pageInfo, pageSize } =
+        usePaginatedStorageMappings(cursor);
```
jshearer, line 32: The pagination state isn't tied to the selected tenant. `Settings` keeps this table mounted when `TenantSelector` changes `selectedTenant`, so if I'm on page 2+ for tenant A and select tenant B, this sends A's `after` cursor with B's `underPrefix`. Best case we skip B's first page; worst case the query errors or returns zero edges. The latter strands the table because the footer below is gated on `storageMappings.length > 0`, so there isn't even a Previous button to recover.

Can we reset to page 0 when the tenant changes (or key/remount the paginated table by tenant)? I'd also mirror the Access Links / Refresh Tokens empty-page recovery using `goToPage(currentPage - 1)` after a completed zero-row fetch.

src/components/tables/StorageMappings/index.tsx
```diff
@@ -1,82 +1,117 @@
...
-        rowsPerPage,
-        setRowsPerPage,
-        searchQuery,
-        setSearchQuery,
-        sortDirection,
-        setSortDirection,
-        columnToSort,
-        setColumnToSort,
-    } = useTableState('sm', 'catalog_prefix');
+    const { currentPage, cursor, onPageChange } = useCursorPagination();
```
jshearer, line 29: `useCursorPagination` has a history bug that this makes reachable: after moving backwards, forward navigation always appends instead of truncating/replacing the abandoned tail. Reproduced sequence: `1 → 2 → 3 → 2 → 3 → 4 → 5 → 4`; the final "page 4" uses the page-3 cursor. It needs enough rows and some backtracking, so low likelihood, but the label and data diverge once it happens.

Worth fixing the shared hook's forward branch to trim history to the target page before adding `nextCursor`, and pinning it with this sequence.

src/components/tables/StorageMappings.tsx
```diff
@@ -1,82 +1,117 @@
...
-        sortDirection,
-        setSortDirection,
-        columnToSort,
-        setColumnToSort,
-    } = useTableState('sm', 'catalog_prefix');
+    const { currentPage, cursor, onPageChange } = useCursorPagination();
 
-    const selectedTenant = useTenantStore((state) => state.selectedTenant);
+    const { storageMappings, fetching, error, pageInfo, pageSize } =
+        usePaginatedStorageMappings(cursor);
```
GregorShear: i've backed out the tenant filtering on this PR - it'll be added back in with the scoped tokens we've been talking about

src/components/tables/StorageMappings/index.tsx
GregorShear: thanks- this should be fixed now

SeanWhelan (dismissed):
Approving. 2 cosmetic things, not blocking:
- `storageMappingsTable.filterLabel` and `.message2` are orphaned in
  `AdminPage.ts` now that the old table is gone
- `dataPlane.replace('ops/dp/', '')` duplicates `DATA_PLANE_PREFIX` from
  `src/settings/dataPlanes.ts`

SeanWhelan (approved):

PR #2036, opened 2026-07-30
Fix data preview showing acknowledgement documents for runtime v2 collections

jshearer:
## What

For collections written by runtime v2, the data preview selects an acknowledgement document as its first row, labeled with an empty key, instead of collection data. 

## Root cause

The data preview UI has filtered acknowledgements since it was first built (#298), by testing `_meta.ack`. Runtime V1 builds each collection's `ack_template_json` as `{"_meta":{"uuid":...,"ack":true}}`, and the v1 Go runtime appends that template with the UUID substituted in.

Runtime v2 never reads `ack_template_json`. Its acknowledgements come from `publisher::intents::build_transaction_intents` and are shaped `{"_meta":{"uuid":...},"is_ack":true,"hints":[...]}`, so `_meta.ack` is undefined and the filter passes them through.

An acknowledgement carries no collection key, so `ListView` computes an empty row label when it maps the collection's key pointers over one, and the auto-select effect then selects that row.

## Fix

Identify acknowledgements by the `ACK_TXN` flag that Gazette packs into the low 10 bits of the message UUID's clock-sequence field. Both runtime versions set it, and unlike a body field it is not something a user's schema can collide with.

-  Also filters runtime v1 ops acknowledgements, whose `ack` field sits at the document root (`go/runtime/ops_publisher.go`) where `_meta.ack` never matched it.
-  Keeps documents whose own schema defines `is_ack`, which a body-field filter would wrongly hide.
-  `loadDocuments` is the only journal reader in the UI, so this covers the data preview and the ops logs table.

## Automated tests

-   15 unit tests in `dataPlane-utils.test.ts` covering v1 collection and ops acknowledgements, v2 acknowledgements (with hints, without hints, and carrying a backfill marker), `CONTINUE_TXN` and `OUTSIDE_TXN` documents, a document defining its own `is_ack`, acknowledgements and documents with clock-sequence bits set, and uninterpretable UUIDs.

### DIFF
--- a/src/hooks/journals/shared.ts
+++ b/src/hooks/journals/shared.ts
@@ -18,6 +18,36 @@ import { CustomEvents } from 'src/services/types';
 import { INCREMENT } from 'src/utils/dataPlane-utils';
 import { journalStatusIsError } from 'src/utils/misc-utils';
 
+// Gazette packs message flags into the low 10 bits of a v1 UUID's clock-sequence
+// field, which is the fourth hyphen-delimited group. ACK_TXN marks a document as
+// the acknowledgement of a committed transaction: runtime bookkeeping rather than
+// collection data, so it should not be shown to users.
+//
+// The flag is what we test because the acknowledgement body differs by runtime
+// version. Runtime v1 writes the collection's ack template, `{"_meta":{"uuid":...,
+// "ack":true}}`, while runtime v2 writes `{"_meta":{"uuid":...},"is_ack":true,...}`.
+// Only the UUID is common to both, and `is_ack` sits in the user's own namespace
+// where a real document could legitimately carry it.
+//
+// See `build` and `parse` in crates/proto-gazette/src/uuid.rs of estuary/flow.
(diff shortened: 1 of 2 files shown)


src/hooks/journals/shared.ts
```diff
@@ -15,7 +15,7 @@ import { parseJournalDocuments } from 'data-plane-gateway';
 
 import { logRocketConsole, logRocketEvent } from 'src/services/shared';
 import { CustomEvents } from 'src/services/types';
-import { INCREMENT } from 'src/utils/dataPlane-utils';
+import { hasAckTxnFlag, INCREMENT } from 'src/utils/dataPlane-utils';
```
GregorShear: nit: i'd make a case for just defining these right in journals/shared.ts - they're not used anywhere else

jshearer: done!

GregorShear (approved):

PR #2037, opened 2026-07-31
Fix clipped storage connection errors

GregorShear:
Wraps long storage connection errors inside the authorization dialog and keeps Retry visible.

bug:
Screenshot 2026-07-31 at 11 20 02 AM

fixed:
Screenshot 2026-07-31 at 11 23 37 AM


kiahna-tucker (approved):
Approved without manual review (given the small scope of the change and supplementary images).

**ASIDE:** I am curious how accessible our semantic colors are when used for type. While I know that we are not dominantly concerned about accessibility as a company right now -- outside of a few, core areas -- it is worth keeping an eye on, reviewing/addressing it when there is time.

PR #2038, opened 2026-07-31
Render storage mapping instructions from Markdown

GregorShear:
Replaces bespoke provider instruction components with Markdown files and a shared renderer.


src/components/admin/Settings/StorageMappings/Dialog/ConnectionTest/ConnectionInstructions.tsx
```diff
@@ -1,9 +1,71 @@
...
-import { AwsInstructions } from 'src/components/admin/Settings/StorageMappings/Dialog/ConnectionTest/instructions/AwsInstructions';
-import { AzureInstructions } from 'src/components/admin/Settings/StorageMappings/Dialog/ConnectionTest/instructions/AzureInstructions';
-import { GcpInstructions } from 'src/components/admin/Settings/StorageMappings/Dialog/ConnectionTest/instructions/GcpInstructions';
+import { useMemo } from 'react';
+
+import {
+    useAwsArnsForBucket,
+    useBucketPolicy,
+} from 'src/components/admin/Settings/StorageMappings/Dialog/ConnectionTest/awsHooks';
+import markdown from 'src/components/admin/Settings/StorageMappings/Dialog/ConnectionTest/instructions/aws.md?raw';
```
jshearer: `markdown` here is `aws.md`, and it's the only markdown file this module imports: `Azure` (line 52) and `Gcp` (line 67) pass that same string to `MarkdownInstructions`, so all three providers render the AWS instructions. `azure.md` and `gcp.md` have no importer anywhere in `src/`.

- GCP renders "create the S3 bucket", a blank **Region** row, an empty policy block, and `aws s3api put-bucket-policy --bucket <gcp-bucket> --policy ''`.
- Azure passes neither `bucket` nor `region`, so it renders the AWS page with those blank too.
- Nothing catches it locally: typecheck and lint pass, and `interpolate`'s `variables[key] ?? ''` blanks the unsupplied placeholders instead of failing.

Fix: import all three (`awsMd`/`azureMd`/`gcpMd`) and pass the matching one from each component.

src/components/admin/Settings/StorageMappings/Dialog/shared/MarkdownInstructions.tsx
```diff
@@ -0,0 +1,138 @@
...
+}: MarkdownInstructionsProps) {
+    const theme = useTheme();
+
+    const content = useMemo(
+        () => interpolate(markdown, variables),
+        [markdown, variables]
+    );
+    const markdownOptions = useMemo(
+        () => ({
+            overrides: {
```
jshearer, line 47: The `{{...}}` values interpolated here come from stored storage-mapping config with no markdown escaping, and `markdown-to-jsx` parses raw HTML by default, so a value containing a backtick can break out of its code span and render HTML in another admin's dialog; disabling raw-HTML parsing closes that off.

```suggestion
            disableParsingRawHTML: true,
            overrides: {
```

jshearer (dismissed):
LGTM % one thing Claude found. Not worried about it because it doesn't permit xss or even cross-tenant injections afaiui

SeanWhelan (approved):

PR #2045, opened 2026-08-05
onboarding: data-plane picker in the signup form

SeanWhelan:
## Issues

estuary/sre#29 (Phase 2 — tracked in a roadmap doc, not an estuary/ui issue)

### sre#29

- Adds a `DataPlaneSelector` to the signup form (`BetaOnboard.tsx`), between the organization name field and the survey. It's backed by the existing unauthenticated `publicDataPlanes` query, which this PR consumes for the first time — the authenticated `dataPlanes` query returns nothing for a brand-new signup, since there are no grants yet.
- Auto-selects `ops/dp/public/aws-us-east-1-c1` when it's available, falling back to the first plane otherwise, so submitting without touching the picker still records an explicit, valid choice.
- Fails safe: if the plane list can't load, the picker renders nothing and the claim just omits `requestedDataPlane` rather than sending it as an explicit null, so the backend falls back to its own default instead of the signup form breaking.
- Extends the `betaOnboard` directive's `generateUserClaim` to forward `requestedDataPlane` only when it's set, so backends that don't know the field yet aren't affected.
- Adds a `usePublicDataPlanes()` hook and `PublicDataPlaneNode` type, following the existing `useDataPlanes()`/`DataPlaneNode` pattern already in the codebase.

Options show the region and the full catalog name together, and group by cloud provider. The catalog name is deliberate rather than noise: nothing stops two public planes sharing a region (already true for GCP), so the name is what makes the choice unambiguous.

This depends on estuary/flow#3318 merging **and deploying** first. The agent rejects unknown claim fields, so if this ships before that one is live, every signup fails with `invalidClaims`. Merging flow isn't sufficient on its own — an old agent still running will reject the claim.

### Manually tested

I ran a full signup through the real onboarding form against a local flow stack. The picker listed the seeded public planes grouped by cloud provider, auto-selected the platform default, and a completed registration produced the correct `storage_mappings` row with the picked plane recorded as the tenant's default.

I repeated this against a synthetic AWS-shaped plane with `COLOCATED_TRIAL_BUCKETS` enabled on the backend. The resulting storage mapping showed the derived S3 bucket, confirming the picker's choice flows through correctly end to end.

I also checked what happens against a real, older backend that predates this feature: the query fails with a clear GraphQL error, the picker just doesn't render, and the rest of the form stays fully usable. No crash.

Finally, with the companion PR's migration applied, a data plane marked `closed` disappears from the picker's options — the list and the backend agree on what's selectable.

### Automated tests

`tsc --noEmit` and `npm run lint` both pass clean. I didn't add new unit tests since this is UI wiring around an already-tested backend contract; flagging that for reviewer input rather than assuming it's fine.

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

picker-collapsed
picker-open


src/directives/Onboard/DataPlaneSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+import {
+    Autocomplete,
+    FormControl,
+    FormLabel,
+    inputBaseClasses,
+    TextField,
+} from '@mui/material';
+
+import { useIntl } from 'react-intl';
```
GregorShear: we're slowly deleting `react-intl` - user-facing text can just live in the components

src/lang/en-US/Authentication.ts
```diff
@@ -106,6 +106,8 @@ export const Authentication: Record<string, string> = {
     'tenant.input.label': `Organization Name`,
     'tenant.input.placeholder': `acmeCo`,
     'tenant.errorMessage.empty': `You must provide an organization name before continuing.`,
+    'tenant.dataPlane.label': `Data Plane`,
+    'tenant.dataPlane.helper': `Where your data is processed and stored. Pick the region closest to your data sources.`,
```
GregorShear: these can move inline in the component

src/directives/Onboard/Store/hooks.ts
```diff
@@ -17,6 +17,20 @@ export const useOnboardingStore_setRequestedTenant = () => {
     >(OnboardingStoreNames.GENERAL, (state) => state.setRequestedTenant);
 };
 
+export const useOnboardingStore_requestedDataPlane = () => {
+    return useLocalZustandStore<
+        OnboardingState,
+        OnboardingState['requestedDataPlane']
+    >(OnboardingStoreNames.GENERAL, (state) => state.requestedDataPlane);
+};
```
adrian-estuary, line 25: @SeanWhelan - I think the plan is to migrate away from Zustand. I would recommend just using react's `useState` to manage the selected data plane in the onboarding form. 

cc: @GregorShear - Correct me if I'm wrong here. 🙏

src/lang/en-US/Authentication.ts
```diff
@@ -106,6 +106,8 @@ export const Authentication: Record<string, string> = {
     'tenant.input.label': `Organization Name`,
     'tenant.input.placeholder': `acmeCo`,
     'tenant.errorMessage.empty': `You must provide an organization name before continuing.`,
+    'tenant.dataPlane.label': `Data Plane`,
+    'tenant.dataPlane.helper': `Where your data is processed. Pick the region closest to your data sources.`,
```
adrian-estuary, line 110: We're also migrating away from react-intl. So better to just put the strings in the component. 

cc: @GregorShear

adrian-estuary (changes requested):
Overall looks good. Just a couple of points about not using Zustand and react-intl.

src/directives/Onboard/DataPlaneSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    const currentOption = options.find((option) => option.name === selected);
+
+    return (
+        <FormControl>
+            <FormLabel id="requestedDataPlane" sx={{ mb: 1, fontSize: 20 }}>
+                {intl.formatMessage({ id: 'tenant.dataPlane.label' })}
+            </FormLabel>
+
+            <Autocomplete
```
adrian-estuary, line 83: Should add `disableClearable` to hide the `x` button to clear the selection. Seems like we don't want this to be unset. Also, the useEffect above will just trigger, setting it back to the default anyway.

src/directives/Onboard/DataPlaneSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+const INPUT_SX = {
+    maxWidth: 424,
+    [`& .${inputBaseClasses.root}`]: { borderRadius: 3 },
+};
+
+// The region and full catalog name are both shown: multiple public planes can
+// share a region, so the name is what makes the choice unambiguous.
+const optionLabel = (option: PublicDataPlaneNode) =>
+    `${option.region} (${option.name})`;
```
adrian-estuary, line 34: Is it required to show the full plane name to the user? Seems a bit noisy. Perhaps just the region and suffix would do?

So `us-east-1 - c1` (or something like that) versus `us-east-1 (ops/dp/public/aws-us-east-1-c1)`.

src/directives/Onboard/DataPlaneSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+            setSelected(preferred.name);
+        }
+    }, [options, selected, setSelected]);
+
+    // Fail safe: if the plane list can't be fetched, render nothing. The
+    // claim simply omits requestedDataPlane and the backend applies its
+    // own default instead of blocking signup on this field.
+    if (error || (!loading && options.length === 0)) {
+        return null;
+    }
```
adrian-estuary, line 73: Should failing API calls log somewhere? What do we use for frontend logging?

src/directives/Onboard/Store/hooks.ts
GregorShear: Generally yes, Zustand is overkill in a lot of places and should be removed. It's still appropriate when it shares state for some complex flow (setting up a new capture for example?) or a deep subtree. Not the case here.

(and in some future PR we could remove zustand entirely from Onboarding)

src/directives/Onboard/DataPlaneSelector.tsx
GregorShear: Oh good call 

```tsx
postHog.capture('Entity:Action', {
    someData,
});
```

src/directives/Onboard/DataPlaneSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+                        })}
+                        sx={INPUT_SX}
+                    />
+                )}
+            />
+        </FormControl>
+    );
+}
+
+export default DataPlaneSelector;
```
GregorShear, line 121: Pretty minor but would love help establishing a new pattern where we favor named exports 
```suggestion
```

src/directives/Onboard/DataPlaneSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+    maxWidth: 424,
+    [`& .${inputBaseClasses.root}`]: { borderRadius: 3 },
+};
+
+// The region and full catalog name are both shown: multiple public planes can
+// share a region, so the name is what makes the choice unambiguous.
+const optionLabel = (option: PublicDataPlaneNode) =>
+    `${option.region} (${option.name})`;
+
+function DataPlaneSelector() {
```
GregorShear, line 36:
```suggestion
export function DataPlaneSelector() {
```

SeanWhelan:
Thanks both, all addressed in a1cde3e and 90c99d6:

- Zustand: removed the store additions. The selected plane is now `useState` in `BetaOnboard`, and `DataPlaneSelector` is a controlled component (value/onChange).
- react-intl: strings are inline now and the two `tenant.dataPlane.*` keys are gone.
- disableClearable: added. Agreed, a plane is always preselected, so clearing it didn't make sense.
- Label: it's region + cluster now, e.g. `us-east-1 c1`, under the provider group heading. That still tells apart planes sharing a region, which was the reason for the full name. If a name doesn't parse, it falls back to the full name. The full name is still what gets submitted.
- Logging: a failed `publicDataPlanes` query now fires an `Onboard:DataPlanes` PostHog event with the error message. The picker still hides itself and signup carries on.
- Named exports: done.

I also trimmed the code comments down to the non-obvious ones in line with the feedback on the Bindings PR.

adrian-estuary (approved):
LGTM! Thanks for addressing the feedback! 🙏

PR #2048, opened 2026-08-07
Update tabs and card surfaces, align page content

SeanWhelan:
## Changes

This branch changes the page frame, the tabs, the card surfaces and the alignment of page content. It adds no new files and no new components. Every change is to code that exists already.

This branch is the base of a stack. The branch `sean/task-details-redesign` is above it.

### Surfaces

-   `CardWrapper` now shows a `palette.divider` border of 1 px. Before, it showed a shadow. The shadow had three layers. One layer moved up, and one layer moved down. Thus the card had light from two directions.
-   The shadow also did no work. In light mode, a white card is already different from the page behind it. In dark mode, a black shadow at 10 % on a dark page is not visible.
-   Light mode now has three surfaces. Dark mode had three surfaces before this change: canvas `grey[900]`, panel `grey[800]`, and cards with 5 % white. Light mode had two: canvas `grey[100]`, panel white, and cards white. The panel used the same colour as the cards. Only the border made them different.
-   Light mode now uses `grey[200]`, then `grey[100]`, then white.
-   **Dark mode does not change.** We measured the three colours again after the change. They are the same.
-   This change removes `defaultBoxShadow` and `opaqueLightModeBorder`. `CardWrapper` was the only user of each one.

### Alignment

-   `PageContainer` no longer sets its padding from the sidebar state. Before, it used `px: navigationOpen ? 1 : { xs: 1, md: 5 }`. This gave 8 px when the sidebar was open, and 40 px when the sidebar was closed.
-   The header panel above it always uses `px: 2`. The two panels have the same corner radius, thus they look like one card. When the user closed the sidebar, the page content moved 32 px to the right of the page title.
-   The alignment was correct with the sidebar open only by chance. On the details page, 8 px of panel padding plus 8 px of page margin is equal to the 16 px of the header. Pages without that margin were 8 px to the left.
-   `PageContainer` now always uses 16 px.
-   The list pages `Captures`, `Collections` and `Materializations` put their content in a `Toolbar` without `disableGutters`. This gave 24 px. They also put it in a `Box` with 16 px. Thus the button and the table started 30 px to the right of the page title. All other `Toolbar` components in this codebase use `disableGutters`.
-   The four Admin tabs had twelve horizontal margins and paddings. These were `m: 2` in `StandAloneTableTitle`, `m: 2` in `AccessGrants`, `ml: 2` in `tables/AccessGrants`, `margin: '0 16px'` in `tables/PrefixAlerts`, and `p: 2` or `mx: 2` in Settings, Api, Billing and the two RefreshToken sections. Headings, text, buttons and tables each started at a different position.
-   All of these are now vertical only. No content moves up or down.

**Note:** Three items keep their horizontal position on purpose. These are the centred empty states in tables, the second button of a button pair, and all content inside a `CardWrapper`. Content in a card aligns to the card.

### Tabs

-   The selected tab now has a filled background. Before, it had a line below it. That line needs a rule below the tab row. That rule was one of three rules at the top of a details page.
-   `scrollButtons="auto"` kept 40 px for buttons that it did not show. MUI disables the arrows when the row cannot move, but MUI keeps their space. The left arrow is disabled when the row is fully to the left. This is the usual state. Thus the Admin tab rows started to the right of the page edge, and no row was too long for its space.
-   Disabled arrows now have `display: none`.

**Note:** MUI keeps the space for the arrows to stop the row from moving when an arrow becomes enabled. This change accepts that risk. A row that is long enough to move will move 40 px the first time the user scrolls it. No tab row in this application is too long at desktop width.

-   The alerts badge is now 16 px. Before, it was 20 px with `top: -4`. Thus it started 6 px above the 34 px tab, and MUI cut the top of it. MUI sets `overflow: hidden` on the tab, the scroller and the root. We cannot remove these. The overflow of the scroller is necessary for `variant="scrollable"`.
-   The badge is now red, not orange. Red is the colour of the alerts panel.
-   The badge text colour is set to white in the code. White gives a contrast ratio of 4.92:1 on this red. Text of 10 px needs 4.5:1. But the `contrastThreshold` of the theme is 5. Thus `getContrastText` refuses white and gives dark text at 3.94:1.

### Other changes

-   `DataByHourGraph` now changes size with its container. Before, it listened only for `window.resize`. The sidebar changes the size of the container but not the size of the window. Thus the canvas kept its first width and became larger than its card.
-   A `ResizeObserver` replaces the window listener. We removed the listener. The `resize()` function of echarts does not stop when the size is the same, and each call causes two synchronous layout operations. Two listeners thus did this work two times for each change of the window size.
-   Toggle button text is no longer in capital letters. MUI puts `ToggleButton` text into capital letters. The theme stops this for `MuiButton` and `MuiTab`, but not for `MuiToggleButton`. Thus `Data`, `Docs`, `Private` and `Public` showed as DATA, DOCS, PRIVATE and PUBLIC. The language files always had the correct text.
-   The fix is in `OutlinedToggleButton`, not in the theme. This component includes all seven toggle buttons that show text. It does not include the `ToggleButton` components in the data preview, which show icons only.

### Manually tested

We measured all positions with `getBoundingClientRect()`. We did not estimate them by eye.

-   Content left edge against the page title, with the sidebar open and closed. Details page: 32 px to 0. `/captures`: −8 px to 0. `/admin/billing`: −8 px to 0.
-   All four Admin tabs. Each heading, text, button, rule and table starts at the page edge. Three items keep a different position on purpose. See the Note above.
-   Light mode and dark mode on `/welcome`, `/captures`, `/collections`, `/materializations`, the four Admin tabs, and a capture details page.
-   The three dark-mode surface colours. They did not change.
-   Tab row after the arrow change. The row starts at the page edge. There is 8 px between tabs. Each tab has 8 px of padding. The first tab is at the edge.
-   Alerts badge. It is fully inside the tab, with 2 px above it and 11 px to the right of it.
-   Chart size. We changed the observed element from 718 px to 380 px, then back to 718 px. The canvas followed both changes.
-   Menus, popovers and the ECharts tooltip. These use `paperBackground`. The tooltip is now `#F7F9FC` on the white chart card. Before, it was white on white.

**Note:** To test the chart size, you must cause a paint operation between the steps. The browser automation tool does not paint during script evaluation. A test without a paint operation gives an incorrect failure.

### Automated tests

CI runs seven gates. We ran all seven on this branch alone. This branch is the base of a stack, thus we tested it separately from the branch above it.

-   `licenses` — pass
-   `typecheck` — pass
-   `lint` — pass
-   `format` (Prettier) — pass
-   `test` (286 vitest tests) — pass
-   `codegen:check` — pass
-   `build` — pass

This branch adds no unit tests. It changes no logic. It changes only layout values and colour values.

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

Welcome (before):
1a-welcome-before

Welcome (after):
1b-welcome-after

Admin settings (before):
2a-admin-settings-before

Admin settings (after):
2b-admin-settings-after

Task details (before):
4a-task-details-before

Task details (after):
4b-task-details-after

Sidebar collapsed (before):
5a-sidebar-collapsed-before

Sidebar collapsed (after):
5b-sidebar-collapsed-after

Tabs (before):
6a-tab-row-before

Tabs (after):
6b-tab-row-after


src/components/shared/PageContainer.tsx
```diff
@@ -154,10 +155,15 @@ function PageContainer({ children, hideBackground }: Props) {
 
             <Paper
                 sx={{
-                    px: navigationOpen ? 1 : { xs: 1, md: 5 },
-                    transition: (t) =>
-                        `padding ${t.transitions.duration.shortest}ms`,
-                    py: 2,
```
GregorShear, line 160: I'd like to keep this extra padding when the nav is collapsed

src/components/shared/CardWrapper.tsx
```diff
@@ -41,16 +42,13 @@ function CardWrapper({
...
                 borderRadius: 3,
                 rowGap: 2,
                 minWidth: disableMinWidth ? undefined : 'min-content',
                 background: opaqueLightMode
                     ? opaqueLightModeBackground[theme.palette.mode]
                     : semiTransparentBackground[theme.palette.mode],
-                border: opaqueLightMode
-                    ? opaqueLightModeBorder[theme.palette.mode]
-                    : undefined,
+                border: `1px solid ${theme.palette.divider}`,
```
GregorShear: agree the border is better than a shadow, but imo dark mode looks better with no border (or shadow)

Image

Image

src/components/shared/buttons/OutlinedToggleButton.tsx
```diff
@@ -18,6 +18,12 @@ const OutlinedToggleButton = styled(ToggleButton)(({
     const colorKey = color === 'standard' || !color ? 'primary' : color;
 
     return {
+        // MUI uppercases ToggleButton labels. The theme opts MuiButton and
+        // MuiTab out of that but not MuiToggleButton, so every label here
+        // arrived shouting. Owned at this level rather than in the theme
+        // because a theme override would also catch the icon-only raw
+        // ToggleButtons in the data preview.
+        'textTransform': 'none',
```
GregorShear: let's put this in the theme instead

```
  MuiToggleButton: {
      styleOverrides: {
          root: { textTransform: 'none' },
      },
  },
```

GregorShear (approved):

PR #2049, opened 2026-08-08
details: Show a materialization's data backlog

GregorShear:
Adds a **Data Backlog** row to the Overview details card for materializations, showing how many bytes are left to read across all bindings.

The value comes from `bytesBehind` in `catalog_stats.flow_document` (`taskStats.materialize`, keyed by source collection). Only the V2 runtime reports it, so the row is gated on the `estuary.dev/flag/enable-runtime-v2` shard label. It's a gauge, not a counter, so only the newest stats row with binding data is used — heartbeat rows open the hour without any. A binding that omits the field is caught up.

Also included: scoped the vitest `include` to `src/` so sibling worktrees aren't picked up, and dropped the ESLint plugin from the Vite dev-server checker (it was hitting EMFILE on macOS).

### DIFF
--- a/src/api/stats.ts
+++ b/src/api/stats.ts
@@ -2,6 +2,7 @@ import type { PostgrestResponse } from '@supabase/postgrest-js';
 import type { DataByHourRange } from 'src/components/graphs/types';
 import type {
     CatalogStats,
+    CatalogStats_Backlog,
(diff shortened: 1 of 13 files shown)


dyaffe (approved):
Lgtm

PR #2050, opened 2026-08-08
details: Show how far behind a materialization is in time

GregorShear:
## Changes

Adds a **Time Behind** row to the materialization details card, beside Data Backlog: how far behind the task is in source-publication time — the gap between the newest document published to a source collection and the newest one the task has processed.

- Shown for V2-runtime materializations only. Renders `—` when the lag can't be computed (e.g. a source collection whose producing capture isn't on V2, so no `lastPublishedAt` is recorded).
- Task-level figure is the worst binding's lag, since a task is only as caught up as its furthest-behind binding.
- Durations render to two units ("1 day, 6 hours") via a reusable `toHumanDuration` luxon helper.

### DIFF
--- a/src/api/stats.ts
+++ b/src/api/stats.ts
@@ -5,6 +5,7 @@ import type {
     CatalogStats_Backlog,
     CatalogStats_Dashboard,
     CatalogStats_Details,
+    CatalogStats_LastPublished,
     Entity,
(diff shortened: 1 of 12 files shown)


dyaffe (approved):
LGTM

PR #1332, opened 2024-10-29
PROD: CSP report only settings

travjenkins:
Cleaning up handing always

## Issues

https://github.com/estuary/security/issues/196

### 196

- Cleaning up some old stuff
- Adding a new section to test out changes and report back to cspr.io what is happening

PR #1316, opened 2024-10-16
Update insecure CSP header settings

travjenkins:
## Issues

https://github.com/estuary/security/issues/196

### 196

-  Start listing out specific domains we need to reach

### Manually tested

-   scenarios you manually tested

### Automated tests

-   unit testing covered

#### Playwright tests ran locally

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

_If applicable - please include some screenshots of the new UI_

PR #2052, opened 2026-08-12
Add ad.doubleclick.net to CSP connect-src

skord:
## Issues

No issue filed. Reported by marketing as a problem in the GTM configuration for
approved events: the tag targeting `https://ad.doubleclick.net` cannot fire
because our CSP blocks it.

## Changes

-   Added `https://ad.doubleclick.net` to the `connect-src` directive in
    `public/nginx.conf`.
-   Placed inline immediately after the existing `https://stats.g.doubleclick.net`
    so the doubleclick hosts stay grouped, and kept on a single line per the
    warning in this file that newlines in the CSP block previously broke prod.

Context on why this was missing: `ad.doubleclick.net` has never been present in
this file, so this is not a re-add of something previously removed. The other
doubleclick hosts were enumerated in Oct 2024 (#1316, #1332) from observed CSP
violation reports, during a window when `connect-src` still carried a bare
`https:` wildcard. That wildcard was dropped in #1346, which is when the
enumeration became restrictive. This host simply never appeared in a report at
the time.

Existing doubleclick coverage for reference:

| Directive | Hosts |
| --- | --- |
| `connect-src` | `stats.g.doubleclick.net`, `ad.doubleclick.net` (new) |
| `frame-src` | `td.doubleclick.net` |
| `script-src` | `*.g.doubleclick.net` |

Note that `img-src` is `'self' data: https:`, so image-based pixels on any
doubleclick host were already permitted regardless of this change.

### Manually tested

-   Not verifiable in local dev. This header is served by nginx, and Vite does
    not apply `public/nginx.conf` under `npm start`, so the change is inert
    locally. Needs verification in a deployed environment by confirming the GTM
    `connect-src` violation for `ad.doubleclick.net` no longer appears in the
    browser console, and that marketing sees the approved events landing.

### Automated tests

-   None. There is no test coverage over the nginx CSP header.

Not run: this change only affects the nginx-served CSP header, which is not
applied under the Vite dev server these tests run against. Every suite would
behave identically with and without this commit.

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

_Not applicable, no UI change._


Alex-Bair (approved):
LGTM

PR #2053, opened 2026-08-13
Give a task's bindings their own section, polished

SeanWhelan:
## Changes

- Added a dedicated Bindings section to the task Overview tab (capture and materialization), with a sortable/searchable/paginated table replacing the old chip-list-in-a-rail treatment.
- Row density, hover/selection state, empty state, loading skeletons, and dark-mode card border brought in line with the rest of the design system.
- Status column now distinguishes enabled / disabled / no-data-in-range at a glance. A task-wide connector error shows once as a header chip instead of repainting every enabled row's status pill red.
- Volume and "last data" columns right-align their digits in a fixed-width column so they line up regardless of unit-suffix length (e.g. "5 minutes ago" vs "1 hour ago").
- Search field no longer relies on the browser's native `type="search"` clear glyph — swapped for an `Xmark` icon consistent with the rest of the app.
- CSV export goes through the existing `CsvDownload`/`ExportButton` pattern (`generateFileName`, `tableExportSeparator`) instead of a hand-rolled Blob/anchor download.
- Collection name cell drops its own link styling (color/underline/tooltip) when the whole row is already the click target — new opt-in `plain` prop on `EntityNameDetailsLink`/`LinkWrapper` — while staying a real, keyboard-focusable anchor.
- `CardWrapperProps.sx` is now typed as `SxProps<Theme>` (was bare `SxProps`), removing an `as SxProps` cast at its one call site.

### Manually tested

- Capture and materialization variants, light and dark mode, in Storybook (empty state, loading state, task-erroring state, 1200-binding pagination with mixed enabled/disabled/no-data rows).
- Search filter + clear button, status chip filters, column sorting, CSV export.
- Verified against a real local backend (VM stack) — login, no console errors, TypeScript watch clean.

### Automated tests

- `npm run typecheck`, `npm run lint`, `npm run format` all clean.
- `npx vitest run` — 364/364 passing, including existing coverage for `Bindings/shared.ts`.

- [ ] Admin
- [ ] Captures
- [ ] Collections
- [ ] HomePage
- [ ] Login
- [ ] Materialization

## Screenshots

Before ("Writes To" in side card):
image

After (bindings card):
image

### DIFF
--- a/src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
+++ b/src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
@@ -138,19 +138,3 @@ export const VolumesLoading: Story = {
         />
     ),
 };
-
-/**
- * The connector's shards are reporting errors. A "Connector error" chip
- * appears in the card header — the failure is task-wide, not a property of
- * any one row — while each row's own status pill keeps showing whether that
- * particular binding actually moved data in the range.
- */
-export const TaskErroring: Story = {
-    render: () => (
-        <BindingsHarness
-            bindings={buildCaptureRows(CAPTURE_STREAMS)}
-            entityType="capture"
-            taskHasError
-        />
-    ),
-};
--- a/src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
+++ b/src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
@@ -138,3 +138,19 @@ export const VolumesLoading: Story = {
         />
     ),
 };
+
+/**
+ * The connector's shards are reporting errors. A "Connector error" chip
+ * appears in the card header — the failure is task-wide, not a property of
+ * any one row — while each row's own status pill keeps showing whether that
+ * particular binding actually moved data in the range.
+ */
+export const TaskErroring: Story = {
+    render: () => (
+        <BindingsHarness
+            bindings={buildCaptureRows(CAPTURE_STREAMS)}
+            entityType="capture"
+            taskHasError
+        />
+    ),
+};
--- a/src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
+++ b/src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
@@ -9,7 +9,9 @@ import {
     buildCaptureRows,
     buildLargeTaskStreams,
     buildMaterializationRows,
+    buildMaterializationRowsWithBacklog,
     CAPTURE_STREAMS,
+    MIXED_STATUS_STREAMS,
 } from 'src/components/shared/Entity/Details/Overview/Bindings/Bindings.fixtures';
 import enUSMessages from 'src/lang/en-US';
 
@@ -56,6 +58,24 @@ export const Materialization: Story = {
     ),
 };
 
+/**
+ * The two lag columns, materialization-only. Every third row cycles through a
+ * caught-up binding (a literal 0 reading, rendered as "Caught up" rather than
+ * a blank — every stream here has a reading, so the "no reading yet" dash
+ * never shows), a moderately-behind one (about an hour of bytes, a few hours
+ * of source time), and a heavily-behind one (about a day of bytes, a couple
+ * of days of source time) — see `buildMaterializationRowsWithBacklog` for
+ * exactly how each tier is derived.
+ */
+export const MaterializationBacklog: Story = {
+    render: () => (
+        <BindingsHarness
+            bindings={buildMaterializationRowsWithBacklog(CAPTURE_STREAMS)}
+            entityType="materialization"
+        />
+    ),
+};
+
 /** A task with a single binding still has to read as a table. */
 export const SingleBinding: Story = {
     render: () => (
@@ -92,6 +112,22 @@ export const NoneEnabled: Story = {
 /** A spec with no bindings at all. */
 export const Empty: Story = {
     render: () => <BindingsHarness bindings={[]} entityType="capture" />,
+};
+
+/**
+ * 12 bindings, so the default 10-per-page pagination actually kicks in —
+ * and every status the table can show (enabled, disabled, no data) plus the
+ * full range of bar lengths all land on that first page regardless, without
+ * paging through to see it. `job_openings` is disabled but still carries a
+ * real bar; `eeoc` is enabled with none.
+ */
+export const MixedStatusesOnePage: Story = {
+    render: () => (
+        <BindingsHarness
+            bindings={buildCaptureRows(MIXED_STATUS_STREAMS)}
+            entityType="capture"
+        />
+    ),
 };
 
 /**
--- a/src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
+++ b/src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
@@ -1,0 +1,140 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+
+import { IntlProvider } from 'react-intl';
+import { MemoryRouter } from 'react-router-dom';
+
+import { DataGrains } from 'src/components/graphs/types';
+import {
+    BindingsHarness,
+    buildCaptureRows,
+    buildLargeTaskStreams,
+    buildMaterializationRows,
+    CAPTURE_STREAMS,
+} from 'src/components/shared/Entity/Details/Overview/Bindings/Bindings.fixtures';
+import enUSMessages from 'src/lang/en-US';
+
+// ── Meta ─────────────────────────────────────────────────────────────
+
+const meta: Meta<typeof BindingsHarness> = {
+    title: 'Details/Bindings',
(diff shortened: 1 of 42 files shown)


src/components/shared/Entity/Details/Overview/Bindings/__tests__/shared.test.ts
GregorShear, line 1: Honestly I'm inclined to exclude tests from these PRs until someone wants to put some thought into a holistic approach to testing and QA. 

(Except maybe for regression tests to defend subtle bug fixes related to essential domain complexity...)

src/components/shared/Entity/Details/Overview/Bindings/BindingsCardHeader.tsx
```diff
@@ -0,0 +1,213 @@
...
+ */
+function BindingsCardHeader({
+    count,
+    entityType,
+    loading,
+    range,
+    rows,
+    totalBytes,
+}: Props) {
+    const intl = useIntl();
```
GregorShear, line 60: I probably have some notes on copy but please remove the react-intl stuff first so my comments don't land on outdated code

src/components/shared/Entity/Details/Overview/Bindings/BindingsCardHeader.tsx
```diff
@@ -0,0 +1,213 @@
...
+                            </IconButton>
+                        </span>
+                    </Tooltip>
+                </CsvDownload>
+            </Stack>
+        </Stack>
+    );
+}
+
+export default BindingsCardHeader;
```
GregorShear, line 213: same comment i left on another PR - favor named over default exports 
```suggestion
```

src/components/shared/Entity/Details/Overview/Bindings/LastDataCell.tsx
```diff
@@ -0,0 +1,97 @@
...
+    return (
+        <TableCell align="right" sx={sx}>
+            {/* Uncoloured, whatever the age. A fixed staleness threshold was
+                tried and does not survive the range picker: a capture stamps
+                this when it publishes, so on a 6h range no capture binding can
+                reach 24h and the warning is unreachable — while on a 30d range
+                it lights up permanently for reference tables that update
+                weekly by design, which is how a warning teaches people to stop
+                reading it. Sorting the column finds the quiet bindings without
+                asserting a threshold that has no stable meaning here. */}
```
GregorShear, line 48: i take it this is Claude commenting on some design element you removed - I find myself begging claude to stop describing paths not taken.

generally i'd like to keep code comments concise and meaningful, only describing things that are both important and non-obvious. This probably means zero comments about style or design choices...

src/components/shared/Entity/Details/Overview/Bindings/LastDataCell.tsx
```diff
@@ -0,0 +1,97 @@
...
+                        whiteSpace: 'nowrap',
+                    }}
+                >
+                    {/* Fixed-width and right-aligned for the same reason as
+                        the volume column's digit span: elapsed is always 1-2
+                        digits (minutes top out at 59, hours at 23, days at the
+                        30-day retention window), but "hour" vs "minutes" vs
+                        "days" suffixes differ enough in width that without a
+                        fixed box here, the digit itself drifts left and right
+                        between rows as the suffix changes length. */}
```
GregorShear, line 67: ???

src/components/shared/Entity/Details/Overview/Bindings/LastDataCell.tsx
```diff
@@ -0,0 +1,97 @@
...
+                            minWidth: 20,
+                            textAlign: 'right',
+                        }}
+                    >
+                        {elapsed.value}
+                    </Box>
+
+                    {/* Unit and "ago" read at the same weight as the figure
+                        they belong to, matching the volume and lag columns'
+                        own unit suffixes. */}
```
GregorShear, line 81: not important

src/components/shared/Entity/Details/Overview/Bindings/LastDataCell.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { getElapsed } from 'src/components/shared/Entity/Details/Overview/shared';
+
+interface Props {
+    // Null when this binding moved nothing in the selected range, which is a
+    // fact about the range rather than about the binding — so the cell stays
+    // blank instead of claiming "never".
+    lastPublishedAt: string | null;
+    loading: boolean;
+    // Lets the table nudge this last column's padding to align with the
+    // card's own edge without every other cell needing the same override.
```
GregorShear, line 17: `sx` props are common, no need to comment

src/components/shared/Entity/Details/Overview/Bindings/LastDataCell.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import { DateTime } from 'luxon';
+import { useIntl } from 'react-intl';
+
+import { getElapsed } from 'src/components/shared/Entity/Details/Overview/shared';
+
+interface Props {
+    // Null when this binding moved nothing in the selected range, which is a
+    // fact about the range rather than about the binding — so the cell stays
+    // blank instead of claiming "never".
```
GregorShear, line 13: IMO this is meaningful and non-obvious. i'd keep it

src/components/shared/Entity/Details/Overview/Bindings/Bindings.stories.tsx
GregorShear, line 1: curious if you were using storybook to actually develop this new UI without access to a good test data source? Or did claude build these stories as part of a test suite?

GregorShear:
@SeanWhelan kicking this back over to you. Please remove `react-intl` and scrutinize the code comments (i left my own meta-commentary on one of the files as an example. Happy to chat about any you're on the fence about.)

Once that's done I'll take another look and review the actual feature :) Looks awesome!

SeanWhelan:
Thank you @GregorShear! I pushed 0334681 with all of this is addressed and CI is green again (the Prettier failure on `BindingsTable` is fixed too).

- Tests - dropped both new test files. Four helpers in `shared.ts` were only reachable from them so they're module-private now rather than exported.
- react-intl - removed from all the new components. All 37 message keys this branch added are gone and `Details.ts` is back to matching main so the PR now adds none. Copy is all plain strings now.
- Named exports - converted throughout, 11 components plus the two new hooks. The only default left is `export default meta` in the stories files, which Storybook's CSF format requires.
- Comments - Cut the styling and design rationale and the paths not taken, what's left is domain facts that aren't recoverable from the code. Why volume reads `right` and not `out` for a materialization, why the spec drives the rows rather than `writes_to` etc. I kept the `lastPublishedAt` one you flagged
- Storybook - Those are deliberate, they turned out to be a genuinely great way to work on a component like this without standing up the full UI and hunting for a task with the right shape of data. If you want it gone, I can scrap it or exclude it from tracking.

SeanWhelan:
Pushed 19f4c0c with main merged in and two follow-ups:

- Binding terms - switched to `BINDING_TERMS` from #2090, so a materialization's section and counts read "destination tables" and match the backfill section. Captures still say "collections".
- Comments - last round I only went through files this PR created. I've now applied the same cut to comments the branch added to files it modified (`types/index.ts`, `api/stats.ts`, `entity.ts` and a few others).

src/hooks/details/useShardStatusNeedsAttention.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Whether a task's shard status is anything but running.
+ *
+ * Compared against the colour rather than `shardsHaveErrors` /
+ * `shardsHaveWarnings`: those only populate from the FAILED branch's
+ * inferred-schema check (see `getEverythingForDictionary` in
+ * `stores/ShardDetail/Store.ts`), so an IDLE, STANDBY or BACKFILL shard would
+ * leave both flags false and never surface here.
+ */
+export function useShardStatusNeedsAttention(
```
adrian-estuary: Does this need to be a new hook? Looks like it's only used in `ShardAwareSectionOrder.tsx` and it's not doing much in terms of logic. Might be simpler to just move this into the `ShardAwareSectionOrder` component.

src/components/shared/Entity/Details/Overview/Bindings/BindingsCardHeader.tsx
```diff
@@ -0,0 +1,152 @@
...
+                justifyContent: 'space-between',
+                width: '100%',
+            }}
+        >
+            <Stack direction="row" spacing={1} sx={{ alignItems: 'baseline' }}>
+                <Typography component="span" sx={cardHeaderSx_emphasized}>
+                    {heading}
+                </Typography>
+
+                <RangeChip range={range} />
```
adrian-estuary: So the bindings table can set a different range than the usage chart on the same page? Wondering if it would be better to hoist the range selector up in the page and have both share a single range. 🤔 

@GregorShear - thoughts on this?

adrian-estuary (commented):
A couple of comments for now. Still going through the rest of the PR.

PR #2055, opened 2026-08-17
Hide managed `collection-data` storage suffix

GregorShear:
Storage mapping mutations automatically append `collection-data/` to the storage bucket prefix. The API used to strip that addition so the user would see the same value they entered into the UI, but that stripping behavior is moving to the UI in this PR. 

The API will return the full prefix so that other clients (`flowctl`) may use it.


jgraettinger (approved):
LGTM

PR #2067, opened 2026-08-26
Update DataPlaneTable in admin view to use GQL instead of Supabase

### LINKED ISSUE #2065: Switch data plane table queries over to GQL
Serve the admin data planes table (`src/components/tables/DataPlanes/`) from the GQL `dataPlanes` query instead of PostgREST.

Keep the same UX, and add pagination.

adrian-estuary:
## Issues

https://github.com/estuary/ui/issues/2065

## Changes

- New GQL query hook to fetch data planes using cursor pagination.
  - ~Requires landing https://github.com/estuary/flow/pull/3403 first, as it utilizes the new `public` filter.~ Done!
  - ~Requires landing https://github.com/estuary/flow/pull/3437, as it utilizes the new `tenant` filter.~ - Done!  
- Updated the `DataPlanesTable` component to use the GQL hook.
- Updated the `ToggleDataPlaneScope` to simplify scope selection.
  - Addressed a new render loop issue when using the cursor pagination state.
- Deprecated the `EnityTable` component.
- Removed intl code from `DataPlanesTable`.
- Deleted the now unused supabase hook for fetching data planes.

## Test plan

- [x] `npm run format`
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (310 passed)
- [x] Manual: verify the admin Data Planes table (public + private toggle) renders and the detail dialog opens

## Screenshots

**Public data planes (default 1 local plane)**
Screenshot 2026-08-26 at 2 53 01 PM

**Private data planes (none in local)**
Screenshot 2026-08-26 at 2 53 31 PM

**Many data planes - paginated (seeded data)**
Screenshot 2026-08-26 at 3 45 56 PM

**Private data planes (only 1 page, no pagination controls show up)**
Screenshot 2026-08-26 at 4 29 25 PM

**Dialog**
Screenshot 2026-08-26 at 4 27 47 PM

**GQL query in network tab**
Screenshot 2026-08-26 at 2 57 45 PM


src/components/tables/DataPlanes/DataPlaneDialog.tsx
```diff
@@ -23,8 +22,14 @@ function DataPlaneDialog({ onClose, dataPlane }: DataPlaneDialogProps) {
...
-    const { dataPlaneName, scope } = generateDataPlaneOption(dataPlane);
-    const { ipv4, ipv6 } = parseCidrBlocks(dataPlane.cidr_blocks);
+    const { cloudProvider, scope } = dataPlane;
+    const { ipv4, ipv6 } = parseCidrBlocks(dataPlane.cidrBlocks);
+
+    // The region is parsed out of the data plane name rather than read off the
+    // `region` field so it stays in sync with the displayed name. They differ
+    // for the local data plane: `ops/dp/public/local-flow` parses to `flow`,
+    // while the field reports `local`.
+    const { region } = parseDataPlaneName(dataPlane.name, scope);
```
adrian-estuary: Not sure if this is just a local environment quirk and if this logic is really needed for production. Thoughts?

src/components/tables/DataPlanes/index.tsx
```diff
@@ -1,91 +1,192 @@
...
-                hideFilter
-                toolbar={<ToggleDataPlaneScope />}
-            />
-        </TableHydrator>
+        <Box data-public>
+            <Toolbar sx={toolbarStyle} disableGutters>
+                <ToggleDataPlaneScope
+                    scope={dataPlaneScope}
+                    onChange={handleDataPlaneScopeChange}
+                />
```
adrian-estuary, line 101: Updated the `ToggleDataPlaneScope` component to take in the scope and handle changes via a callback. There was some wonky interaction with it directly updating the context that caused a render loop.

src/components/tables/EntityTable/index.tsx
```diff
@@ -30,6 +30,9 @@ import { selectableTableStoreSelectors } from 'src/stores/Tables/Store';
 import { TableStatuses } from 'src/types';
 import { getPagination, getStartingPage } from 'src/utils/table-utils';
 
+/**
+ * @deprecated Use the `@mui/material` table components instead.
+ */
 function EntityTable({
```
adrian-estuary, line 36: As discussed in Slack, deprecating this component in favor of using basic MUI table components.

src/components/tables/DataPlanes/DataPlaneDialog.tsx
GregorShear: While you're in here would you remove this dependency? Copy and delete strings from AdminPage.ts (and wherever else) and paste them into the component body below.

src/api/dataPlanes.ts
```diff
@@ -96,4 +69,4 @@ const getDataPlanesForTable = (
 //     return data;
 // };
```
GregorShear: safe to delete this comment now

src/components/tables/DataPlanes/types.ts
```diff
@@ -1,18 +1,18 @@
...
 
 export interface RowProps {
-    row: BaseDataPlaneQuery;
+    row: DataPlaneNode;
     rowSx: any;
-    onRowClick: (row: BaseDataPlaneQuery) => void;
+    onRowClick: (row: DataPlaneNode) => void;
 }
 
 export interface DataPlaneDialogFieldProps {
```
GregorShear: This is partly a subjective style preference, but I'm inclined to move these types into the files that use them, at least when they only have one consumer (which these do).

GregorShear:
left a few low-conviction comments so far - i'll get to the stuff that actually matters with fresh eyes tomorrow.

src/components/tables/DataPlanes/DataPlaneDialog.tsx
adrian-estuary: Updated!

src/api/dataPlanes.ts
adrian-estuary: Done!

src/components/tables/DataPlanes/index.tsx
```diff
@@ -1,91 +1,192 @@
...
-        setRowsPerPage,
-        searchQuery,
-        setSearchQuery,
-        sortDirection,
-        setSortDirection,
-        columnToSort,
-        setColumnToSort,
-    } = useTableState(TablePrefixes.dataPlanes, 'data_plane_name', 'asc');
-
-    const selectedTenant = useTenantStore((state) => state.selectedTenant);
```
adrian-estuary, line 31: @GregorShear - How does the current dataPlanes GQL query handle tenant filtering? Can a user theoretically have access to multiple tenants and therefore get back data planes for multiple tenants in a single query? Wondering if the GQL query needs a tenant filter... 🤔

GregorShear: yep good question. users can and do have access to multiple tenants - we have a couple of data consultant customers who will set up tenants for other companies.

i've been wanting to solve this more broadly by adding a `prefix_scope` (or something) claim to the token, or maybe a header included with every request that limits authorization to a single branch in the user's grant-graph. IMO we (maybe you?) should do this as soon as brian finishes the capability-masked token stuff.

adrian-estuary: Gotcha. So for now, is this work blocked on the token work? Or should we proceed with a prefix filter on the GQL query filter. 

Adding it to the GQL filter seems inline with the current supabase implementation, but not sure it's worth it if we want to do it another way in the future. Thoughts?

GregorShear: let's forego the prefix filter on this query for now. it'll be no worse than the current UX (users can see data planes for all tenants they have access to)

GregorShear: for posterity: we chatted about this and i'm wrong, we do actually need the prefix filter, even if it will be superseded by the tenant header soon.

src/components/tables/DataPlanes/types.ts
adrian-estuary: @GregorShear - Do you mean moving the `...Props` types to their respective component files? If so, I agree and can def update this PR to do that.

GregorShear: yeah IIRC each of the exports in this file are only imported in one place, so they could just be defined there.

adrian-estuary: Done!

src/components/tables/DataPlanes/index.tsx
```diff
@@ -1,91 +1,195 @@
...
-        sortDirection,
-        setSortDirection,
-        columnToSort,
-        setColumnToSort,
-    } = useTableState(TablePrefixes.dataPlanes, 'data_plane_name', 'asc');
+    const { rowsPerPage, setRowsPerPage } = useTableState(
+        TablePrefixes.dataPlanes,
+        'name',
+        'asc'
+    );
```
GregorShear: we can go even further and delete this (does stuff like stores the rows-per-page selection in the url, like `admin/settings?dpt-r=25` IMO not necessary)

GregorShear: and just declare `const PAGE_SIZE = 10` to replace `rowsPerPage`

src/components/tables/DataPlanes/index.tsx
```diff
@@ -1,91 +1,195 @@
...
+                        noExistingDataContentIds={noExistingDataContentIds}
+                    />
+
+                    <TableFooter>
+                        <TableRow>
+                            <TablePagination
+                                count={-1}
+                                page={currentPage}
+                                rowsPerPage={rowsPerPage}
+                                onPageChange={handlePageChange}
```
GregorShear, line 140:
```suggestion
                               rowsPerPageOptions={[PAGE_SIZE]} // hides the rows-per-page selector (which i wish was just off by default - who actually ever touches that?
                               onPageChange={handlePageChange}
```

src/components/tables/DataPlanes/index.tsx
```diff
@@ -1,91 +1,195 @@
...
+            actions: {
+                previousButton: {
+                    disabled: previousButtonDisabled,
+                },
+                nextButton: {
+                    disabled: nextButtonDisabled,
+                },
+            },
+        };
+    }, [pageInfo?.hasNextPage, pageInfo?.hasPreviousPage]);
```
GregorShear: the `<TablePagination />` component isn't memoized, so stabilizing `slotProps` (and many of its siblings) does nothing. Could probably inline some of them in the component props
```suggestion
     const slotProps = {
        actions: {
            previousButton: {
                disabled: !pageInfo?.hasPreviousPage,
            },
            nextButton: {
                disabled: !pageInfo?.hasNextPage,
            },
        },
    };
```

src/components/tables/DataPlanes/DataPlaneDialog.tsx
GregorShear: I'm not following here - `region` is in fact parsed from the name in the gql resolver so i wouldn't expect this logic to change anything in the UI. 

I sometimes wonder if we could just get rid of the "local" region and call it GCP or whatever, would simplify a bunch of little annoying things.

adrian-estuary: @GregorShear - Basically the `dataPlane.region` field comes back as `local`, while currently with the parsing logic it shows up in the ui as `flow`. 

My question is if this difference is only a thing that shows up in the local environment, or if it could potentially change what region is displayed for production data planes. My gut feeling is this is just a local env issue and that we probably don't need this logic. But kept it the same, just in case. 

I'm ok with removing this logic and just relying on the `dataPlane.region` field if you're aligned.

src/components/tables/DataPlanes/index.tsx
adrian-estuary: So, just confirming, the requirement is to hide the page size selector? Can do, if we're aligned here.

GregorShear: yes please :)

src/components/tables/DataPlanes/index.tsx
adrian-estuary: I'd rather just add `memo` to the `TablePagination` component. Any reasons not to?

GregorShear: it's MUI's component - is there an issue the memo would fix?

src/components/tables/DataPlanes/DataPlaneDialog.tsx
adrian-estuary: Discussed on slack. Aligned it should be fine to remove the parsing logic and just rely on the `dataPlane.region` field. Updated!

src/components/tables/DataPlanes/index.tsx
adrian-estuary: Not really. Just was of the mindset that components should pretty much always be memoized. Was how we approached it in my last job. But not a big deal here since the pagination component is super simple and re-renders should be cheap.

adrian-estuary: Updated!

src/components/tables/DataPlanes/index.tsx
adrian-estuary: Done!

src/components/tables/DataPlanes/index.tsx
adrian-estuary: Done!

src/components/tables/DataPlanes/index.tsx
```diff
@@ -1,91 +1,170 @@
...
+    );
+
+    const labelDisplayedRows = useCallback(
+        ({ from }: { from: number }) => {
+            return dataPlanes.length > 0
+                ? `${from} – ${from + dataPlanes.length - 1}`
+                : '';
+        },
+        [dataPlanes.length]
+    );
```
GregorShear: No need to change this before merging but prefer to omit memoization (`useCallback` and friends) until there's some performance reason to use it

GregorShear (dismissed):

GregorShear (approved):

src/components/tables/DataPlanes/index.tsx
adrian-estuary: Updated!

PR #2073, opened 2026-08-28
Add a shared BandedDiv component

GregorShear:
## Changes

Adds `theme.radius` and `src/components/shared/BandedDiv.tsx`. Nothing consumes either yet; the service accounts UI (#1989) is the first caller and stacks on top of this PR.

### `theme.radius`

A named border-radius scale on the augmented theme (`sm` 4px, `md` 8px, `lg` 12px, `xl` 16px, `full`), so roundness tracks an element's size and prominence and surfaces of the same class round the same way. Values are px strings, which `sx` and `styled` use literally — a *number* in `borderRadius` is multiplied by `theme.shape.borderRadius` instead.

### `BandedDiv`

A rounded frame whose face sits over a colored band poking out one edge.

- The band is pulled under the face by one corner radius, so the face's rounded corners curve away onto band color rather than opening gaps to the page. The face paints above it (`zIndex: 1`, opaque), and the frame's `overflow: hidden` clips the band's outer corners, so the band needs no radius of its own.
- `side` picks the banded edge and defaults to `left`. `bandColor` is required, and accepts any color the browser can parse — named CSS colors included, normalized through a canvas `fillStyle` because MUI's color math rejects them.
- `label` fills a slot spanning the band's visible run. A single element owns its own layout; a fragment, string, or array is laid onto a default centered row (caption type, small gap) so a bare icon-and-text label works without a wrapper. Long labels wrap and thicken the band.
- Vertical bands rotate the label 270° as one unit and hand it the band's full run. Since a CSS transform never affects layout, `RotatedLabel` measures with a `ResizeObserver`: the label's natural height becomes the band's thickness, and the stretched spacer's length becomes the rotated label's width.
- With `onClick` set, the frame becomes a `ButtonBase` and the band rests desaturated, coming up to full color on hover. `desaturate` and `BAND_REST_SATURATION` are exported so an owner can hold other surfaces to the same rest color, and `BANDED_DIV_FACE_CLASS` lets an owner restyle the face from the frame's hover.
- The face defaults to an opaque padded surface with a border tinted toward the band color. `faceSx` merges onto the face; `sx` targets the frame.

### Manually tested

-   Rendered through the service account cards on #1989: bottom band on expiring keys, hover raising band and border together, band label wrapping at narrow card widths.

### Automated tests

-   None. Typecheck, ESLint, and Prettier pass.

### DIFF
--- a/src/components/shared/BandedDiv.tsx
+++ b/src/components/shared/BandedDiv.tsx
@@ -38,6 +38,30 @@ export function desaturate(color: string, amount: number) {
     });
 }
 
+// MUI's color math (decomposeColor, getContrastText) rejects named CSS colors
+// like `tomato`. Assigning a color to a canvas fillStyle makes the browser
+// parse and normalize it to hex or rgba(), which MUI accepts. The context is
+// reset to black first, so an unparseable color degrades to a black band
+// rather than silently reusing the previous call's color.
+let canvasContext: CanvasRenderingContext2D | null | undefined;
+
+function toDecomposableColor(color: string): string {
+    if (color.startsWith('#') || color.includes('(')) {
+        return color;
+    }
+
+    canvasContext ??= document.createElement('canvas').getContext('2d');
+
+    if (!canvasContext) {
+        return color;
+    }
+
+    canvasContext.fillStyle = '#000000';
+    canvasContext.fillStyle = color;
+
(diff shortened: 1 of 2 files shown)


GregorShear:
Unwinding the split — BandedDiv goes back into greg/access (#1989).

PR #2075, opened 2026-08-28
Rebuild AlertBox on BandedDiv

GregorShear:
## Changes

Rebuilds `AlertBox` on the `BandedDiv` from #2073. The severity color moves into the band; the copy sits on the face.

Stacked on #2073 (BandedDiv) and under #1989 (service accounts).

### The component

- The MUI `Alert` is gone, and with it the `alertClasses` overrides that were reshaping its icon, message, and action slots, plus `iconMapping` and `SHARED_STYLING`. The frame, the band tuck, and the severity-tinted face border all come from `BandedDiv`.
- The band is bare — its color carries the severity on its own. `side="left"`, `radius="sm"` (4px, matching the old alert), `bandColor={theme.palette[severity][mode]}` (unchanged from the old border color).
- `role="alert"` and the forwarded ref land on the alert's content, which is what `ValidationErrorSummary` scrolls into view.
- `onClose` renders its own close button now that MUI isn't supplying one.

### The API

`AlertBoxProps` is `severity`, `children`, `onClose`, `sx`, `title`, and it moved next to the component. Removed:

- **`short`** — 86 call sites set it, 82 of them the same way. A band reads the same whatever the message length, so there was nothing left for the compact/tall split to express.
- **`headerMessage`** and the four `alert.*` messages — the `h4` header they fed is gone. The one caller passes its copy as `title`.
- **`hideIcon`** — nothing to hide once the icon went.
- **`condensed` on `Error`** — it only ever picked the AlertBox presentation. Off 26 call sites.

`alertBackground` and `alertTextPrimary` came out of `Theme.tsx`, which had no other consumer for them.

`AlertBox` now has a named export alongside the default; the 77 existing importers are untouched.

### Manually tested

-   `/admin/billing` against the local stack, both payment-provider alerts: band `rgb(193,75,96)` at a 24px visible run, face border `1px solid rgb(193,75,96)`, `role="alert"` present.

### Automated tests

-   None. Typecheck, ESLint, and Prettier pass.

## Screenshots

_Band-only alert, dark mode — severity color at the left edge, copy on the face._

### DIFF
--- a/src/app/guards/GrantGuard/RedeemInviteLink.tsx
+++ b/src/app/guards/GrantGuard/RedeemInviteLink.tsx
@@ -67,7 +67,7 @@ export function RedeemInviteLink({ grantToken }: Props) {
                         })}
                     </Typography>
 
-                    <Error error={redeemError} condensed />
+                    <Error error={redeemError} />
 
                     <MessageWithLink messageID="tenant.grantDirective.error.message.help" />
                 </Stack>
--- a/src/app/guards/TenantGuard/SsoUserMessage.tsx
+++ b/src/app/guards/TenantGuard/SsoUserMessage.tsx
@@ -13,7 +13,6 @@ function SsoUserMessage() {
     return (
         <FullPageWrapper fullWidth>
             <AlertBox
-                short
                 severity="info"
                 title={intl.formatMessage({ id: 'tenant.usedSso.title' })}
             >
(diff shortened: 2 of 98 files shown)


GregorShear:
Unwinding this — AlertBox stays on MUI Alert, and BandedDiv goes back into greg/access.

PR #2083, opened 2026-09-08
Deprecate StandAloneTableTitle i18n props and remove DataPlaneScopeContext

GregorShear:
- Deprecate `StandAloneTableTitle` i18n props in continued effort to remove `react-intl`
- Remove `DataPlaneScopeContext` which isn't earning its keep given some recent gql-related simplifications


adrian-estuary (approved):
LGTM!

PR #2090, opened 2026-09-14
Only show enabled collection count when backfilling

adrian-estuary:
## Issues

The source backfill UI includes all bindings (even disabled ones) in the selection count, even though the disabled bindings will be filtered out server-side. The UI explains this with copy, but it would be clearer to instead update the count to not include disabled bindings. This issue was mentioned in the public slack channel.  
Screenshot 2026-09-14 at 12 30 23 AM

### Main change:
Updating the source backfill UI to only show the count of enabled bindings when choosing to backfill all, instead of the current behavior of showing ALL bindings (enabled + disabled). Updated the copy text as well to explicitly state disabled bindings will not be backfilled. See below screenshots.

**Original copy:**
```
Trigger a backfill of all collections from the source when published. The UI will mark all collections to be 
backfilled but the server will filter out those that cannot be backfilled (e.g. disabled collections).
```

**New copy:**
```
Trigger a backfill of all enabled collections from the source when published. Disabled collections will not be 
backfilled.
```

### Other changes:
Added VSCode settings to enable auto-formatting on save via Prettier. Also added a recommended extensions list.

### Manually tested

- Locally tested source with all 81 collections enabled. Shows 81/81 as expected.
- Locally tested source with only 12/81 collections enabled. Shows 12/81 as expected.
- Verified enabling/disabling collection while the backfill all option is currently selected updates the count.
- Verified count matches the bindings table footer.

### Automated tests

-  Ran existing unit tests.

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materialization

## Screenshots

**All bindings enabled. Shows "81/81" in chip and "all backfilled" in table footer.**
Screenshot 2026-09-14 at 12 22 45 AM

**Subset of bindings enabled. Shows "12/81" in chip and shows the count in the footer.**
Screenshot 2026-09-14 at 12 24 55 AM


src/settings/entity.ts
```diff
@@ -105,6 +109,8 @@ export const ENTITY_SETTINGS: { [k in Entity]: EntitySetting } = {
         Icon: CloudDownload,
         background: semiTransparentBackground_purple,
         backgroundNesting: semiTransparentBackground_purple_nesting,
+        bindingTermSingular: 'collection',
+        bindingTermPlural: 'collections',
         bindingTermId: 'terms.collections.plural',
         pluralId: 'terms.destinations.plural',
```
adrian-estuary, line 115: Even though this is the "materialization" section, I set the `bindingTerm*` fields to "collection(s)" because the existing `bindingTermId` was set that way. Not sure if that is intentional or copy-pasta, so kept it as-is.

src/settings/types.ts
```diff
@@ -25,6 +25,8 @@ export interface EntitySetting {
     >;
     background: { light: string; dark: string };
     backgroundNesting: { light: string | undefined; dark: string | undefined };
+    bindingTermSingular: string;
+    bindingTermPlural: string;
```
GregorShear: IMO this EntitySetting is overgeneralized and I'm reluctant to add to it, but this is the only way to add this text without a big refactor. Let's keep an eye on it and look for ways to chip away.

src/lang/en-US/Workflows.ts
```diff
@@ -105,18 +105,13 @@ export const Workflows: Record<string, string> = {
     'workflows.collectionSelector.manualBackfill.notSupported.message.docPath': `${CommonMessages['support.email']}`,
 
     'workflows.collectionSelector.manualBackfill.message.capture': `Trigger a backfill of this collection from the source when published.`,
-    'workflows.collectionSelector.manualBackfill.message.capture.allBindings': `Trigger a backfill of all collections from the source when published. The UI will mark all collections to be backfilled but the server will filter out those that cannot be backfilled (e.g. disabled collections).`,
+    'workflows.collectionSelector.manualBackfill.message.capture.allBindings': `Trigger a backfill of all enabled collections from the source when published. Disabled collections will not be backfilled.`,
```
GregorShear: Move this whole group to the BackfillDescription component please

`workflows.collectionSelector.manualBackfill.message.${entityType}${suffix}`
`workflows.collectionSelector.manualBackfill.message.materialization${suffix}.warning`

src/settings/entity.ts
```diff
@@ -63,6 +65,8 @@ export const ENTITY_SETTINGS: { [k in Entity]: EntitySetting } = {
         Icon: DatabaseScript,
         background: semiTransparentBackground_blue,
         backgroundNesting: semiTransparentBackground_blue_nesting,
+        bindingTermSingular: 'collection',
+        bindingTermPlural: 'collections',
```
GregorShear: AFAIK `collection.bindingTerm` is never used - evidence that `EntitySetting` is the wrong abstraction (not asking you to change anything, just calling it out, hope to unwind this someday)

src/settings/entity.ts
```diff
@@ -18,6 +18,8 @@ export const ENTITY_SETTINGS: { [k in Entity]: EntitySetting } = {
         Icon: CloudUpload,
         background: semiTransparentBackground_teal,
         backgroundNesting: semiTransparentBackground_teal_nesting,
+        bindingTermSingular: 'binding',
+        bindingTermPlural: 'bindings',
```
GregorShear: "captures" backfill "collections"

src/settings/entity.ts
GregorShear: "materializations" backfill "destination tables" (or something? i'm kinda making this up)
https://estuaryworkspace.slack.com/archives/C0AN2JHAGPJ/p1789743052382199

src/hooks/bindings/useBackfillCountMessage.ts
```diff
@@ -1,97 +1,50 @@
...
+    const evolvedCollectionsCount =
+        useBinding_enabledEvolvedCollections_count();
+    const backfillCount = useBinding_enabledBackfilledBindings_count();
     const bindingsTotal = useBinding_collections_count();
 
     // We shouldn't have any overlap (duplicate counting) with manual backfilling because backfill button is disabled
-    //  when evolved. So only time we could get duplicate counting is from the backfill all button
+    // when evolved. So only time we could get duplicate counting is from the backfill all button
     const calculatedCount =
         backfillCount === bindingsTotal
```
GregorShear, line 37: i haven't verified this, but agent is telling me the this comparison now "mixes enabled-only and total counts, which causes double-counting"

src/settings/entity.ts
adrian-estuary: Updated!

src/settings/types.ts
adrian-estuary: Moved this out to a couple of `Record<Entity, string>` maps in the component. So ended up not needing to add these.

src/lang/en-US/Workflows.ts
adrian-estuary: Done!

src/settings/entity.ts
adrian-estuary: Ended up moving these to const records (as mentioned above). So these fields are no longer needed.

src/settings/entity.ts
adrian-estuary: Done!

src/hooks/bindings/useBackfillCountMessage.ts
adrian-estuary: Yea, I think there could be a bug here. 🤔  Digging into it more. Thanks for the catch!

adrian-estuary: Ok, pushed a fix. 

Simplified the count logic by introducing a new helper+hook to go through the resource configs and count the bindings marked for backfill and the evolved bindings (de-duping by the binding uuid to prevent double counting). 

Since the new hook handles that, I was able to revert the previous two hooks I added to get the individual counts. So overall simpler, I think.

playwright-tests/tests/captures.spec.ts
```diff
@@ -17,7 +17,7 @@ import {
 } from '../helpers/utils';
 import { expect, Page, test } from '@playwright/test';
 
-const defaultBackfillMessage = 'no bindings marked for backfill';
+const defaultBackfillMessage = 'no collections marked for backfill';
```
adrian-estuary: binding teminology changed, so updated the test to match

adrian-estuary:
@GregorShear - Due to the changes in the binding terminology, some places in the UI are a bit inconsistent. For example, on the edit capture page, the "backfill pill" and the text above the table uses the term "collections", but the table itself used "bindings". Should I go ahead an update the table to also use collection?

Screenshot 2026-09-19 at 12 39 17 PM

src/components/collection/Selector/List/CollectionSelectorFooter.tsx
```diff
@@ -9,22 +9,16 @@ import {
...
 
-import { useIntl } from 'react-intl';
-
 import { DEFAULT_ROW_HEIGHT } from 'src/components/collection/Selector/List/shared';
 import { defaultOutlineColor } from 'src/context/Theme';
 import { useBackfillCountMessage } from 'src/hooks/bindings/useBackfillCountMessage';
 import { useBinding_enabledBindings_count } from 'src/stores/Binding/hooks';
 
 function CollectionSelectorFooter({
-    columnCount,
```
adrian-estuary, line 20: This prop isn't used anywhere, so removing it.

playwright-tests/tests/captures.spec.ts
```diff
@@ -17,7 +17,9 @@ import {
...
+// "Backfill All" marks every binding, so the two counts in the chip must match.
+const allBackfilledMessage = /(\d+) of \1 collections? will be backfilled/;
```
adrian-estuary, line 22: binding teminology changed, so updated the test to match

playwright-tests/tests/captures.spec.ts
```diff
@@ -58,12 +60,12 @@ test.describe.serial('Captures:', () => {
...
 
-        // Checking defailt
-        await expect(backfillCounter).toHaveScreenshot();
+        // Checking default
+        await expect(backfillCounter).toContainText(defaultBackfillMessage);
 
         // Ensuring backfill all shows status
         await backfillAllButton.click();
-        await expect(backfillCounter).toHaveScreenshot();
+        await expect(backfillCounter).toContainText(allBackfilledMessage);
```
adrian-estuary, line 68: Seemed like a process to regen the screenshots for the appropriate environment (CI runs in linux). Screenshots seem way too overkill for checking the text inside the pill, so opted to instead use `toContainText` assertions and remove the screenshots.

playwright-tests/tests/captures.spec.ts
jwhartley: For terminology I believe we use binding when talking about a collection tied to a specific task, as in this case.
i.e. a collection is stand-alone, a binding is a collection configured on a capture or materialization.

adrian-estuary: Discussed in slack. We aligned on "collections" as the terminology.

GregorShear (dismissed):
lots of good cleanup in here - thanks!

src/settings/entity.ts
```diff
@@ -13,12 +13,19 @@ import {
     semiTransparentBackground_teal_nesting,
 } from 'src/context/Theme';
 
+// Plain-string nouns for an entity's bindings, as [singular, plural]. Shared so that every
+// chip in the backfill section names bindings the same way.
+export const BINDING_TERMS: Record<Entity, [string, string]> = {
+    capture: ['collection', 'collections'],
+    collection: ['collection', 'collections'],
```
GregorShear, line 20: `collection` key is only present for typecheck, right? not actually consumed anywhere.

adrian-estuary: Correct. Should be functionally equivalent to the `bindingTermId` setting it replaced. So if nothing consumed it before, it should not be consumed now.

src/components/editor/Bindings/Backfill/EvolvedCount.tsx
```diff
@@ -1,51 +1,28 @@
...
+    const bindingTerm =
+        evolvedCollectionsCount === 1 ? bindingTermSingular : bindingTermPlural;
+
     return (
         <Chip
-            aria-label={intl.formatMessage({
-                id: 'workflows.collectionSelector.manualBackfill.count.aria',
-            })}
+            label={`${evolvedCollectionsCount} ${bindingTerm} reversioning`}
+            aria-label="Reversioned count"
```
adrian-estuary, line 23: This was an existing issue. Fixed by updated this aria-label because the `BackfillCount` component already uses "Backfill count". Since it's possible for both components to be mounted at the same time, them having the same aria-label technically breaks `getByLabel` in strict mode.

GregorShear (approved):

PR #2092, opened 2026-09-17
Check the sender and state of OAuth popup responses

skord:
### use-oauth2

-   The message listener acted on any `message` event whose data carried
    `type: 'estuary.oauth2-response'`, regardless of which window sent it. It
    now requires that `message.source` is the popup the hook opened, that
    `message.origin` is our own origin, and that a success payload carries the
    state saved for this flow, compared with the same `base64RemovePadding`
    normalization the popup page uses.
-   Errors arrive without a state, so they rest on the first two checks. A
    message failing any check is ignored and the flow stays open.
-   `OAuthPopup` passes an explicit `targetOrigin` in its `postMessage` rather
    than relying on the default.

### Manually tested

-   not manually tested; needs a real connector OAuth run before merge

### Automated tests

-   new `use-oauth2.test.ts`: the popup's own response, a response from a
    different window, and a response carrying a mismatched state
-   full Vitest suite passes (23 files, 319 tests)

Not run. The E2E suites have no OAuth coverage, so nothing in them reaches the
changed files. The capture path uses `source-hello-world`, which has no OAuth
field.

-   [ ] Admin
-   [ ] Captures
-   [ ] Collections
-   [ ] HomePage
-   [ ] Login
-   [ ] Materializations


skord:
re-review @strix-security

jshearer (approved):

PR #2094, opened 2026-09-18
Add previous/next change navigation to History diffs

### LINKED ISSUE #2093: Add next/previous change navigation to the History diff viewer
## Enhancement

The History tab on an entity details page has no way to jump to a specific change in the diff.

The overview ruler does mark changes, so the problem is not a total lack of a signal. The problem is precision. A ruler marker has a minimum rendered height, so on a spec of several thousand lines a single marker covers a wide range of lines. Clicking it scrolls to that neighbourhood, which can be hundreds of lines, and the actual change may be a single line somewhere inside it. You then hunt for the changed line by eye.

This is common on collections with large inferred schemas, where a publication changes a handful of lines inside a spec several thousand lines long. Support hits this when reconstructing what changed and when.

## Expected

Next change and previous change controls in the History diff toolbar, next to the two publication date labels, that move the diff editor to the change itself rather than to an approximate scroll position.

## Suggested implementation

`DiffViewer.tsx` carries a comment saying this was blocked:

```
// Go to next diff action has an issue -> https://github.com/Microsoft/monaco-editor/issues/2556
```

That comment looks stale. microsoft/monaco-editor#2556 is from 2021, is still open in Backlog, and describes the F7 keybinding selecting the next word instead of jumping to the next change. It predates Monaco's diff editor rewrite.

The `monaco-editor` version currently pinned in `package.json` (`^0.53.0`) exposes a dedicated method rather than the keybinding:

```ts
// monaco.d.ts, on IDiffEditor
/**
 * Jumps to the next or previous diff.
 */
goToDiff(target: 'next' | 'previous'): void;
```

`IStandaloneDiffEditor` extends `IDiffEditor`, and `DiffViewer.tsx` already stores that reference in `diffEditorRef` from its `onMount` handler. So this should be:

1. Two `IconButton` controls in the existing toolbar `Grid`, calling `diffEditorRef.current?.goToDiff('next')` and `goToDiff('previous')`.
2. Two `react-intl` message IDs for the tooltips.
3. A guard on the existing `editorReady` state.
4. Remove the stale comment.

A change counter in the toolbar (for example, "3 of 14") would make this better still, since it tells you how many changes the publication actually contains. `getLineChanges()` returns that list. Optional, and fine to leave out of a first pass.

## Open questions

- `goToDiff` has not been tested in this app. Given the history above it is worth a manual check rather than assuming it works.
- The diff computation is async. `revealFirstDiff()` exists in the same API specifically because of that ("Waits until the diff computation finished"), so the buttons may need the same guard to behave on first render.

If `goToDiff` turns out to misbehave, the fallback is `getLineChanges()` plus `revealLineInCenter` on the modified editor, which gives full control at the cost of reimplementing the navigation.

## Related

- #2035 covers the inline vs side-by-side rendering of the same diff view.

## Additional context

Reported from Solutions while reading publication history on a collection with a large inferred schema.

GregorShear:
Add prev/next buttons to the diff viewer.

It's possible I'm missing an opportunity to clean up some tech debt here, but trying to satisfy a quick request...

Closes #2093.

Screenshot 2026-09-18 at 5 05 02 PM


src/components/shared/Entity/Details/History/DiffViewer.tsx
```diff
@@ -24,7 +31,44 @@ import { useHistoryDiffQueries } from 'src/hooks/useHistoryDiffQueries';
...
+                                    <ArrowUp />
+                                ) : (
+                                    <ArrowDown />
+                                )}
+                            </IconButton>
+                        </span>
+                    </Tooltip>
+                );
+            })}
+        </Grid>
```
adrian-estuary, line 68: Might be easier to read as explicit prev/next buttons instead of using the map + conditionals. I don't foresee needing other states here, so mapping seems a bit overkill. Doesn't seem to be any more LoC either.

```suggestion
        <Grid>
            <Tooltip key={direction} title={label}>
                <span>
                    <IconButton
                        aria-label="Previous change"
                        size="small"
                        disabled={disabled}
                        onClick={() => onNavigate('previous')}
                    >
                        <ArrowUp />
                    </IconButton>
                </span>
            </Tooltip>
            <Tooltip key={direction} title={label}>
                <span>
                    <IconButton
                        aria-label="Next change"
                        size="small"
                        disabled={disabled}
                        onClick={() => onNavigate('next')}
                    >
                        <ArrowDown />
                    </IconButton>
                </span>
            </Tooltip>
        </Grid>
```

src/components/shared/Entity/Details/History/DiffViewer.tsx
```diff
@@ -24,7 +31,44 @@ import { useHistoryDiffQueries } from 'src/hooks/useHistoryDiffQueries';
...
+        <Grid>
+            {(['previous', 'next'] as const).map((direction) => {
+                const label =
+                    direction === 'previous'
+                        ? 'Previous change'
+                        : 'Next change';
+
+                return (
+                    <Tooltip key={direction} title={label}>
+                        <span>
```
adrian-estuary, line 51: Are these `span`s around the `IconButton` necessary? Does `Tooltip` require them?

adrian-estuary (approved):
Just a couple of small comments/questions. Nothing blocking.

src/components/shared/Entity/Details/History/DiffViewer.tsx
GregorShear: Technically yeah, `span` keeps the tooltip working when the button is disabled

https://mui.com/material-ui/react-tooltip/#disabled-elements

src/components/shared/Entity/Details/History/DiffViewer.tsx
GregorShear: i agree the map is weird but it's the best i could do without splitting up `direction`, `label`, `disabled`...

